PR #17, opened 2012-04-14
[Scala] By default, use Scala version configured by SBT (unless a build matrix is defined in .travis.yml)

gildegoma:
Since Scala projects are growing on Travis, I would like to propose following changes
- major: project built with SBT that does not explicitly require a specific Scala version in .travis.yml should be built according to SBT settings (build.sbt and /project/*)
- minor: Try to make clearer that **Using Scala x.y.z** message is more an expectation, rather than a fact
- minor: remove duplicated environment variable SCALA_VERSION
### Details about the changes:
- For SBT: don't force the Scala 2.9.1 by default, but use the version
  defined in SBT own settings. 
  - I think it's definitively better to rely on sbt settings, unless we explicitly require a build-matrix in .travis.yml.
  - Since sbt can automatically download and use any Scala release (e.g. 2.9.2, 2.10.0-M2,...), it will save us the painful administration task to upgrade the default version of the travis-scala builder.
  - Maybe it breaks the convention, that Travis builder always set a **default version**. In this case **undefined** means **scala-version-managed-by-build-tool-own-configuration**.
  - Maybe there is a better keyword rather 'undefined', like _automatic_, _dynamic_, _unmanaged_,... (or even empty string "" ?)
- By the way, I also removed the export of SCALA_VERSION variable (duplicated by TRAVIS_SCALA_VERSION). This variable was designed before all builders were refactored to export TRAVIS_language_VERSION variable. The original variable is thus useless...

### DIFF
--- a/features/test_scala.feature
+++ b/features/test_scala.feature
@@ -2,16 +2,45 @@ Feature: Testing a Scala project
 
   Background:
    Given the following test payload
-     | repository | travis-ci/travis-ci             |
-     | commit     | 1234567                         |
+     | repository | travis-ci/travis-ci                         |
+     | commit     | 1234567                                     |
      | config     | language: scala, scala: 2.9.1, env: FOO=foo |
 
+  Scenario: A successful SBT build without forcing Scala version 
+   Given the following test payload
+     | repository | travis-ci/travis-ci                         |
+     | commit     | 1234567                                     |
+     | config     | language: scala, env: FOO=foo               |
+
+    When it starts a job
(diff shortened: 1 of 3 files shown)


raboof:
Sorry for reviving this ancient thread, but could we reconsider using the scala version from the sbt configuration?

For me it's very confusing when my project is built with a different scala version from the one in my sbt configuration, just because I forgot I have to update both my sbt configuration and the `.travis.yml` when I want to change Scala versions.

Thoughts?

PR #23, opened 2012-07-10
Convert 'go get .' method to 'go get -d && go build'

peterbourgon:
After some discussion on the mailing list and with some colleagues, I think this method is a lot more robust.

We still establish a GOPATH at an arbitrary location, but we no longer need to put the checked-out code in a specific location within it. Instead, we use GOPATH simply as a place to store downloaded dependencies (via `go get -d`) and, later, to look for them during the build step (via `go build`).

`go build` makes no demands on the location of the folder it's called in. It simply uses GOPATH to find referenced imports, and builds whatever artifacts it will build in-place.

I think this is a robust building method that should work well for any Go 1.x.x project. It closely tracks what happens when someone would `go get` a repo remotely, which is the canonical use-case for Go code and something Travis should take care not to conflict with.

### DIFF
--- a/lib/travis/build/job/test/go.rb
+++ b/lib/travis/build/job/test/go.rb
@@ -7,24 +7,18 @@ module Travis
           end
 
           def setup
-            # Here we set up GOPATH + subdirectories structure Go build tool expects
+            # Here we set up GOPATH + subdirectories structure Go build tool expects.
+            # We will just put dependencies here, as source only, via 'go get -d'.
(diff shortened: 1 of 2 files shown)


steenzout:
doesn't this:

```
go_import_path: example.org/pkg/foo
```

on a `.travis.yml` file make it work?

philpennock:
Probably, but note that the incompatible breakage is from 2013 and significantly predates `go_import_path` existing.

PR #66, opened 2013-01-28
WIP: Windows support

Keno:
@joshk as discussed on Chat


photodude:
Sad that this is Stale, windows support seems like an interesting addition.

BanzaiMan:
This has become stale. We may revisit Windows support in the future.

PR #91, opened 2013-03-30
Haskell: Run `cabal update` before running any other scripts (again!)

sol:


juhp:
It's 2020 and I would like the option to run `cabal new-update` instead of `cabal update`.
(`before_install` seems totally fine to me, and would allow this.)

PR #117, opened 2013-07-08
use exit code 113 to denote an error and not exit code 2

joshk:
exit code 2 is a reserved exit code, 113 is recommended for use by
http://www.tldp.org/LDP/abs/html/exitcodes.html


BanzaiMan:
Closing this for now. We would probably want to revisit it in the not-too-distant future, however.

PR #148, opened 2013-10-01
enable bundler caching

rkh:

### DIFF
--- a/lib/travis/build/data.rb
+++ b/lib/travis/build/data.rb
@@ -27,7 +27,7 @@ module Travis
 
       DEFAULT_CACHES = {
         apt:     false,
-        bundler: false
+        bundler: true
       }
 
       attr_reader :data
--- a/spec/data_spec.rb
+++ b/spec/data_spec.rb
@@ -28,6 +28,13 @@ describe Travis::Build::Data do
   describe 'cache' do
     subject(:data) { Travis::Build::Data.new(config: { cache: cache }) }
 
+    describe "default" do
+      let(:cache) { nil }
+      its(:cache) { should be == { } }
+      it { should be_cache(:bundler) }
+      it { should_not be_cache(:edge) }
+    end
+
     describe "single value" do
       let(:cache) { 'bundler' }
       its(:cache) { should be == { bundler: true } }


sferik:
I’m withdrawing my objection to this pull request based on new behavior in recent versions of Bundler.

:+1: for merging this change, assuming the Bundler 1.8 (or later) is part of the default build environment.

joshk:
@sferik whats the new behaviour?

sferik:
@joshk `bundle install` behaves like `bundle update` when there is no `Gemfile.lock`.

I’m not sure exactly when that change was introduced to Bundler. I tried to find it in the history but it looks like the `--local` flag has existed since 2010, so it’s possible I was mistaken when I said `bundle install` has a bias toward local gems.

joshk:
Hmmmm, thanks for the info, have you tested it out on a project on Travis to confirm?

sferik:
@joshk I’ve tested it locally. I can test it on a project on Travis CI now.

joshk:
That would be super awesome!




Does this rely on Bundler 1.8? or will 1.7 ?

sferik:
:+1: [It works!](https://github.com/sferik/twitter/commit/cceef612c4eb33a7b65c382f810e839ee4f0b272) ([Build](https://travis-ci.org/sferik/twitter/builds/52667790))

It seems to work with Bundler 1.7. That said, I’d be happy to see Bundler updated to version 1.8 in the build environment for other reasons.

joshk:
Ok, but I see it using the cache fine, does it download the latest gem update if one exists though?

sferik:
@joshk I guess I’ll have to wait for a new dependent gem to be released. I’ll update this issue as soon as that happens.

joshk:
Cool, thank you, otherwise we can add a `bundle update` after the install if there is no `Gemfile.lock`

sferik:
Here we go! ^^

sferik:
Confirmed!

screen shot 2015-03-03 at 12 42 00 am

joshk:
Welllllll, that is awesome! I'll talk to the team about this!

Thank you so much @sferik !!!!

sferik:
If you decide the release this, be sure to warn people that vendoring gems could cause issues with code coverage. SimpleCov will include all the vendored code as uncovered, potentially causing the tests to fail. I was able to solve this by adding the following line to `spec/spec_helper.rb`:

```diff
 SimpleCov.start do
   add_filter '/spec/'
+  add_filter '/vendor/'
   minimum_coverage(99.62)
 end
```

I believe this could be avoided by vendoring the gems into a hidden directory (e.g. `.bundle`) or a directory outside the build directory (e.g. `/tmp`). Just something to think about before you release the feature.

ZurabWeb:
@rkh Needs rebase.

mvdkleijn:
I was about to enable caching for my bundler / ruby dependencies based on http://docs.travis-ci.com/user/caching/ when I read this...

Is this going to be released soon-ish? If so, I'll wait for it to be automatically enabled.

joshk:
No sorry, we don't have an ETA on this.

BanzaiMan:
Closing for now.

PR #219, opened 2014-03-27
feat(services): start services using Homebrew on OS X

sarahhodne:
This is one of several steps towards making the OS X and Linux environment more similar.

This still needs to be updated more to use the right service names on OS X (I also need to pre-install the services that should be pre-installed), but I figured I'd start the discussion by showing some code.

@svenfuchs I think you were working on something to extract starting services into a Bash script? Maybe it would be a good idea to wait until that's merged.

### DIFF
--- a/lib/travis/build/script/services.rb
+++ b/lib/travis/build/script/services.rb
@@ -12,7 +12,8 @@ module Travis
 
         def start_services
           services.each do |name|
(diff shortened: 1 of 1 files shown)


BanzaiMan:
https://github.com/travis-ci/travis-ci/issues/7158 asks for this, basically.

@joshk Should we pick this up again?

meatballhat:
This was addressed in a different way with #1467

PR #235, opened 2014-04-30
Make git commands much more efficient

albus522:
When the branch is set, we only need to pull down that branch:
Adding --single-branch to the clone command tells git to only pull down the branch we need

When grabbing a specific ref, we only need the one ref:
Instead of pulling down 50 commits from each branch, initialize an empty git repo, setup the remote, and pull down the PR ref.

In the context of one of the apps I am working on, this lowers the git download from 140MB to about 1.6MB. While I realize everyone won't save that much, this should still help a lot.

For another reference, this would save about 140MB of download bandwidth per current rails PR build

### DIFF
--- a/lib/travis/build/script/git.rb
+++ b/lib/travis/build/script/git.rb
@@ -13,9 +13,13 @@ module Travis
           if tarball_clone?
             download_tarball
           else
-            git_clone
-            cd dir
-            fetch_ref if fetch_ref?
+            if fetch_ref?
+              git_init
+              fetch_ref
+            else
+              git_clone
+              cd dir
+            end
             git_checkout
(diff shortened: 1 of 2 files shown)


JeremybellEU:
What blocks this PR from being merged?

whit537:
> What blocks this PR from being merged?

Merge conflicts, apparently, at the very least.

joshk:
Are there any downsides to adopting this approach?

BanzaiMan:
I can resurrect this PR and work on it, or I can wait for @albus522 to redo it.

Either way, this PR is stale because we moved code around, unfortunately, so I'm closing this one.

PR #251, opened 2014-06-16
Add --depth flag to 'git submodule update'

BanzaiMan:
Fixes https://github.com/travis-ci/travis-ci/issues/1867

### DIFF
--- a/spec/shared/git.rb
+++ b/spec/shared/git.rb
@@ -148,7 +148,7 @@ shared_examples_for 'a git repo' do
         end
 
         it 'updates submodules with that depth' do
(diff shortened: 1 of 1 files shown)


nijel:
Is this documented somewhere? In the end I've realized how to configure this, but it would be way easier if it's described in the docs somewhere...

In case somebody else will be looking here, the right way to configure depth for submodules is:

```yml
git:
  submodules_depth: 100
```

thorade:
What happens if the specified commit is not within the specified range?

BanzaiMan:
@thorade The checkout will fail, and mostly so will your build.

PR #260, opened 2014-06-27
Set TRAVIS_TEST_RESULT after executing coverity script

aceofspades:
Previously the result of the Coverity Scan build command was not propagating status.

### DIFF
--- a/lib/travis/build/script/addons/coverity_scan.rb
+++ b/lib/travis/build/script/addons/coverity_scan.rb
@@ -73,6 +73,7 @@ SH
                 env << "COVERITY_SCAN_BUILD_COMMAND_PREPEND=\"#{@config[:build_command_prepend]}\""
                 env << "COVERITY_SCAN_BRANCH_PATTERN=#{@config[:branch_pattern]}"
                 script.cmd "curl -s #{@config[:build_script_url]} | #{env.join(' ')} bash", echo: true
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/addons/coverity_scan.rb
```diff
@@ -73,6 +73,7 @@ def build_command
                 env << "COVERITY_SCAN_BUILD_COMMAND_PREPEND=\"#{@config[:build_command_prepend]}\""
                 env << "COVERITY_SCAN_BRANCH_PATTERN=#{@config[:branch_pattern]}"
                 script.cmd "curl -s #{@config[:build_script_url]} | #{env.join(' ')} bash", echo: true
```
BanzaiMan: If all that is wanted is to fail the build when this `curl` command fails, then it is much cleaner to pass `-f` to `curl` and assert this command:

```ruby
script.cmd "curl -sf #{@config[:build_script_url]} | #{env.join(' ')} bash", echo: true, assert: true
```

BanzaiMan:
Sorry for the delay.

BanzaiMan:
@aceofspades Are you still interested? If you are not, I can just work on this, but your commit might not (though it might) make it to the history.

aceofspades:
@BanzaiMan Not really focus on this at this point, have at it! Cheers

BanzaiMan:
@aceofspades Thanks!

BanzaiMan:
I see a few problems with the addon's logic dealing with various errors. It'll take a little more work than I'd originally thought.

PR #278, opened 2014-08-04
Enhancement: allows install to be performed through checked in composer....

till:
...phar
- re-activates automatic composer install on travis-ci
- supports running install with a checked in composer.phar
- doesn't self-update the checked in composer.phar

Wrote blog post for updates:
- http://till.klampaeckel.de/blog/archives/204-Whats-wrong-with-composer-and-your-.travis.yml.html

Write tests for:
- [x] `composer self-update` is run
- [x] `composer install` is run
- [x] `composer.phar install` is run
- [ ] `composer self-update`is not run
- [x] `composer_args` are used

### DIFF
--- a/lib/travis/build/script/php.rb
+++ b/lib/travis/build/script/php.rb
@@ -46,7 +46,7 @@ module Travis
             end
             self.else do |sub|
               sub.cmd "composer install #{config[:composer_args]}".strip, fold: 'install.composer'
-            end 
+            end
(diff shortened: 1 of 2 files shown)


meatballhat:
Hello and sorry about the very long delay!  Please holler if you are an interested party and you would like to see this work picked up again, assuming it hasn't been addressed another way already.

PR #285, opened 2014-08-21
Warn if no 'build-tools' is specified in an Android project

BanzaiMan:
As discussed in https://github.com/travis-ci/travis-cookbooks/pull/360#issuecomment-52902921
Communicate with the user that build-tools should be specified.

If there is a possibility that an Android project does not need one,
then we should detect it and suppress this warning.


BanzaiMan:
This was invalid (and has also gone stale). A new one will be coming soon.

PR #323, opened 2014-11-06
Alias swift to Objective-C

kylef:


BanzaiMan:
I've merged this. If Swift's build logic diverges from that of Objective-C, we can look at separating them.

BanzaiMan:
It's been deployed to production, but the builds will fail until we route the builds to the Mac infrastructure.

PR #349, opened 2014-11-25
add upload_pull_request option to artifacts addon

rkh:
Currently artefacts are never uploaded for PR builds. If a project has a PR focused workflow and uses artifacts to upload for instance extended build logs, then the artefacts feature is more or less unusable without an option like this.

Not sure `upload_pull_request` is the best name for the option.

Current syntax would be:

``` yaml
addons:
  artifacts:
    key: "YOUR AWS ACCESS KEY"
    secret: "YOUR AWS SECRET KEY"
    bucket: "S3 Bucket"
    upload_pull_request: true
```

### DIFF
--- a/lib/travis/build/addons/artifacts/validator.rb
+++ b/lib/travis/build/addons/artifacts/validator.rb
@@ -36,11 +36,15 @@ module Travis
             end
 
             def validate_push_request
-              error :pull_request if pull_request?
+              error :pull_request if pull_request? and not upload_pull_request?
             end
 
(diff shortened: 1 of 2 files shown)


andrusha:
Was there any progress on this issue? Is it going to be ever done?

meatballhat:
This dusty PR is headed to the *Closed* heap!  Please holler if you are an interested party and you would like to see this work picked up again, assuming it hasn't been addressed another way already.

PR #352, opened 2014-12-04
Coverity: Handle '"upload_permitted":false' as a warning, not an error

fxlb:
Using matrix of builds can give errors because when a job uploaded
analysis results, the curl command status (line 43) change from
`{"upload_permitted":true}` to
`{"upload_permitted":false,"next_upload_permitted_at":"<date>"}`

For some jobs of the same build, the status can be set to failed (`exit 1`),
so the build status is failed, same for the project.

To avoid this, it would be useful to handle `"upload_permitted":false` as a
warning, changing `exit 1` to `exit 0`.

The results of Coverity analysis is given via Coverity postprocessing.


fxlb:
What do you think about it?

fxlb:
The merge is done.
Thank you!.
Closing.

PR #356, opened 2014-12-12
JWT addon

sebv:
JWT token authentication for PR.

Users need to add this to `.travis.yml`.

```
addons:
  jwt: 
     secure: <SAUCE_ACCESS_KEY=123456789 encrypted> 
```

The JWT addon will then create a time-limited JWT token and export it to the SAUCE_ACCESS_KEY environment variable.

This is not specific to sauce labs several services may be configured, for instance:
```
addons:
  jwt:
     saucelabs:
        secure: <SAUCE_ACCESS_KEY=123456789 encrypted>
     moonwalk:
        secure: <MOONWALK_SECRET=123456789 encrypted>
```
In this case 2 tokens will be generated, one in the SAUCE_ACCESS_KEY env variable, the other in MOONWALK_SECRET.

### DIFF
--- a/lib/travis/build/addons/jwt.rb
+++ b/lib/travis/build/addons/jwt.rb
@@ -1,0 +1,38 @@
+require 'travis/build/addons/base'
+require 'jwt'
+
+module Travis
+  module Build
+    class Addons
+      class Jwt < Base
+        SUPER_USER_SAFE = true
+
+        def before_before_script
+          secrets = nil
+          if config.is_a?(String)
+            secrets = [config]
+          else
+            secrets = config.values
+          end
+
+          tokens = {}
+          secrets.each do |secret|
+            key, secret = secret.split('=').map(&:strip)
+            pull_request = self.data.pull_request ? self.data.pull_request : ""
+            payload = {"slug" => self.data.slug,
+                      "pull-request" => pull_request,
+                      "iat" => Time.now.to_i()}
+            token = JWT.encode(payload, secret)
+            tokens[key] = token
+          end
+          sh.fold 'addons_jwt' do
+            sh.echo 'Initializing JWT', ansi: :yellow
+            tokens.each do |key, val|
+              sh.export key, val, echo: false
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/travis/build/addons/jwt.rb
+++ b/lib/travis/build/addons/jwt.rb
@@ -10,7 +10,6 @@ module Travis
         def before_before_script
           tokens = {}
           Array(config).each do |secret|
-            key, secret = secret.split('=').map(&:strip)
             pull_request = self.data.pull_request ? self.data.pull_request : ""
             now = Time.now.to_i()
             payload = {
@@ -21,12 +20,13 @@ module Travis
               "iat" => now
             }
             begin
+              key, secret = secret.split('=').map(&:strip)
               tokens[key] = JWT.encode(payload, secret)
-            rescue Exception => e
-              sh.failure "JWT Encode Error: #{e.message}"
-              []
+            rescue Exception
+              sh.echo "There was an error while encoding JWT. If the secret is encrypted, ensure that it is encrypted correctly.", ansi: :yellow
             end
           end
+          return if tokens.empty?
           sh.fold 'addons_jwt' do
             sh.echo 'Initializing JWT', ansi: :yellow
             tokens.each do |key, val|
--- a/lib/travis/build/addons/jwt.rb
+++ b/lib/travis/build/addons/jwt.rb
@@ -19,9 +19,11 @@ module Travis
           secrets.each do |secret|
             key, secret = secret.split('=').map(&:strip)
             pull_request = self.data.pull_request ? self.data.pull_request : ""
+            now = Time.now.to_i()
             payload = {"slug" => self.data.slug,
                       "pull-request" => pull_request,
-                      "iat" => Time.now.to_i()}
+                      "exp" => now+5400,
+                      "iat" => now}
             token = JWT.encode(payload, secret)
             tokens[key] = token
           end
--- a/lib/travis/build/addons/jwt.rb
+++ b/lib/travis/build/addons/jwt.rb
@@ -20,7 +20,8 @@ module Travis
             key, secret = secret.split('=').map(&:strip)
             pull_request = self.data.pull_request ? self.data.pull_request : ""
             now = Time.now.to_i()
-            payload = {"slug" => self.data.slug,
+            payload = {"iss" => "travis-ci.org",
+                      "slug" => self.data.slug,
                       "pull-request" => pull_request,
                       "exp" => now+5400,
                       "iat" => now}
--- a/lib/travis/build/addons/jwt.rb
+++ b/lib/travis/build/addons/jwt.rb
@@ -8,27 +8,25 @@ module Travis
         SUPER_USER_SAFE = true
 
         def before_before_script
-          secrets = nil
-          if config.is_a?(String)
-            secrets = [config]
-          elsif config.is_a?(Array)
-            secrets = config
-          else
-            secrets = config.values
-          end
-
           tokens = {}
-          secrets.each do |secret|
+          Array(config).each do |secret|
             key, secret = secret.split('=').map(&:strip)
             pull_request = self.data.pull_request ? self.data.pull_request : ""
             now = Time.now.to_i()
-            payload = {"iss" => "travis-ci.org",
-                      "slug" => self.data.slug,
-                      "pull-request" => pull_request,
-                      "exp" => now+5400,
-                      "iat" => now}
-            token = JWT.encode(payload, secret)
-            tokens[key] = token
+            payload = {
+              "iss" => "Travis CI, GmbH",
+              "slug" => self.data.slug,
+              "pull-request" => pull_request,
+              "exp" => now+5400,
+              "iat" => now
+            }
+            begin
+              tokens[key] = JWT.encode(payload, secret)
+            rescue Exception => e
+              sh.echo "JWT Encode Error: #{e.message}", ansi: :red
+              warn e
+              []
+            end
           end
           sh.fold 'addons_jwt' do
             sh.echo 'Initializing JWT', ansi: :yellow
--- a/lib/travis/build/addons/jwt.rb
+++ b/lib/travis/build/addons/jwt.rb
@@ -11,6 +11,8 @@ module Travis
           secrets = nil
           if config.is_a?(String)
             secrets = [config]
+          elsif config.is_a?(Array)
+            secrets = config
           else
             secrets = config.values
           end
--- a/lib/travis/build/addons/jwt.rb
+++ b/lib/travis/build/addons/jwt.rb
@@ -23,8 +23,7 @@ module Travis
             begin
               tokens[key] = JWT.encode(payload, secret)
             rescue Exception => e
-              sh.echo "JWT Encode Error: #{e.message}", ansi: :red
-              warn e
+              sh.failure "JWT Encode Error: #{e.message}"
               []
             end
           end
--- a/spec/build/addons/jwt_spec.rb
+++ b/spec/build/addons/jwt_spec.rb
@@ -1,0 +1,35 @@
+require 'spec_helper'
+
+describe Travis::Build::Addons::Jwt, :sexp do
+  let(:script) { stub('script') }
+  let(:data)   { payload_for(:push, :ruby, config: { addons: { jwt: config } }) }
+  let(:sh)     { Travis::Shell::Builder.new }
+  let(:addon)  { described_class.new(script, sh, Travis::Build::Data.new(data), config) }
+  subject      { sh.to_sexp }
+  before       { Time.stubs(:now).returns(Time.mktime(1970,1,1)) }
+  before       { addon.before_before_script }
+
+  describe 'jwt token, one secret' do
+    let(:config) { 'MY_ACCESS_KEY=987654321' }
+    it "should work" do
+      subject.should include_sexp [:echo, 'Initializing JWT', ansi: :yellow]
+      expected = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbHVnIjoidHJhdmlzLWNpL3RyYXZpcy1jaSIsInB1bGwtcmVxdWVzdCI6IiIsImlhdCI6MH0.vHZdjuxQt6CdFmWPEJLAsVbI_KCxbwJVbgT7jaTxkK4"
+      subject.should include_sexp [:export, ['MY_ACCESS_KEY', expected]]
+    end
+  end
+
+  describe 'jwt token, several secrets' do
+    let(:config) { {
+      secret1: 'MY_ACCESS_KEY_1=123456789',
+      secret2: 'MY_ACCESS_KEY_2=ABCDEF'
+    } }
+    it "should work" do
+      subject.should include_sexp [:echo, 'Initializing JWT', ansi: :yellow]
+      expected1 = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbHVnIjoidHJhdmlzLWNpL3RyYXZpcy1jaSIsInB1bGwtcmVxdWVzdCI6IiIsImlhdCI6MH0.iO-I8CNXhQlgw_dL8R9CDaDPplSC3yf9J399Uumn6CE"
+      subject.should include_sexp [:export, ['MY_ACCESS_KEY_1', expected1]]
+      expected2 = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbHVnIjoidHJhdmlzLWNpL3RyYXZpcy1jaSIsInB1bGwtcmVxdWVzdCI6IiIsImlhdCI6MH0.p1B9HnYCh-z5Igek5tr_QVPb1zV1ucwcPZNvY039WKg"
+      subject.should include_sexp [:export, ['MY_ACCESS_KEY_2', expected2]]
+    end
+  end
+end
+
--- a/spec/build/addons/jwt_spec.rb
+++ b/spec/build/addons/jwt_spec.rb
@@ -33,7 +33,7 @@ describe Travis::Build::Addons::Jwt, :sexp do
       end
     end
 
-    describe 'handle malformed secret/key' do
+    describe 'handle raising an exception on 100%' do
       let(:config) { ["JUSTKEY"]; }
 
       it "should output warning about bad data" do
@@ -41,16 +41,8 @@ describe Travis::Build::Addons::Jwt, :sexp do
       end
     end
 
-    describe 'handle raising an exception on 100%' do
-      let(:config) { [nil]; }
-
-      it "should output warning about bad data" do
-        expect(subject).to include_sexp [:echo, "There was an error while encoding JWT. If the secret is encrypted, ensure that it is encrypted correctly.", {:ansi=>:yellow}]
-      end
-    end
-
     describe 'handle raising an exception on 50%' do
-      let(:config) { [nil, 'BAD_ACCESS_KEY=abc123']; }
+      let(:config) { ["JUSTKEY", 'BAD_ACCESS_KEY=abc123']; }
 
       it "should output warning about bad data" do
         # first one throws an exception
--- a/spec/build/addons/jwt_spec.rb
+++ b/spec/build/addons/jwt_spec.rb
@@ -13,7 +13,7 @@ describe Travis::Build::Addons::Jwt, :sexp do
     let(:config) { 'MY_ACCESS_KEY=987654321' }
     it "should work" do
       subject.should include_sexp [:echo, 'Initializing JWT', ansi: :yellow]
-      expected = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJ0cmF2aXMtY2kub3JnIiwic2x1ZyI6InRyYXZpcy1jaS90cmF2aXMtY2kiLCJwdWxsLXJlcXVlc3QiOiIiLCJleHAiOjU0MDAsImlhdCI6MH0.soQJgHR6cGNr9Lj_N6yL2Nk5SQug-hXGUPenJy1QTVc"
+      expected = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJ0cmF2aXMtY2kub3JnIiwic2x1ZyI6InRyYXZpcy1jaS90cmF2aXMtY2kiLCJwdWxsLXJlcXVlc3QiOiIiLCJleHAiOjM0MjAwLCJpYXQiOjI4ODAwfQ.NQLflEZgXkZY80frfSPfUQVYk0chvStFseUrU1HDRJk"
       subject.should include_sexp [:export, ['MY_ACCESS_KEY', expected]]
     end
   end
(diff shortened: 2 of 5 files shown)


BanzaiMan:
Could you add documentation on how to use it, and security implications thereof, in the form of a PR to https://github.com/travis-ci/docs-travis-ci-com?

BanzaiMan:
This has been deployed to staging. Please go to https://staging.travis-ci.org/profile, sync data with GitHub and enable a repository to test this.

The staging environment might be refreshed with a different code without prior notice, so if you don't see `Initializing JWT` in the build log when you have the addon set, do let us know, so that we can get the correct code out.

sourishkrout:
Thank you @BanzaiMan for moving this along. The JWT addon is not expect to work without the changes submitted as part of the PR in https://github.com/travis-ci/travis-core/pull/418. - I now realize we should have included that in the initial description. Is it possible to get the travis-core changes deployed to staging as well?

BanzaiMan:
@sourishkrout The PR to travis-core has been deployed to staging. Could you try it? (Make sure that you use no critically sensitive data.)

sourishkrout:
Unfortunately it doesn't look like it worked. 

Added jwt addon config to my .travis.yml (https://github.com/sourishkrout/osbdemo/blob/master/.travis.yml) according to this:
```
Repository slug: |travis-ci/travis-core| sourishkrout/osbdemo
Reading from stdin, press Ctrl+D when done
SAUCE_BLA=12345678
Please add the following to your .travis.yml file:

  secure: "eDQEcNFA0nQzTUj1HmnedM8+vWtaxLehmEgDzf/EEI/Bk6tYcnJMqpA0VMklUDsdmkpxpMRgspdrv1+T4LdYFGMJUOJ+4A7DjkEiTvpKammhgRRYja3VjSpBjx8svxvxwLOHY6l81xRglsdjyWrCx87gUNBr5jFZ5Pd0SJM3LqI="
```

Which didn't produce the JWT ENV var I expected (`SAUCE_BLA=<jwt token>`):

```
> export | grep SAUCE
export SAUCE_USERNAME='sourishkrout'
export npm_lifecycle_script='export | grep SAUCE'
export npm_package_scripts_test='export | grep SAUCE'
```

@BanzaiMan I wonder how we can debug this. I'm curious to find out if the decrypted JWT addon config from travis-core made it into travis-build at all?

Travis CI Job: https://staging.travis-ci.org/sourishkrout/osbdemo/builds/400203

BanzaiMan:
Hi, @sourishkrout and @sebv,

We are itching to test this further. Could you advise us when we can coordinate our efforts to do so? You can respond here, or email support@travis-ci.com.

Thank you!

sebv:
@BanzaiMan also rebased this one.

sebv:
btw  [this yml format](https://github.com/BanzaiMan/travis_staging_test/blob/c60885ce90cd6420ba557ac120b30650476a0cd1/.travis.yml#L10-L12) is correct.

admc:
Hi @BanzaiMan, we are really interested in getting this project completed, Any chance you can reach out to me admc@saucelabs.com so I can work to arrange a live session where we can get everyone to work together to finish this? Thanks!

BanzaiMan:
Sorry for the delay. I'll look at this shortly!

spec/build/addons/jwt_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  describe 'jwt token, several secrets' do
+    let(:config) { {
+      secret1: 'MY_ACCESS_KEY_1=123456789',
+      secret2: 'MY_ACCESS_KEY_2=ABCDEF'
+    } }
+    it "should work" do
+      subject.should include_sexp [:echo, 'Initializing JWT', ansi: :yellow]
+      expected1 = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbHVnIjoidHJhdmlzLWNpL3RyYXZpcy1jaSIsInB1bGwtcmVxdWVzdCI6IiIsImlhdCI6MH0.iO-I8CNXhQlgw_dL8R9CDaDPplSC3yf9J399Uumn6CE"
+      subject.should include_sexp [:export, ['MY_ACCESS_KEY_1', expected1]]
+      expected2 = "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbHVnIjoidHJhdmlzLWNpL3RyYXZpcy1jaSIsInB1bGwtcmVxdWVzdCI6IiIsImlhdCI6MH0.p1B9HnYCh-z5Igek5tr_QVPb1zV1ucwcPZNvY039WKg"
```
BanzaiMan: Are these expected values perhaps platform-dependent?

On my Mac, I get this failure:

```
  2) Travis::Build::Addons::Jwt jwt token, several secrets should include sexp :export and ["MY_ACCESS_KEY_1", "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbHVnIjoidHJhdmlzLWNpL3RyYXZpcy1jaSIsInB1bGwtcmVxdWVzdCI6IiIsImlhdCI6MH0.iO-I8CNXhQlgw_dL8R9CDaDPplSC3yf9J399Uumn6CE"]
     Failure/Error: should include_sexp [:export, ['MY_ACCESS_KEY_1', expected1]]
       Expected the generated sexp to include:

       [:export, ["MY_ACCESS_KEY_1", "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbHVnIjoidHJhdmlzLWNpL3RyYXZpcy1jaSIsInB1bGwtcmVxdWVzdCI6IiIsImlhdCI6MH0.iO-I8CNXhQlgw_dL8R9CDaDPplSC3yf9J399Uumn6CE"]]

       Found the following similar nodes:

       [:export, ["MY_ACCESS_KEY_1", "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbHVnIjoidHJhdmlzLWNpL3RyYXZpcy1jaSIsInB1bGwtcmVxdWVzdCI6IiIsImlhdCI6MTgwMDB9.6s1WY-acH-kd9B0VsMjnzwrTgN2X9lEBbNfDDLfSG8M"]]
       [:export, ["MY_ACCESS_KEY_2", "eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbHVnIjoidHJhdmlzLWNpL3RyYXZpcy1jaSIsInB1bGwtcmVxdWVzdCI6IiIsImlhdCI6MTgwMDB9.FrkyJ7CI8xpAcVB8vea6NV2aLKG3ElN3R3J1nVbkgWQ"]]
     # ./spec/build/addons/jwt_spec.rb:29:in `block (3 levels) in <top (required)>'
```

which we do not get on Travis CI.

If so, this is a poor spec in my opinion.

I am not sure, though, how we might test these otherwise. Thoughts?

BanzaiMan:
By the way, here's the build from the rebased try on Travis: https://travis-ci.org/travis-ci/travis-build/builds/63575373, which passes. (Note that I made slight changes to the spec syntax, to silence the warnings about old "should" syntax.)

I apologize for not understanding the mechanism behind this encryption scheme enough to offer a better approach. How can we make these specs platform-independent? I see that a good portion of the encrypted data do match up in the beginning. Is that good enough for specs?

sebv:
@BanzaiMan could you put some debug here: https://github.com/saucelabs/travis-build/blob/jwt/lib/travis/build/addons/jwt.rb#L25 and run the test for mac and linux, so we can see the difference before encoding. 

We never ran the test on mac, not sure how to set those up.

sourishkrout:
I didn't to the digging to figure out why but the `iat` (issued at) is different for aboves spec run on Mac which will result in different payloads and signed hashes:

```
{
  "slug": "travis-ci/travis-ci",
  "pull-request": "",
  "iat": 18000
}
```

vs

```
{
  "slug": "travis-ci/travis-ci",
  "pull-request": "",
  "iat": 0
}
```

sebv:
Looks like timezone issue, should probably initialize with GMT here https://github.com/saucelabs/travis-build/blob/jwt/spec/build/addons/jwt_spec.rb#L9 .

joshk:
@BanzaiMan could you also please make sure you fully understand and grasp how JWT works and how it can be used with Sauce Labs.

BanzaiMan:
Confirmed that the specs are time (zone) dependent.

BanzaiMan:
I fixed the specs with https://github.com/travis-ci/travis-build/commit/e1e54d0

BanzaiMan:
This (along with https://github.com/travis-ci/travis-core/pull/418) has been deployed to staging. My limited test seems reasonable (works as far as I can tell), and I would like to have a larger set of users test it.

To use this in staging, follow the following steps:

1. Enable a repository at https://staging.travis-ci.org/profile (You can build against only one of Travis environment, so if you are currently using Travis CI production, i.e., https://travis-ci.org, you need to switch back to it when you are done with the experiment.)
1. Encrypt the variable against the staging API endpoint:

    `travis encrypt -r REPO/OWNER -e https://api-staging.travis-ci.org SUPER_SECRET_STUFF`
1. Create a PR against it. When you try dumping `SUPER_SECRET_STUFF`, you should see JWT-encoded string instead of the plain text value (https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/430639#L101-L102):

    `eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzbHVnIjoiQmFuemFpTWFuL3RyYXZpc19zdGFnaW5nX3Rlc3QiLCJwdWxsLXJlcXVlc3QiOjksImlhdCI6MTQzMjgzNzA4OH0.7GBV_50n04rTwBHzxTpBMFBqWAkojdlsW5yMTnrzQ_Y`

The staging environment may be used for other tests, so if you suspect it's not working, please check with me if the correct code is deployed. Thank you!

BanzaiMan:
@sebv @sourishkrout @samccone @santiycr

Could you describe how the user is supposed to decode the payload? We will need documentation on usage when we deploy this to production!

Thank you.

sebv:
@BanzaiMan I think @sourishkrout already has an example here: https://github.com/sourishkrout/osbdemo , just need to be configured with staging. 

The user doesn't need to decode, just use the variable SAUCE_ACCESS_KEY (or else)  in the test script.

The missing doc is probably what thirdparty need to do to validate the key. (This has already been implemented on Sauce side.)

BanzaiMan:
@sebv Ah, OK. I guess I misunderstood the use case.

Yes, if @sourishkrout could reconfigure his demo to use staging, and ensure it works as expected, that would be great.

Please write up a HOWTO on this addon. Thanks.

sourishkrout:
I'm in the process of re-configuring the `sourishkrout/osbdemo` to test JWT on staging. I'll report my findings in a bit.

samccone:
I am in the same boat as @sebv with the sauce key.  Also it looks like Orgs do not work on the staging server, so it is hard for me to test the todoMVC example currently.

sebv:
@samccone you can probably just fork the repo to test it against staging, merge back the changes later.

sourishkrout:
I've set up another repo (https://staging.travis-ci.org/sourishkrout/grunt-saucelabs) and hooked it up to the Travis' staging. Good news is that the JWT addon generates valid tokens and passes them into the build job ENV as expected. It's being rejected (HTTP 401) on our REST API now which I will debug and hopefully fix. I'll keep you guys posted.

Great work @BanzaiMan + @sebv. That's a great milestone! I think we're really close.

```
POST /rest/v1/osbdemo/js-tests HTTP/1.0
X_FORWARDED_PROTO: https
X_FORWARDED_HOST: sebastian.dev.saucelabs.net
X_FORWARDED_FOR: 199.91.170.222
Authorization: Basic
<JWT TOKEN stripped>
Host: resto_clones
Connection: close
accept: application/json
content-type: application/json
content-length: 281

{"platforms":[["Windows 7","chrome",""]],"url":"http://127.0.0.1:9999/custom/succeeds.html","framework":"custom","build":"430668","public":"team","tags":["v8.6.1","grunt-saucelabs"],"name":"saucelabs-custom:tunnel-test","video-upload-on-pass":false,"tunnel-identifier":"202172686"}HTTP/1.0 401 Unauthorized

Date: Fri, 29 May 2015 23:04:48 GMT
Content-Type: application/json
WWW-Authenticate: Basic realm="Sauce"
Server: monocle/0.35

{"error": "Not authorized"}
```

BanzaiMan:
@sourishkrout @sebv @samccone How is the document coming along?

sebv:
Once @sourishkrout has finished his testing, sauce ecoteam need to plan a writedown (cc @jlipps).

BanzaiMan:
@sebv OK. As I indicated earlier, staging env may not have the code deployed at any given time, since we are testing a few things. If you don't see it working there, do get in touch.

sourishkrout:
@BanzaiMan the JWT addon works as expected. However, I've added a `exp` property to the token. That'll make sure the tokens will expire eventually and, even if leaked, can't cause harm indefinitely.

I'm still running into an issue with Sauce's implementation on our side. We will have to make additional changes to `Sauce Connect` to make the end to end integration work. I'm currently working on that.
The `JWT addon` and the remaining work on `Sauce Connect` aren't mutually exclusive so I suggest we get the `JWT addon` merged asap.

lib/travis/build/addons/jwt.rb
```diff
@@ -0,0 +1,40 @@
...
+          end
+
+          tokens = {}
+          secrets.each do |secret|
+            key, secret = secret.split('=').map(&:strip)
+            pull_request = self.data.pull_request ? self.data.pull_request : ""
+            now = Time.now.to_i()
+            payload = {"slug" => self.data.slug,
+                      "pull-request" => pull_request,
+                      "exp" => now+5400,
```
sourishkrout: @BanzaiMan `5400` seconds for 1.5 hours. Any idea how we could get hold of the timeout associated with Travis builds? Couldn't spot anything obvious.

BanzaiMan: @sourishkrout Looking at the typical build payload, I do not think that information is available. It seems to me that we would need to add that.

@svenfuchs, could you confirm?

sourishkrout: @BanzaiMan I feel like it makes sense to set this to a value, e.g. `4 hours`, to make sure the tokens expire eventually if it's too hard to get the max build run time into the build payload?

sourishkrout:
@BanzaiMan I see you've merged the code. Does this mean it'll be available in production soon?

BanzaiMan:
@sourishkrout I did? If I had done so, it would have been an accident. I am waiting for the documentation before merging it all in. :-D

sebv:
@BanzaiMan I think the doc text is ready, we need to move it somewhere, could you give us indication.

See: https://github.com/sebv/jwt-doc/issues/4

joeltine:
We're eagerly awaiting this change to use Travis+Sauce securely for Closure. Is this still moving along? Any idea when it will land?

Borets:
Hi @BanzaiMan. I represnet the Product Team here at Sauce Labs and we've just released the version of Sauce Connect that provides a full support for JWT. At this point this pull request is the only thing that is left before the process can be considered complete. Please let us know if there is anything else we need to for you to merge it. The documentation can be found here https://github.com/sebv/jwt-doc.

johanneswuerbach:
Any updates here? Waiting eagerly for secure SauceLabs pull requests.

jlipps:
hey @joshk just wanted to bump this up. What do we need to do to "synergize" this over the "goal line"? 💼

halkeye:
@BanzaiMan I've rebased both patches so they should now work against master again

Also updated the old tests to the new rspec format

Let me know if there's anything else.

BanzaiMan:
I'll be testing this shortly.

lib/travis/build/addons/jwt.rb
```diff
@@ -0,0 +1,41 @@
...
+    class Addons
+      class Jwt < Base
+        SUPER_USER_SAFE = true
+
+        def before_before_script
+          secrets = nil
+          if config.is_a?(String)
+            secrets = [config]
+          else
+            secrets = config.values
```
BanzaiMan: I think it is possible that `config` may not respond to `#values`. For example, when it is an array.

lib/travis/build/addons/jwt.rb
```diff
@@ -0,0 +1,41 @@
...
+          end
+
+          tokens = {}
+          secrets.each do |secret|
```
BanzaiMan: When decryption fails, `secrets` will contain `nil`…

lib/travis/build/addons/jwt.rb
```diff
@@ -0,0 +1,41 @@
...
+            key, secret = secret.split('=').map(&:strip)
+            pull_request = self.data.pull_request ? self.data.pull_request : ""
+            now = Time.now.to_i()
+            payload = {"iss" => "travis-ci.org",
+                      "slug" => self.data.slug,
+                      "pull-request" => pull_request,
+                      "exp" => now+5400,
+                      "iat" => now}
+            token = JWT.encode(payload, secret)
```
BanzaiMan: …and this call to `JWT.encode` will throw exception. This should be accounted for. And communicate that to the user.

BanzaiMan: This is particularly important, because the push builds on the fork will throw this exception.

BanzaiMan: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/507059 shows what we see when we encounter the exception.

BanzaiMan:
https://staging.travis-ci.org/travis-ci/travis_staging_test/builds/507067#L148 shows the JWT token thus generated and seen in a PR build.

lib/travis/build/addons/jwt.rb
BanzaiMan: `config` doesn't have to be a Hash or a String, right?

BanzaiMan: https://github.com/travis-ci/docs-travis-ci-com/pull/674/files#r72269084

lib/travis/build/addons/jwt.rb
```diff
@@ -0,0 +1,43 @@
...
+        SUPER_USER_SAFE = true
+
+        def before_before_script
+          secrets = nil
+          if config.is_a?(String)
+            secrets = [config]
+          elsif config.is_a?(Array)
+            secrets = config
+          else
+            secrets = config.values
```
BanzaiMan: I am not sure if this should be a desirable default.

As I mentioned previously, I do not think `config` should be a Hash at all, since the keys are discarded.

With that in mind, this portion can be simplified to:

```ruby
secrets = Array(config)
```

halkeye: Okay I'll look into it more.

Does supporting strings at all make sense? I'm not super comfortable with
writing Travis yaml files so I'm still learning as I go.

BanzaiMan: I don't think singling out strings make sense here.

halkeye: Awesome.

Next break I'll update the code, tests, and docs

lib/travis/build/addons/jwt.rb
BanzaiMan: You can stub `JWT.encode` and raise an exception.

lib/travis/build/addons/jwt.rb
```diff
@@ -0,0 +1,43 @@
...
+          end
+
+          tokens = {}
+          secrets.each do |secret|
+            key, secret = secret.split('=').map(&:strip)
+            pull_request = self.data.pull_request ? self.data.pull_request : ""
+            now = Time.now.to_i()
+            payload = {"iss" => "travis-ci.org",
```
BanzaiMan: Not sure if this is a reasonable issuer. Keep in mind that it could come from other domains, more generic value, say `Travis CI, GmbH` might be better.

Also, it is worth noting this value (whatever we end up using) in the document so that other third-party implementer could verify `iss`.

halkeye: is `Travis CI, GmbH` a static string? or a constant somewhere I can use?

BanzaiMan: Good question. I don't know if that is static, or what.

@joshk Do you have opinions?

halkeye: Or even a way to pull more useful ones? Like what product/domain we are on?

BanzaiMan: Not sure if that is a good idea. Having `iss` dynamically assigned means that the consumer will have to know all possible values in advance and deal with them accordingly.

lib/travis/build/addons/jwt.rb
halkeye: @BanzaiMan spec/build/addons/jwt_spec.rb doesn't cover decryption right?

lib/travis/build/addons/jwt.rb
halkeye: I was trying to look at what other addons do.

Looks like just wrap it in a begin/rescue then warn then return []

Does that seem like right?

BanzaiMan: I think that is reasonable to do here.

halkeye: Every way I look at it, JWT.encode doesn't throw an exception
I can certainly handle an exception in case, but the only way I figure is if payload isn't a valid object.

@BanzaiMan as you said earlier, config could be null? Any idea how to replicate that?
`let(:config) { nil; }` returns still makes config = `{}`

halkeye: yea, lib/travis/build/addons/base.rb has:
```
        def normalize_config(config)
          case config
          when Fixnum, Float, String, Array, Hash
            config
          else
            {}
          end
        end
```

Which "fixes" that data.

BanzaiMan: It's quite possible that my initial analysis was wrong. An exception does occur here in some cases, though, so it is important that you handle it. I'll take another look later.

halkeye: Okay, i'll commit my code that handles an exception and the test, but I
won't know that is the exception till its tested again.

halkeye:
@BanzaiMan should all be addressed now

lib/travis/build/addons/jwt.rb
BanzaiMan: The build that failed initially (https://staging.travis-ci.org/travis-ci/travis_staging_test/builds/507055) had the data encrypted incorrectly, I believe, using the encryption key of a different repository. This results in a stack trace such as this: https://gist.github.com/BanzaiMan/a8e7116c6a80abf34e1fa48efa1e9dcb. (The trace corresponds to https://github.com/travis-ci/travis-build/pull/356/commits/bb16a55b414a22dd66ab41c924251d85a665efff.)

lib/travis/build/addons/jwt.rb
```diff
@@ -0,0 +1,40 @@
...
+              "iss" => "Travis CI, GmbH",
+              "slug" => self.data.slug,
+              "pull-request" => pull_request,
+              "exp" => now+5400,
+              "iat" => now
+            }
+            begin
+              tokens[key] = JWT.encode(payload, secret)
+            rescue Exception => e
+              sh.failure "JWT Encode Error: #{e.message}"
```
BanzaiMan: Note that `sh.failure` would halt the build immediately. See https://staging.travis-ci.org/travis-ci/travis_staging_test/builds/507053#L138.

I tend to think this is not the desired behavior. I think is better to just inform the user of the error with `sh.echo` (the usefulness of the message `e.message` is dubious, as you can see from the stack trace above; `no implicit conversion of nil into String` does not tell the user what to do) and move on.

A useful message might be something like:

    There was an error while encoding JWT. If the secret is encrypted, ensure that it is encrypted correctly.

lib/travis/build/addons/jwt.rb
```diff
@@ -0,0 +1,40 @@
...
+              []
```
BanzaiMan: This value is not used anywhere, so you can remove it.

lib/travis/build/addons/jwt.rb
```diff
@@ -0,0 +1,40 @@
...
+            end
+          end
```
BanzaiMan: If `tokens` is empty, there is no reason to move on.

```ruby
return if tokens.empty?
```

BanzaiMan:
With the suggested changes (https://github.com/travis-ci/travis-build/commit/e46ff5a6316baef36ea43c081e53d42c9cf18b92), it would look like this: https://staging.travis-ci.org/travis-ci/travis_staging_test/builds/507077#L138

halkeye:
Awesome

I figured out how to reproduce the testcase, and moved things slightly, but that should be everything.

I'll do another pass over of the docs tomorrow

halkeye:
I guess https://github.com/travis-ci/travis-core/pull/418 needs reviewing too

BanzaiMan:
@halkeye `travis-core` is getting deprecated, so that PR is replaced by other logic. I'll look at the PR and update it for the new logic.

halkeye:
Heh. That's what happens when a pr is over a year old :D

That's great. Let me know if you need anything else.

BanzaiMan:
For reference, https://github.com/travis-ci/travis-scheduler/pull/29 is the PR.

I ran a test here: https://staging.travis-ci.org/travis-ci/travis_staging_test/builds/507127

It ran the JWT addon: https://github.com/travis-ci/travis_staging_test/blob/ae7a834f9e7b7122f9bdd31df1fe024cc9305a2c/.travis.yml#L5-L7
which encrypted the value: `SAUCE_ACCESS_KEY=foobar`.

As you can see on https://staging.travis-ci.org/travis-ci/travis_staging_test/builds/507127#L135, the token is:

    eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJUcmF2aXMgQ0ksIEdtYkgiLCJzbHVnIjoidHJhdmlzLWNpL3RyYXZpc19zdGFnaW5nX3Rlc3QiLCJwdWxsLXJlcXVlc3QiOiIiLCJleHAiOjE0Njk3MjY1MjQsImlhdCI6MTQ2OTcyMTEyNH0.FDO_9iHOX5a1WcWUul-hshKYeJ2H7TIQFPbuA2mrcnk

Using https://jwt.io/, with the secret `foobar`, we see that the signature is correct.

json_web_tokens_-_jwt_io

BanzaiMan:
I see that travis-ci/travis-scheduler#29 needs more work, to always let `jwt` decrypt. (Which was a part of the old `travis-core` PR.) I'll work on this.

BanzaiMan:
Or, maybe no further work is necessary. In [this PR build](https://staging.travis-ci.org/travis-ci/travis_staging_test/builds/507135#L144), you can see that the PR receives a valid JWT.

json_web_tokens_-_jwt_io

halkeye:
Yea, if foobar was your original secured token, that looks good to me.

BanzaiMan:
All right, then. I'll merge this (and the scheduler PR) and deploy to production shortly. I recommend using test data and credentials while we work out the rough edges.

Thanks for the efforts and patience, everyone!

BanzaiMan:
Once we have good results in production, we'll merge the docs PR and announce it.

halkeye:
@BanzaiMan how will we know when it's in production to start testing?

BanzaiMan:
It's there now.

halkeye:
@BanzaiMan I just put in a support request, but could the abuse detected/gatekeeper stuff have any relation to this PR?

https://travis-ci.org/saucelabs-sample-test-frameworks/Java-TestNG-Selenium/requests

BanzaiMan:
@halkeye That should be lifted now.

halkeye:
@BanzaiMan Okay, looking at this test case.
it looks like the jwt stuff is running after install.

Should the function be changed to before_before_install (is that a thing)?

https://travis-ci.org/saucelabs-sample-test-frameworks/Java-TestNG-Selenium/jobs/148427092

BanzaiMan:
Yes, you can use `before_before_install` if it is the appropriate place to run the code.

BanzaiMan:
@halkeye Could you open a new issue in https://github.com/travis-ci/travis-ci to track further improvements (except code changes, of course)?

halkeye:
I updated this branch, but I can totally create a new issue if you want

halkeye:
@BanzaiMan new PR or new issue?

halkeye:
should I create a new PR for the change?

halkeye:
It wasn't picking up the new code and retesting, so just opened a new PR

PR #367, opened 2015-01-06
Add Elixir managment with kiex

BanzaiMan:
Assuming https://github.com/travis-ci/travis-cookbooks/pull/430, we source
the kiex env file to activate the desired version of Elixir during the build.

Here, as well as in the cookbooks PR, there is no safeguard for the user who
attempts to run Elixir while an unsupported Erlang version is active.


BanzaiMan:
Closing this in favor of #371.

PR #368, opened 2015-01-07
Directory caches

BanzaiMan:
Needs testing

### DIFF
--- a/lib/travis/build/script.rb
+++ b/lib/travis/build/script.rb
@@ -111,6 +111,11 @@ module Travis
         def disable_sudo
           apply :disable_sudo
         end
+
+        def config_env_vars
+          @config_env_vars ||= Build::Env::Config.new(data, config)
+          Array(@config_env_vars.data[:env])
+        end
     end
   end
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ cache: bundler
 
 sudo: false
 
-script: "bundle exec rspec spec"
+script: false # "bundle exec rspec spec"
 before_deploy: ruby script/build_s3_index_html.rb
 
 deploy:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ cache: bundler
 
 sudo: false
 
-script: false # "bundle exec rspec spec"
+script: "bundle exec rspec spec"
 before_deploy: ruby script/build_s3_index_html.rb
 
 deploy:
--- a/lib/travis/build/data.rb
+++ b/lib/travis/build/data.rb
@@ -13,7 +13,9 @@ module Travis
         apt:       false,
         bundler:   false,
         cocoapods: false,
-        composer:  false
+        composer:  false,
+        ccache:    false,
+        pip:       false
       }
 
       attr_reader :data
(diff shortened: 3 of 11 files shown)


BanzaiMan:
Tests were failing because I was overriding `install`. Duh.

BanzaiMan:
Mostly works, except that setting `CCACHE_DIR` in `.travis.yml` doesn't work for ccache. (Settings panel works.)

BanzaiMan:
The logic involving the `.travis.yml`-based env values needs some cleaning up. But it will be another PR or work.

BanzaiMan:
The pip cache doesn't need the `.travis.yml`-based env var logic, since we do not configure the cache location.

BanzaiMan:
Documentation updates will be added later.

BanzaiMan:
When you merge this, be sure to ping me, or make sure that the cleanup work is done as appropriate.

lib/travis/build/script.rb
```diff
@@ -111,6 +111,15 @@ def prepare
...
         end
+
+        def config_env_vars
+          @config_env_vars ||= Build::Env::Config.new(data, config)
+          Array(@config_env_vars.data[:env])
+        end
+
+        def has_env(name)
+          data.env_vars.any? { |v| v[:name] == name } || config_env_vars.any? {|v| v =~ Regexp.new("\\b#{name}=")}
+        end
```
BanzaiMan: This portion needs further work.

thedrow:
Please ensure that CCACHE_DISABLE is set to true by default unless the ccache bucket is specified in order to avoid caching when it's not desired.

BanzaiMan:
Why is the non-default value more desirable? The very presence of `cache` indicates that the user wants the cache. If they don't, they can delete the key.

thedrow:
Say you haven't specified ccache as one of your cache buckets.
Since ccache is installed by the cookbook gcc will still attempt to fetch the compiled objects from the cache and will cache them later on unless ```CCACHE_DISABLE``` is set to true.

When the ccache bucket is specified the envvar should not be set at all since that's the intended behavior.

Am I clearer now?

BanzaiMan:
@thedrow I still don't see your argument. You can set `$CCACHE_DISABLE` to `true` in `.travis.yml` (or in the script you execute). If the user pulls cache and gets tripped by it without setting `$CCACHE_DISABLE` to `true`, this is precisely what happens locally if the user ran `ccache` under the same condition.

thedrow:
I'm not talking about the case where the ccache bucket is **enabled** e.g.
```yaml
cache:
  - ccache
```

When the ccache bucket is disabled (or not sepcified) e.g.
```yaml
cache:
  - ccache: false
```
```CCACHE_DISABLE``` should be set to true because gcc will still use ccache in order to find cached objects and will cache the objects that were not available before.
These cached objects would be discarded in the next build since they will not being saved to s3.

Install ccache youself and install say uwsgi. You'll see that that ccache is activated automatically without any intervention by your side and that it will cache the compilation result for further installations. If you set ```CCACHE_DISABLE``` then ccache won't be used.

BanzaiMan:
@thedrow Thanks for the further clarification. I've incorporated that into this PR.

thedrow:
Thank you. LGTM unless you have something else to add.

PR #369, opened 2015-01-07
git clone: add an optional --quiet option

fxlb:
This can be turned on by setting (in the .travis.yml)
git:
  quiet: true

This change allows to avoid many useless lines in the logs:
remote:...
[...]
Receiving objects:...
[...]
Resolving deltas:...
[...]

### DIFF
--- a/lib/travis/build/git/clone.rb
+++ b/lib/travis/build/git/clone.rb
@@ -40,6 +40,7 @@ module Travis
           def clone_args
             args = "--depth=#{depth}"
             args << " --branch=#{branch}" unless data.ref
+	    args << " #{quiet_o}"
             args
           end
 
@@ -49,6 +50,14 @@ module Travis
 
           def branch
             data.branch.shellescape
+          end
+
+          def quiet_o
+            "--quiet" if quiet?
+          end
+
+          def quiet?
+            config[:git][:quiet]
           end
 
           def dir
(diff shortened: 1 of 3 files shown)


fxlb:
This PR follow the closed travis-ci/travis-build/pull/364. the `--quiet` is now optional, off by default to keep existing behavior.

Any way to test this change ?

lib/travis/build/git/clone.rb
```diff
@@ -40,6 +40,7 @@ def checkout
           def clone_args
             args = "--depth=#{depth}"
             args << " --branch=#{branch}" unless data.ref
+	    args << " #{quiet_o}"
```
joshk: why not remove the need for the `quiet_o` method and just add it similar to how the `--branch` is added?

also, looks like the indenting is a little off

fxlb: like `args << " --quiet" if quiet?` ?
(I see the bad <tab>)

fxlb: or `args << "--quiet" if config[:git][:quiet]` ?
(so no need of quiet?)

joshk: I prefer `args << " --quiet" if quiet?`

fxlb:
OK, a new a PR is coming. Closing.

PR #370, opened 2015-01-07
git clone: add an optional --quiet option

fxlb:
`[Patch v3]`
This can be turned on by setting (in the .travis.yml)
```
git:
  quiet: true
```
This change allows to avoid many useless lines in the logs:
```
remote:...
[...]
Receiving objects:...
[...]
Resolving deltas:...
[...]
```

### DIFF
--- a/lib/travis/build/git/clone.rb
+++ b/lib/travis/build/git/clone.rb
@@ -40,6 +40,7 @@ module Travis
           def clone_args
             args = "--depth=#{depth}"
             args << " --branch=#{branch}" unless data.ref
+            args << "--quiet" if quiet?
             args
           end
 
@@ -49,6 +50,10 @@ module Travis
 
           def branch
             data.branch.shellescape
+          end
+
+          def quiet?
+            config[:git][:quiet]
           end
 
           def dir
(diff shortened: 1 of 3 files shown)


lib/travis/build/git/clone.rb
```diff
@@ -40,6 +40,7 @@ def checkout
           def clone_args
             args = "--depth=#{depth}"
             args << " --branch=#{branch}" unless data.ref
+            args << "--quiet" if quiet?
```
joshk: you need to add a space before the `--`, also, you don't need to keep closing and opening new PRs :)

joshk:
I don't see any tests for this either.

@BanzaiMan could we look at adding a test for this and deploying it to staging?

lib/travis/build/git/clone.rb
fxlb: Any way to avoid 2 commits ?

fxlb:
@joshk, @BanzaiMan, Thank you !

BanzaiMan:
It's been deployed to both .org and .com.

fxlb:
diff on logs will be easier now !

joshk:
We also plan to fix this issue for when we archive logs, but this is a great stop gap. Thanks for the great PR!

PR #372, opened 2015-01-21
apt-get install addon

meatballhat:
specifically so that anybody running on container-based infrastructure has a formalized method for requesting a given package to be whitelisted.  This is a stop-gap until we have a suitable solution for allowing passwordless sudo within the container-based infrastructure.

### DIFF
--- a/lib/travis/build/addons/apt_packages.rb
+++ b/lib/travis/build/addons/apt_packages.rb
@@ -35,7 +35,7 @@ module Travis
           end
 
           def whitelist
-            @whitelist ||= ENV['TRAVIS_BUILD_APT_WHITELIST'].to_s.split(/\s*,\s*/)
+            @whitelist ||= ENV['TRAVIS_BUILD_APT_WHITELIST'].to_s.split(/\s*,\s*/).map(&:strip)
           end
       end
     end
--- a/lib/travis/build/addons/apt_packages.rb
+++ b/lib/travis/build/addons/apt_packages.rb
@@ -11,12 +11,23 @@ module Travis
             sh.echo "Installing APT Packages (BETA)", ansi: :yellow
 
             whitelisted = []
+            disallowed = []
+
             config.each do |package|
               if whitelist.include?(package)
                 whitelisted << package
               else
-                sh.echo "Ignoring unknown/disallowed package #{package.inspect}"
+                disallowed << package
               end
+            end
+
+            unless disallowed.empty?
+              sh.echo "Disallowing packages: #{disallowed.join(', ')}", ansi: :red
+              sh.echo 'If you require these packages, please submit them for ' \
+                      'review in a new issue:'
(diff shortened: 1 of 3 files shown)


lib/travis/build/addons/apt_packages.rb
```diff
@@ -0,0 +1,43 @@
...
+        end
+
+        private
+
+          def config
+            Array(super)
+          end
+
+          def whitelist
+            @whitelist ||= ENV['TRAVIS_BUILD_APT_WHITELIST'].to_s.split(/\s*,\s*/)
```
BanzaiMan: I don't like the comma-delimited list here. White spaces are not legal in apt package names, so why not just set `TRAVIS_BUILD_API_WHITELIST` to space-separated list? (Then split on `/\s+/`.)

meatballhat: @BanzaiMan sounds good to me :+1: :smiley_cat:

henrikhodne: With the danger of starting a bikeshed argument: One case for the commas is that whitespaces are a bit of a pain to enter on the console (you need to escape them), but that might not be an issue if we don't edit the variable on the console.

meatballhat: @BanzaiMan I thought of something, actually.  Having spaces in env vars means we always have to be careful with quoting, which is minor but definitely a pain.  For example:

``` bash
# hk env -a heroku-app-name
TRAVIS_BUILD_APT_WHITELIST=foo bar ham bones
```

... meaning I will always have to munge this output a la `s/=/="/;s/$/"/`.  Again, it's minor, but a pain :scream_cat:

Would you accept `:`-delimited?

meatballhat: trololo like minds

meatballhat:
Example build with server configuration of `TRAVIS_BUILD_APT_WHITELIST=git,curl`: https://staging.travis-ci.org/travis-repos/ruby-test-staging/builds/400373

lib/travis/build/addons/apt_packages.rb
```diff
@@ -0,0 +1,53 @@
...
+              sh.echo "Disallowing packages: #{disallowed.join(', ')}", ansi: :red
+              sh.echo 'If you require these packages, please submit them for ' \
+                      'review in a new issue:'
+              sh.echo '    https://github.com/travis-ci/travis-ci/issues/new' \
+                          "?title=APT+whitelist+request+for+#{disallowed.join(',+')}"
+            end
+
+            unless whitelisted.empty?
+              sh.export 'DEBIAN_FRONTEND', 'noninteractive', echo: true
+              sh.cmd "sudo -E apt-get -yq update", echo: true, timing: true
```
BanzaiMan: Can we make this quieter?

PR #375, opened 2015-01-30
Switch go version management to gimme

meatballhat:
This implementation aims to be independent of build env updates, although once gimme is shipped with the build image(s) we can then remove the presence check + download.  The disabling of gvm was required in order to have the correct go version activated, which I'm guessing will mean we may have to remove gvm entirely from the build env as well.

It's worth noting that the aliasing logic changes with this PR such that point releases such as `1.0` and `1.3` are now left alone rather than being "upgraded" to `1.0.3` and `1.3.3`, respectively.  The reasoning here is that such point releases are legitimate binary releases, whereas version `1.2` will still be "upgraded" to `1.2.2` as there is no `1.2` binary release available (at least in the well-known download locations).

Example build in staging: https://staging.travis-ci.org/travis-repos/go-test-staging/builds/405600

### DIFF
--- a/lib/travis/build/script/go.rb
+++ b/lib/travis/build/script/go.rb
@@ -14,7 +14,8 @@ module Travis
 
         def prepare
           super
-          sh.if '! command -v gimme' do
+          # TODO: remove this bit once we're shipping gimme via chef (?)
+          sh.if "! -x '#{HOME_DIR}/bin/gimme' && ! -x '/usr/local/bin/gimme'" do
             sh.cmd "curl -sLo #{HOME_DIR}/bin/gimme '#{gimme_url}'"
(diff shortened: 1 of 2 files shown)


joshk:
Does this allow for Go tip?

meatballhat:
yep, tip is explicitly handled in gimme

meatballhat:
@joshk specifically, here: https://github.com/meatballhat/gimme/blob/de1e10e1b6ab5a90239d5b0408859f82b88d85b3/gimme#L146

joshk:
HUZZAH! :+1:

meatballhat:
most recent staging build: https://staging.travis-ci.org/travis-repos/go-test-staging/builds/405657

BanzaiMan:
OMG!

PR #376, opened 2015-02-02
Lift restrictions on artifacts addon

meatballhat:
as all configuration may be specified as environment variables.


meatballhat:
example build: https://staging.travis-ci.org/travis-repos/go-test-staging/builds/405618

PR #378, opened 2015-02-03
Allow perl builds to continue even if switching versions fails

haarg:
Up until recently, if you listed a perl version in your build matrix that was not installed, the build would continue running.  This allowed code in the other phases to handle the issue in a more useful way than aborting immediately.  I have [a project](https://github.com/travis-perl/helpers) that is used by many perl projects which relies on this to compile the requested perl version, or sometimes download a pre-built version.

The latest update of the build environment has updated perlbrew, and the `perlbrew use $version` line now exits with an error if the version is not already available.  This has broken a good number of perl projects using my helper scripts.

I know that this is kind of a hack, but until I create an alternate solution, I would like to see the previous behavior restored.

### DIFF
--- a/lib/travis/build/script/perl.rb
+++ b/lib/travis/build/script/perl.rb
@@ -13,7 +13,7 @@ module Travis
 
         def setup
           super
-          sh.cmd "perlbrew use #{version}"
+          sh.cmd "perlbrew use #{version} || true"
         end
 
         def announce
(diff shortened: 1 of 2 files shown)


autarch:
+1 - this causes a lot of false failures for my Perl modules, since I like to test them all against bleadperl using @haarg's travis-perl helpers.

BanzaiMan:
https://github.com/travis-ci/travis-build/commit/e23a4ebe3501a18ca61ad7220ea1725a4f3c201d fixes this in a different way. This has been deployed.

Thanks for bringing this to our attention.

PR #380, opened 2015-02-06
no space in sh.fold in script/dart.rb

a14n:
_fold_ parameter should not have space because `sh.fold 'Content Shell dependencies install'` generates a `travis_fold:start:Content` in the detailed logs and the web page displaying the logs becomes insane.

### DIFF
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -9,7 +9,7 @@ module Travis
         def configure
           super
 
-          sh.fold 'Content Shell dependencies install' do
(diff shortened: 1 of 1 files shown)


a14n:
IMO it more an issue in https://github.com/travis-ci/travis-web where `fold` in logs should be handled in a different way. `travis_fold:end` should close the last `travis_fold:start` (it seems to be the first with the matching name).

joshk:
I think this change looks good and correct.

I don't believe we can use spaces in the fold names in a sane way.

PR #384, opened 2015-02-10
Run composer self-update when applicable

BanzaiMan:
Composer can flag itself "old" after 30 days, and it can
be quite noisy for the user.

Assuming that Composer doesn't break stuff unexpectedly, perhaps
we should guard against such aging.

Fixes https://github.com/travis-ci/travis-ci/issues/3154

### DIFF
--- a/lib/travis/build/script/php.rb
+++ b/lib/travis/build/script/php.rb
@@ -20,6 +20,11 @@ module Travis
         def setup
           super
           sh.cmd "phpenv global #{version}", assert: true
+          unless version == '5.2'
+            sh.if '-f composer.json' do
+              sh.cmd 'composer self-update', echo: false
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
:-1: This is a waste of resources.

GrahamCampbell:
That said, I'm not that :-1: on it tbh. :0

BanzaiMan:
Well, an alternative is to put an advisory if composer is too old.

GrahamCampbell:
Hmmm. Maybe we just need a way to push composer updates to the build environment on a daily basis. That would avoid all this extra resource usage.

BanzaiMan:
I'm going to close this. The warning has been documented.

PR #386, opened 2015-02-13
Add R support for travis-build.

craigcitro:
This adds initial R support into travis, including roughly all the bells and
whistles currently in use by [r-travis](https://github.com/craigcitro/r-travis) users.

For most users, their entire travis config will become `language: r`. This
seems ideal.

However, many users have more complex dependencies, either on new package
versions, alternate repositories, or system dependencies. For these users,
they can add packages to one of `apt_packages`, `r_packages`,
`r_binary_packages`, `bioc_packages`, or `r_github_packages`, as appropriate.
This will allow basically all existing users to move over by shuffling their
dependencies into the appropriate list above.

Last, we include a few configuration options to control how R packages are
built and checked (`r_build_args` and friends), as well as simple options for
installing common dependencies (`latex`, `pandoc`).

I have several questions, which I'll detail in comments below.


craigcitro:
PTAL @joshk and @roidrage -- this is the promised PR from travis-ci/travis-ci#1549. 

cc @hadley and @eddelbuettel, who are now on the hook with me for support. `;)`

craigcitro:
Questions/comments from my side:

1. This is the first ruby code I've ever written, which will surely show. Feel free to be merciless at telling me I don't know how to write ruby. `;)`

2. I feel like I should add more tests, but I don't know what the usual bar is. Feel free to ask for more.

3. I've added the `after_failure` method, but it didn't seem to show up whenever I ran `travis compile`. Is there something I'm missing about how that gets invoked?

4. Can we depend on `python-software-properties` being available in any travis ubuntu image? (Empirically, it is, but this script will fail if that ceases to be true.)

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
+module Travis
+  module Build
+    class Script
+      class R < Script
+        DEFAULTS = {
+          # Basic config options
+          cran: 'http://cran.rstudio.com',
+          warnings_are_errors: false,
```
hadley: I do wonder if we should make this true by default - I set it to true in all my repos.

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+          # Dependencies (installed in this order)
+          apt_packages: [],
+          r_binary_packages: [],
+          r_packages: [],
+          bioc_packages: [],
+          r_github_packages: [],
+          # Build/test options
+          r_build_args: '--no-build-vignettes --no-manual',
```
hadley: We can eliminate these '--no-build-vignettes --no-manual' if we can assume latex is installed, right?

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+          # Turn warnings into errors, if requested.
+          if config[:warnings_are_errors]
+            export_rcheck_dir
+            sh.if 'grep -q -R "${RCHECK_DIR}/**/*.00check" "WARNING"' do
+              sh.failure "Found warnings, treating as errors (as requested)."
+            end
+          end
+          
+          # Check revdeps, if requested.
+          if config[:r_check_revdep]
```
hadley: Should we be encouraging people do this on travis? It seems a bit heavy for a default to me.

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+          sh.cmd "Rscript -e '#{install_script}'"
+        end
+
+        def install_deps
+          setup_devtools
+          if not needs_bioc?
+            install_script = [
+              'library("devtools")',
+              "options(repos = c(CRAN = \"#{config[:cran]}\"))",
+              'install_deps(dependencies = TRUE)',
```
hadley: It might be slightly more elegant to skip attaching the package and just do `devtools::install_deps(dependencies = TRUE)`

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+          pkg_script = [
+            'library("devtools")',
+            'cat(paste0(as.package(".")$package, ".Rcheck"))',
+          ].join('; ')
+          sh.export 'RCHECK_DIR', "$(Rscript -e '#{pkg_script}')"
+        end
+        
+        def dump_logs
+          export_rcheck_dir
+          ['out', 'log', 'fail'].each do |ext|
```
hadley: This also needs to dump the install log in case of failure on installation

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+              ');',
+            ].join(' ')
+            sh.cmd "Rscript -e '#{bioc_install_script}'", retry: true
+          end
+          @bioc_installed = true
+        end
+
+        def setup_devtools
+          if not @devtools_installed
+            devtools_check = '!length(find.package("devtools", quiet = TRUE))'
```
hadley: A slightly nicer check is `!requireNamespace("devtools", quietly = TRUE)`

lib/travis/build/script/r.rb
eddelbuettel: I would not override the R default.  This is clearly a per-user setting to me.

lib/travis/build/script/r.rb
eddelbuettel: My Travis builds never install devtools. It worked so far without forcing this in, why change now?

BanzaiMan:
Super! We can coordinate time to test this in the near future. Also, please prepare an accompanying documentation PR.

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+          # TODO(craigcitro): python-software-properties?
+          sh.echo 'Installing R'
+          case config[:os]
+          when 'linux'
+            # Set up our CRAN mirror.
+            sh.cmd 'sudo add-apt-repository ' +
+                   "\"deb #{config[:cran]}/bin/linux/ubuntu " +
+                   "$(lsb_release -cs)/\""
+            sh.cmd 'sudo apt-key adv --keyserver keyserver.ubuntu.com ' +
+                   '--recv-keys E084DAB9'
```
CBarraford: should the keyserver and rec-keys should be constants?

craigcitro: they haven't changed for O(years) at least; @eddelbuettel is this something we should make more flexible?

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+            sh.cmd 'sudo installer -pkg "/tmp/R-latest.pkg" -target /'
+            sh.rm '/tmp/R-latest.pkg'
+
+          else
+            sh.failure "Operating system not supported: #{config[:os]}"
+          end
+
+          if needs_bioc?
+            setup_bioc
+          end
```
CBarraford: another way to write this in a one liner

```ruby
setup_bioc if needs_bioc?
```

CBarraford: both ways of course works, its up to you if you like the way one looks over the other.

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+          if config[:pandoc]
+            setup_pandoc
+          end
+        end
+
+        def announce
+          super
+
+          sh.cmd 'Rscript -e \'sessionInfo()\''
```
CBarraford: ruby has another cool way of doing strings that comes in handy sometimes. Instead of character escaping, you can use %q...

```ruby
sh.cmd %q[Rscript -e 'sessionInfo()']
```
http://en.wikibooks.org/wiki/Ruby_Programming/Syntax/Literals#The_.25_Notation

Or you could do (to avoid char escaping)
```ruby
sh.cmd "Rscript -e 'sessionInfo()'"
```

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+          tarball_script = [
+            'library("devtools")',
+            'pkg <- as.package(".")',
+            'cat(paste0(pkg$package, "_", pkg$version, ".tar.gz"))',
+          ].join('; ')
+          sh.export 'PKG_TARBALL', "$(Rscript -e '#{tarball_script}')"
+
+          # Test the package
+          sh.echo 'Testing with: R CMD check "${PKG_TARBALL}" ' +
+                  "#{config[:r_check_args]}"
```
CBarraford: you can make this one string (without the `+`)

```ruby
sh.echo %Q[Testing with: RCMD check "${PKG_TARBALL}" #{config[:r_check_args]}]
```

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+            sh.echo "Checking reverse dependencies"
+            revdep_script = [
+              'library("devtools");',
+              'res <- revdep_check();',
+              'if (length(res) > 0) {',
+              ' revdep_check_summary(res);',
+              ' revdep_check_save_logs(res);',
+              ' q(status = 1, save = "no");',
+              '}',
+            ].join(' ')
```
CBarraford: would it improve the readability if you use a multiline string?

```ruby
revdep_script = <<-eos
insert;
r code;
here;
eos
```

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+
+        def script
+          # Build the package
+          sh.echo "Building with: R CMD build ${R_BUILD_ARGS}"
+          sh.cmd "R CMD build #{config[:r_build_args]} ."
+          tarball_script = [
+            'library("devtools")',
+            'pkg <- as.package(".")',
+            'cat(paste0(pkg$package, "_", pkg$version, ".tar.gz"))',
+          ].join('; ')
```
CBarraford: would it improve the readability if you use a multiline string?

```ruby
revdep_script = <<-eos
insert;
r code;
here;
eos

File: lib/travis/build/script/r.rb [hunk 13]
```diff
@@ -0,0 +1,368 @@
...
+
+        def after_failure
+          dump_logs
+          super
+        end
+
+        private
+
+        def needs_bioc?
+          config[:bioc_required] or not config[:bioc_packages].empty?
```
CBarraford: is `or` the correct thing here, or `||`? I'm not sure, just asking.
http://estebanpastorino.com/2013/09/12/quick-ruby-tip-double-pipe-vs-or/

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+
+        def apt_install(packages)
+          pkg_arg = packages.join(' ')
+          sh.echo "Installing apt packages: #{packages.join(', ')}"
+          sh.cmd "sudo apt-get install #{pkg_arg}", retry: true
+        end
+
+        def bioc_install(packages)
+          if not needs_bioc?
+            return
```
CBarraford: another way to write this is...

```ruby
return unless needs_bioc?
```

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+              'echo ">>> Filename: ${name} <<<";',
+              'cat ${name};',
+              'done',
+            ].join(' ')
+            sh.cmd cmd
+          end
+        end
+        
+        def setup_bioc
+          if not @bioc_installed
```
CBarraford: i believe its more ruby-y to use `unless` here instead of `if not`

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+            bioc_install_script = [
+              "source(\"#{config[:bioc]}\");",
+              'tryCatch(',
+              " useDevel(#{as_r_boolean(config[:bioc_use_devel])}),",
+              ' error=function(e) {if (!grepl("already in use", e$message)) {e}}',
+              ');',
+            ].join(' ')
+            sh.cmd "Rscript -e '#{bioc_install_script}'", retry: true
+          end
+          @bioc_installed = true
```
CBarraford: what is the cmd fails, should we still mark bioc as installed?

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+        end
+
+        def setup_devtools
+          if not @devtools_installed
```
CBarraford: you could also use the `unless` statement here

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+            devtools_check = '!length(find.package("devtools", quiet = TRUE))'
+            devtools_install = 'install.packages(c("devtools"), ' +
+                               "repos=\"#{config[:cran]}\")"
+            sh.cmd "Rscript -e 'if (#{devtools_check}) #{devtools_install}'"
+          end
+          @devtools_installed = true
```
CBarraford: if the cmd errors out, should we still mark devtools as installed?

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+        end
+
+        def setup_latex
+          case config[:os]
+          when 'linux'
+            # We add a backports PPA for more recent TeX packages.
+            sh.cmd 'sudo add-apt-repository -y "ppa:texlive-backports/ppa"'
+
+            sh.cmd 'sudo apt-get install --no-install-recommends ' +
```
CBarraford: it would be nice if you had the list of packages to install as an array to ease of maintainability.

```ruby
%w[ lmodern texinfo texlive-base texlive-extra-util etc etc etc ]
```

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+                   'lmodern texinfo texlive-base texlive-extra-utils ' +
+                   'texlive-fonts-extra texlive-fonts-recommended ' +
+                   'texlive-generic-recommended texlive-latex-base ' +
+                   'texlive-latex-extra texlive-latex-recommended ',
+                   retry: true
+          when 'mac'
+            # We use mactex-basic due to disk space constraints.
+            mactex = 'mactex-basic.pkg'
+            sh.cmd 'wget http://ctan.math.utah.edu/ctan/tex-archive/systems/' +
```
CBarraford: this feels like a single point of failure. Should we have a list of mirrors to try in case one is unavailable?

craigcitro:
thanks all -- will handle all the feedback so far tonight. 

@CBarraford thanks a TON -- soon my ruby will look more like ruby and less like python `:D`

@BanzaiMan happy to update the docs -- which repo do i fork?

PeteHaitch:
Super minor point, but `language: R` seems more natural to me than `language: r`. Is it case sensitive?

simi:
@PeteHaitch probably not - https://github.com/travis-ci/travis-yaml/blob/master/lib/travis/yaml/nodes/language.rb#L4

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,368 @@
...
+          r_check_cran_incoming: false,
+          r_check_revdep: false,
+          # Heavy dependencies
+          latex: false,
+          pandoc: false,
+          pandoc_version: '1.12.4.2',
+          # Bioconductor
+          bioc: 'http://bioconductor.org/biocLite.R',
+          bioc_required: false,
+          bioc_use_devel: true,
```
PeteHaitch: I think the default of `bioc_use_devel: true` might require some thought. Bioconductor's release cycle means that, for example, the current Bioc-devel can only be used with R 3.2 and not with R 3.1.x (or earlier). 

E.g. on R 3.1.2:

```r
> source("http://bioconductor.org/biocLite.R")
Installing package into ‘/Users/Peter/Desktop/bioc-devel_3.1.2/packrat/lib/x86_64-apple-darwin13.4.0/3.1.2’
(as ‘lib’ is unspecified)
trying URL 'http://www.bioconductor.org/packages/3.0/bioc/bin/macosx/mavericks/contrib/3.1/BiocInstaller_1.16.1.tgz'
Content type 'application/x-gzip' length 49535 bytes (48 Kb)
opened URL
==================================================
downloaded 48 Kb


The downloaded binary packages are in
	/var/folders/ns/tvfcs46s5b1f0k9bbl_q2k580000gn/T//RtmpK2ZqOJ/downloaded_packages
Bioconductor version 3.0 (BiocInstaller 1.16.1), ?biocLite for help
> useDevel()
Error: 'devel' version requires a more recent R
> sessionInfo()
R version 3.1.2 (2014-10-31)
Platform: x86_64-apple-darwin13.4.0 (64-bit)

locale:
[1] en_AU.UTF-8/en_AU.UTF-8/en_AU.UTF-8/C/en_AU.UTF-8/en_AU.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] BiocInstaller_1.16.1

loaded via a namespace (and not attached):
[1] packrat_0.4.1-1 tools_3.1.2 
`

Versus on R 3.2.0 (note also that you __can't__ use Bioc-release on 3.2 but only Bioc-devel)

```r
> source("http://bioconductor.org/biocLite.R")
Installing package into ‘/Users/Peter/Desktop/bioc-devel_3.2.0/packrat/lib/x86_64-apple-darwin13.4.0/3.2.0’
(as ‘lib’ is unspecified)
trying URL 'http://www.bioconductor.org/packages/3.1/bioc/bin/macosx/mavericks/contrib/3.2/BiocInstaller_1.17.5.tgz'
Content type 'application/x-gzip' length 51621 bytes (50 KB)
==================================================
downloaded 50 KB


The downloaded binary packages are in
	/var/folders/ns/tvfcs46s5b1f0k9bbl_q2k580000gn/T//Rtmp85NfXI/downloaded_packages
Bioconductor version 3.1 (BiocInstaller 1.17.5), ?biocLite for help
> 
> useDevel()
Error: 'devel' version already in use
> useDevel(FALSE)
Error: 'devel' version cannot be down-graded with this version of R
> sessionInfo()
R Under development (unstable) (2015-02-12 r67812)
Platform: x86_64-apple-darwin13.4.0 (64-bit)
Running under: OS X 10.10.2 (Yosemite)

locale:
[1] en_AU.UTF-8/en_AU.UTF-8/en_AU.UTF-8/C/en_AU.UTF-8/en_AU.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] BiocInstaller_1.17.5

loaded via a namespace (and not attached):
[1] tools_3.2.0     packrat_0.4.1-1
`

Bioconductor developers are encouraged to develop against Bioc-devel rather than Bioc-release, so I guess the issues facing Bioconductor developers wanting to use r-travis are:

1. Will it be possible to use R-devel (3.2, in this case) with r-travis? I'm currently using r-builder by @gaborcsardi (https://github.com/metacran/r-builder) in place of r-travis, but it would seem nice to consolidate this within the "official" Travis branch.
2. Can the `bioc_use_devel` flag be sanity checked for compability against the version of R requested by the user?

File: lib/travis/build/script/r.rb (hunk 3 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: well, the *default* is still off -- it's just a question of whether we want to make it available, or force people to do it by hand.

File: lib/travis/build/script/r.rb (hunk 7 above)
Comment (eddelbuettel) [inline lib/travis/build/script/r.rb]: Until the Ubuntu / Canonical Launchpad.net changes. no need.  Constant is fine.

Just this week I reactivated a PPA created in 2009.   Seems pretty "constant" to me.

File: lib/travis/build/script/r.rb (hunk 1 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: i'm actually torn here -- i'm with @eddelbuettel in principle, but this is like `-Werror` in C -- wouldn't the world be a better place if K&R had just changed that default in the 70s? `;)`

File: lib/travis/build/script/r.rb (hunk 2 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: i had left latex optional, for time savings; i'm starting to think i should have it always install, in particular for things like this (since we really *do* want to check the vignettes and manual by default), and that's one more reason to get latex baked into the base image.

File: lib/travis/build/script/r.rb (hunk 4 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: @hadley -- ah, nice. done. that actually shortened most of my inline scripts (save the revdeps one, which needs several functions from devtools).

@eddelbuettel you can still skip it -- if you provide your own `install` step in the `.travis.yml`, you'll avoid calling this one, and `devtools` won't get installed. that said, i suspect you're the **only** person who prefers to explicitly list their dependencies instead of just calling `install_deps`. :smirk:

File: lib/travis/build/script/r.rb (hunk 5 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: good point -- do you happen to know the file extension offhand?

File: lib/travis/build/script/r.rb (hunk 6 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: nice! done.

File: lib/travis/build/script/r.rb (hunk 8 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: i'm updating them to use it where it "feels right" -- i don't have any real feeling for ruby, so if it jumped out at at least one person, it's nice to keep it a little terser. :smile:

File: lib/travis/build/script/r.rb (hunk 9 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: so i read a bunch about `%q`/`%Q` and friends, and originally tried mixing them in to avoid escaping quotes. it mostly felt a little harder to read, since i kept going back and forth -- maybe that was just me.

just to check: my understanding is that single-quoted strings get no interpolation, whereas double-quoted do -- so i did single quotes here (even though it meant escaping), since i knew i definitely didn't want the contents to be tweaked at all. (it's a short enough string that it doesn't matter, but as a general rule.)

File: lib/travis/build/script/r.rb (hunk 10 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: i kept it on two lines to stay *vaguely* near 80 characters. i don't know that anyone has an 80 char limit, but working at google has apparently convinced me that long lines and trailing whitespace are sins of the highest order. (i think the conventional name here is "stockholm syndrome"?)

File: lib/travis/build/script/r.rb (hunk 1 above)
Comment (eddelbuettel) [inline lib/travis/build/script/r.rb]: Please do not make `warning` equal `error`. There _are_ false positives in `R CMD check`.  This would make it useless.  IIRC correctly I have at least one package which _permanently_ exhibits one warning.  Wrongly.

On the margin I think I'd fork, or stick to r-travis which I can modify.  

If you must, make it an option that defaults to `FALSE`.

File: lib/travis/build/script/r.rb (hunk 11 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: IIRC, that string will ultimately become something like

    insert;      r code;      here;

as a value, right? it's not super important, but it seemed like it would read nicer in the generated script without the extra spaces ...

File: lib/travis/build/script/r.rb (hunk 12 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: this one got one line shorter thanks to hadley; it's *probably* short enough to be OK now?

File: lib/travis/build/script/r.rb (hunk 13 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: hah! i had no idea they were different in ruby. (in fact, i didn't even know there was a `||` operator -- i typed `or` and emacs highlighted it, so i figured that was the same as python.)

i think i'm safe here, but having read that article, i switched to `||` anyway.

File: lib/travis/build/script/r.rb (hunk 14 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: nice!

File: lib/travis/build/script/r.rb (hunk 15 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: nice! i didn't even know about `unless`.

i assume the convention is to keep with `if` in the case you actually have two branches? (i'm basing this on emacs lisp.)

File: lib/travis/build/script/r.rb (hunk 16 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: i think we're safe in this case -- if bioc fails to install, we'll exit the build, right? (note that the `retry` is indeed deeper in.)

File: lib/travis/build/script/r.rb (hunk 17 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: done!

File: lib/travis/build/script/r.rb (hunk 18 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: same as above -- but i'd forgotten the `retry`!

File: lib/travis/build/script/r.rb (hunk 19 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: so that is definitely a good idea -- but how do i make it work with a multi-line list? (i'm guessing that the obvious thing doesn't work based solely on emacs not indenting properly, so let me know if i just need to tell emacs i know better.)

File: lib/travis/build/script/r.rb (hunk 20 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: good point -- switched to `mirrors.ctan.org`, which will multiplex between mirrors as needed (or at least claims to). added a TODO to double-check.

File: lib/travis/build/script/r.rb (hunk 21 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: you're totally right about the default: i just copied whatever was in r-travis.

i switched it, because we don't support `R-devel`, which is a requirement for bioc-devel. i do have plans in the works for making `R-devel` available, but it'll be a follow-on to this PR (since a few other pieces need to fall into place to make it easier).

**Comment by craigcitro** [comment]
thanks for all the feedback so far -- i think i've hit everything, modulo deciding what to do about the `warnings_are_errors` default (which i'm leaning towards switching).

also, @BanzaiMan -- i assume https://github.com/travis-ci/docs-travis-ci-com is the docs repo i'm looking for?

File: lib/travis/build/script/r.rb (hunk 1 above)
Comment (craigcitro) [inline lib/travis/build/script/r.rb]: ahh -- i didn't realize that there were well-known false positives. so keep it `false`, but make sure to prominently feature it in the docs for folks who aren't regularly bitten by it?

File: lib/travis/build/script/r.rb (hunk 19 above)
Comment (CBarraford) [inline lib/travis/build/script/r.rb]: the `%w` allows you to have multiline lists. for example....

```ruby
package_to_install = %w[
  lmodern texinfo texlive-base texlive-extra-utils
  texlive-fonts-extra texlive-fonts-recommended
  texlive-generic-recommended texlive-latex-base
  texlive-latex-extra texlive-latex-recommended
]

sh.cmd "sudo apt-get install --no-install-recommends #{package_to_install.join(' ')}", retry : true
```

craigcitro: score -- done.

craigcitro:
@BanzaiMan first pass at a docs PR is out: travis-ci/docs-travis-ci-com#199

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,380 @@
...
+          bioc_packages: [],
+          r_github_packages: [],
+          # Build/test options
+          r_build_args: '',
+          r_check_args: '--as-cran',
+          r_check_cran_incoming: false,
+          r_check_revdep: false,
+          # Heavy dependencies
+          pandoc: false,
+          pandoc_version: '1.12.4.2',
```
craigcitro: Note to self: update this to `1.13.1` to fix issues with templates being added to the binary.

BanzaiMan:
I plan to push this to staging some time tomorrow (say 14:00 UTC). If you want to test it with your repo, go to https://staging.travis-ci.org/profile and enable your repo to run on staging. Thanks.

lib/travis/build/script/r.rb
craigcitro: done.

BanzaiMan:
@27ec640 has been deployed to staging.

BanzaiMan:
Is anyone testing? ;-)

craigcitro:
i'm planning on testing tonight PST (i mostly do this in the evenings :wink:) -- but i can give it at least one quick spin in a few minutes.

eddelbuettel:
Is there a HOWTO page / email / note / $ANYTHING on migrating from r-travis to the staging version?

gaborcsardi:
I am trying, but the VM seems to be stuck, it has not started for two hours: https://staging.travis-ci.org/gaborcsardi/crayon/builds

craigcitro:
@eddelbuettel not yet -- once i know it's working, i'm planning on writing a big note and slapping it at the top of craigcitro/r-travis, and then we can get the word out on the twitters and whatnot once the change is live.

craigcitro:
@BanzaiMan does travis-staging work with multi-language repos?

craigcitro:
i'm in the same boat as @gaborcsardi -- https://staging.travis-ci.org/craigcitro/r-travis/builds/428283

lib/travis/build/script/r.rb
jimhester: You can also avoid having to do the string concatenation by escaping the end of the line.  Ruby automatically combines adjacent quotations together.

```ruby
sh.echo 'Testing with: R CMD check "${PKG_TARBALL}" '\
        "#{config[:r_check_args]}"
```

This can even be done within a string, so you could do
```ruby
sh.echo %Q[Testing with: R CMD check "${PKG_TARBALL}" \
#{config[:r_check_args]}]
```
The downside to the second approach is you can't indent the second line however.

BanzaiMan:
Hi, everyone. Sorry for the delay in response. We are having problems with the staging environment. I'll look into this in more detail tomorrow.

gaborcsardi:
OK, things are running now: https://staging.travis-ci.org/gaborcsardi/crayon/builds/428277

Looks like pandoc needs to be relocated, or installed in the vm image itself.

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,384 @@
...
+
+          # TODO(craigcitro): Confirm that these do, in fact, print as
+          # green. (They're yellow under vagrant.)
+          sh.echo 'R for Travis-CI is not officially supported, ' +
+                  'but is community maintained.', ansi: :green
+          sh.echo 'Please file any issues using the following link',
+                  ansi: :green
+          sh.echo '  https://github.com/travis-ci/travis-ci/issues' +
+                  '/new?labels=community:r', ansi: :green
+          sh.echo 'and mention \`@craigcitro\`, \`@eddelbuettel\` and ' +
```
hadley: Maybe this should be: "And include `cc @craigcitro, @eddelbuettel, @hadley` in the issue. That way people can copy and paste more easily

lib/travis/build/script/r.rb
```diff
@@ -0,0 +1,384 @@
...
+
+          sh.cmd 'Rscript -e \'sessionInfo()\''
+          sh.echo ''
+        end
+
+        def install
+          super
+
+          # Install any declared packages
+          apt_install config[:apt_packages]
```
BanzaiMan: Use of `sudo` here will break when the job is run on [containers](http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/). If you know the packages to install ahead of time, you should install them up front before `sudo` is disabled. See https://github.com/travis-ci/travis-build/blob/4f0ba78d1ef378aec7dcf09704eadbfbc26d7965/lib/travis/build/script/php.rb for an example (it installs hhvm nightly package).

craigcitro: yes -- we know using sudo is not desirable, and is going to prevent container use. the recommendation [we got before](https://github.com/travis-ci/travis-ci/issues/1549#issuecomment-72396184) was to start here and remove `sudo` use in the future.

the main issue is that we don't know the *complete* list of packages up front -- we install user-specified R dependencies via `apt-get` to speed up install time (cf `r_binary_install` below).

BanzaiMan: Copy. As I mentioned in the docs PR, this restriction should be documented.

lib/travis/build/script/r.rb
craigcitro: good call. done.

lib/travis/build/script/r.rb
craigcitro: makes sense, will do.

relatedly, though, if there is a possibility of having a base image (so that we didn't have to spend time installing latex and friends, for time and bandwidth savings), i think a large number of R users *could* move to using containers (and just do source installs).

craigcitro:
i believe i've fixed the `pandoc` install issue -- i was creating the path in ruby, not in the generated shell script. oops. :wink:

craigcitro:
@BanzaiMan just let me know when you deploy a new version and we can give it another round of tests. thanks hiro!

BanzaiMan:
Hello. @b35998a has been deployed to staging. Thanks.

craigcitro:
https://staging.travis-ci.org/craigcitro/r-travis/builds/428413 seems to have stalled, but i'll check back in a bit ...

BanzaiMan:
Sorry. staging appears to be hosed again. I'll look into it.

craigcitro:
score! spotted another easy-to-fix one, will take care of it later tonight.

staging is maybe wedged again? https://staging.travis-ci.org/craigcitro/r-travis/builds/428427

craigcitro:
@BanzaiMan i've pushed a few more commits to fix issues found so far; with these, we'll have at least 5-6 complicated R projects that can run successfully on the bits in staging. woohoo!

is there a way to test OSX builds?

BanzaiMan:
@b3a7cf5 has been deployed to staging.

craigcitro:
fantastic @BanzaiMan -- I now have several working repositories on staging, exercising a bunch of different options.

i'm not sure what all you want to do as part of the review -- but we're now in pretty good shape.

* i'm planning to write a few more bits in the docs (and include a pointer to an r-travis -> native support page)
* i'd love to test mac builds (if possible) -- at the very least, is there a way to force `travis compile` to set `os` to `osx`?
* we can get a handful of folks (such as @gaborcsardi and @hadley already in the thread) to do more testing on staging.

i'll get the first two rolling, but outside that, i think the code's in good shape at this point.

BanzaiMan:
@craigcitro Fantastic. Right now, we do not have a staging worker running for the Mac. So we can't test it. We are a little short on the Mac capacity on the .org side, so we may not be able accommodate new requests for a while.

I plan to write a blog post and deploy it on Thursday around 15:00 UTC.

craigcitro:
w00t! i'm stoked to see this go live.

@BanzaiMan do you want me to squash all the commits in both of these PRs?

BanzaiMan:
@craigcitro There is no need to squash them.

BanzaiMan:
Thanks for the patience. This has been deployed to .org and .com.

craigcitro:
no problem at all -- thanks for all the help @BanzaiMan!

craigcitro:
@BanzaiMan can you create the "community:r" label in travis-ci/travis-ci?

jimhester:
Would it be possible to set `_R_CHECK_CRAN_INCOMING_=false` as an environment variable in these builds by default?  Doing so turns off the `checking CRAN incoming feasibility` check which is unnecessary except for the actual CRAN submission, increases the load on the CRAN servers handling the extra requests and takes ~10 seconds extra time on average.

This option was defaulted to `FALSE` previously in r-travis https://github.com/craigcitro/r-travis/blame/bb0483842206095b0c8ad6eddadefc77242c3152/scripts/travis-tool.sh#L277

I am not sure where else to put this as issues seem to be turned off for this repo.

eddelbuettel:
:+1: on not testing CRAN incoming.

BanzaiMan:
This thread is getting long. The support as this PR was intended to implement is deployed. So if there is anything else to discuss, please open a new issue. I'm closing this.

PR #387, opened 2015-02-16
opt-in for content_shell in dart builds

a14n:
Address https://github.com/travis-ci/travis-ci/issues/3215

- Install of `content_shell` is now an opt-in. `with_content_shell: true` should be defined in `.travis.yml` to activate browser tests. 
- Install of `test_runner` is now done in the script section and will be skipped when the user has a custom build.

### DIFF
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -3,25 +3,28 @@ module Travis
     class Script
       class Dart < Script
         DEFAULTS = {
-          :dart => 'stable'
+          dart: 'stable',
+          with_content_shell: 'false'
         }
 
         def configure
           super
 
-          sh.fold 'content_shell_dependencies_install' do
-            sh.echo 'Installing Content Shell dependencies', ansi: :yellow
+          if with_content_shell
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/dart.rb
```diff
@@ -3,25 +3,28 @@ module Build
     class Script
       class Dart < Script
         DEFAULTS = {
-          :dart => 'stable'
+          dart: 'stable',
+          with_content_shell: 'false'
```
simi: What about to rename it to `content_shell` only?

a14n: I used the `with_` prefix because it implies a _bool_ value.

kevmoo:
Doesn't this break existing folks?

We should be careful about changing the default behavior – at least without a lot of public discussion.

Just a thought for next time...

a14n:
Yep. But we hadn't communicate about the activation of browser tests and they were not documented. Actually it was just a bug that browser's tests were executed ;p

kevmoo:
@a14n Sounds good to me :smile: 

I just sent a branch up to travis with the new config. Also playing with `sudo: false`

PR #391, opened 2015-02-19
Attempt to make rust work again

brson:
The format of the rust installer changed recently and `language: rust` is now broken. This is my *attempt* to fix the problem by running the installer to place the Rust compiler in the ~/rust directory.

I do not know if it works correctly. The existing test case does not seem to confirm that it creates a working compiler and I don't understand Ruby enough to know how to add such a test case.

### DIFF
--- a/lib/travis/build/script/rust.rb
+++ b/lib/travis/build/script/rust.rb
@@ -7,11 +7,6 @@ module Travis
           linux: 'https://static.rust-lang.org/dist/rust-%s-x86_64-unknown-linux-gnu.tar.gz'
         }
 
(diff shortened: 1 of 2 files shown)


MartinNowak:
You can generate a shell script for a .travis.yml using `travis compile`. If you run that in a virtualbox you can test your script locally.

brson:
@MartinNowak thanks for the tip. I've tested manually and think it is working.

BanzaiMan:
This has been deployed to .org and .com. Please test it. (Re-running the builds should suffice.)

Thank you for your contribution!

PR #394, opened 2015-02-28
Add basic version switching for mono. Fixes travis-ci/travis-ci#3018.

Joshua-Anderson:
Can be used like this:

```
language: csharp
mono:
  - latest
  - "3.12.0"
```

Please scrutinize this regexp ` /^(\d{1}\.)\d{1,2}\.\d{1,2}$/` very carefully, because I'm no regexp master and I don't want to introduce a command injection vulnerability.

### DIFF
--- a/lib/travis/build/script/csharp.rb
+++ b/lib/travis/build/script/csharp.rb
@@ -8,8 +8,10 @@ module Travis
     class Script
       class Csharp < Script
         DEFAULTS = {
-          csharp: 'mono',
+          mono: 'latest',
         }
+
+        MONO_VERSION_REGEXP = /^(\d{1})\.(\d{1,2})\.\d{1,2}$/
 
         def configure
           super
@@ -20,16 +22,33 @@ module Travis
 
 
           sh.fold('mono-install') do
-            if config[:csharp] == 'mono'
+            if mono_version_valid?
               sh.echo 'Installing Mono', ansi: :yellow
 
-              sh.cmd 'sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF', echo: false
-              sh.cmd "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian wheezy main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", echo: false
-              sh.cmd "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian wheezy-libtiff-compat main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", echo: false
-              sh.cmd 'sudo apt-get update -qq', timing: true
-              sh.cmd 'sudo apt-get install -qq mono-complete referenceassemblies-pcl nuget mono-vbnc fsharp', timing: true
+              if is_mono_2_10_8
+                sh.cmd 'sudo apt-get install -qq mono-complete mono-vbnc', timing: true, assert: true
+              elsif is_mono_3_2_8
+                sh.cmd 'sudo apt-add-repository ppa:directhex/ppa -y', assert: true # Official ppa of the mono debian maintainer
+                sh.cmd 'sudo apt-get update -qq', timing: true, assert: true
+                sh.cmd 'sudo apt-get install -qq mono-complete mono-vbnc fsharp', timing: true, assert: true
+              else
+                sh.cmd 'sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF', echo: false, assert: true
+                sh.cmd "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian #{mono_repo} main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", echo: false, assert: true
+                sh.cmd "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian wheezy-libtiff-compat main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", echo: false, assert: true
+                sh.cmd 'sudo apt-get update -qq', timing: true, assert: true
+                sh.cmd "sudo apt-get install -qq mono-complete mono-vbnc fsharp nuget #{'referenceassemblies-pcl' if !is_mono_3_8_0}", timing: true, assert: true
+                                                                                      # PCL Assemblies only supported on mono 3.10 and greater
+              end
+
               sh.cmd 'mozroots --import --sync --quiet', timing: true
             end
+          end
+        end
+
+        def setup
+          unless mono_version_valid?
+            sh.echo "\"#{config[:mono]}\" is not a valid version of mono.", ansi: :red
+            sh.cmd 'false', echo: false, timing: false
           end
         end
 
@@ -48,16 +67,46 @@ module Travis
         end
 
         def install
-          sh.cmd "nuget restore #{config[:solution]}", retry: true if config[:solution]
+          sh.cmd "nuget restore #{config[:solution]}", retry: true if config[:solution] && !is_mono_2_10_8 && !is_mono_3_2_8
         end
 
         def script
           if config[:solution]
-            sh.cmd "xbuild /p:Configuration=Release #{config[:solution]}", timing: true if config[:solution]
+            sh.cmd "xbuild /p:Configuration=Release #{config[:solution]}", timing: true
           else
             sh.echo 'No solution or script defined, exiting', ansi: :red
             sh.cmd 'false', echo: false, timing: false
           end
+        end
+
+        def mono_repo
+          if config[:mono] == 'latest'
+            'wheezy'
+          else
+            'wheezy/snapshots/' << config[:mono]
+          end
+        end
+
+        def mono_version_valid?
+          return true if config[:mono] == 'latest'
+          return false unless MONO_VERSION_REGEXP === config[:mono]
+
+          return false if MONO_VERSION_REGEXP.match(config[:mono])[1] == '2' && !is_mono_2_10_8
+          return false if MONO_VERSION_REGEXP.match(config[:mono])[1].to_i < 2
+
+          true
+        end
+
+        def is_mono_2_10_8
+          config[:mono] == '2.10.8'
+        end
+
+        def is_mono_3_2_8
+          config[:mono] == '3.2.8'
+        end
+
+        def is_mono_3_8_0
+          config[:mono] == '3.8.0'
         end
       end
     end
--- a/spec/build/script/csharp_spec.rb
+++ b/spec/build/script/csharp_spec.rb
@@ -13,15 +13,61 @@ describe Travis::Build::Script::Csharp, :sexp do
 
   describe 'configure' do
     it 'sets up package repository' do
-      should include_sexp [:cmd, 'sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF']
-      should include_sexp [:cmd, "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian wheezy main' >> /etc/apt/sources.list.d/mono-xamarin.list\""]
-      should include_sexp [:cmd, "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian wheezy-libtiff-compat main' >> /etc/apt/sources.list.d/mono-xamarin.list\""]
-      should include_sexp [:cmd, 'sudo apt-get update -qq', timing: true]
+      should include_sexp [:cmd, 'sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF', assert: true]
+      should include_sexp [:cmd, "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian wheezy main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", assert: true]
+      should include_sexp [:cmd, "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian wheezy-libtiff-compat main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", assert: true]
+      should include_sexp [:cmd, 'sudo apt-get update -qq', timing: true, assert: true]
     end
 
     it 'installs mono' do
-      should include_sexp [:cmd, 'sudo apt-get install -qq mono-complete referenceassemblies-pcl nuget mono-vbnc fsharp', timing: true]
+      should include_sexp [:cmd, 'sudo apt-get install -qq mono-complete mono-vbnc fsharp nuget referenceassemblies-pcl', timing: true, assert: true]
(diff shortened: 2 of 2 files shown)


lib/travis/build/script/csharp.rb
```diff
@@ -59,6 +68,24 @@ def script
...
+            if MONO_VERSION_REGEXP === config[:mono]
+              'wheezy/snapshots/' << config[:mono]
+            else
+              'wheezy'
+            end
+          end
+        end
+
+        def mono_version_valid?
+          if MONO_VERSION_REGEXP === config[:mono] || config[:mono] == 'latest'
```
akoeplinger: nit: switch the two sides of the expression, "latest" should be the common case and it'll allow short-circuiting the check.

lib/travis/build/script/csharp.rb
```diff
@@ -8,9 +8,11 @@ module Build
     class Script
       class Csharp < Script
         DEFAULTS = {
-          csharp: 'mono',
+          mono: 'latest',
         }
 
+        MONO_VERSION_REGEXP = /^(\d{1}\.)(\d{1,2}\.)(\d{1,2})$/
```
akoeplinger: Put the `\.` outside the capturing group, so we only get the number instead of number+dot.

spec/build/script/csharp_spec.rb
```diff
@@ -25,6 +25,22 @@
     end
   end
 
+  describe 'version switching' do
+    it 'throws a error with a invalid version' do
+      data[:config][:mono] = 'foo'
+      should include_sexp [:echo, 'foo is not a valid version of mono.', {:ansi=>:red}]
+    end
```
akoeplinger: might make sense to add a test that uses an invalid, but number-like version, e.g. 99.1000.1

spec/build/script/csharp_spec.rb
```diff
@@ -25,6 +25,22 @@
     end
   end
 
+  describe 'version switching' do
+    it 'throws a error with a invalid version' do
+      data[:config][:mono] = 'foo'
+      should include_sexp [:echo, 'foo is not a valid version of mono.', {:ansi=>:red}]
```
akoeplinger: can we put the `foo` in quotes? I think it looks nicer.

Joshua-Anderson: I only use quotes when I need escaping. I'm not sure what style the travis team prefers.

spec/build/script/csharp_spec.rb
Joshua-Anderson: done

spec/build/script/csharp_spec.rb
akoeplinger: Nah, sorry, I meant to enclose the version part of the string so the message reads `"foo" is not a valid version` on the console ;)

akoeplinger: Nah, sorry, I meant to enclose the version part of the string so the message reads `"foo" is not a valid version` on the console ;)

lib/travis/build/script/csharp.rb
```diff
@@ -59,6 +72,32 @@ def script
...
+        def mono_version_valid?
+          if config[:mono] == 'latest' || MONO_VERSION_REGEXP === config[:mono]
+            if !is_mono_2? || (is_mono_2? && config[:mono] == '2.10.8')
+              true
+            end
+          end
+        end
+
+        def is_mono_2?
+          if config[:mono].split('.').first == "2"
```
akoeplinger: we should be able to do this without `split`, by getting the value of the first capturing group.

lib/travis/build/script/csharp.rb
```diff
@@ -59,6 +72,32 @@ def script
...
+          end
+        end
+
+        def mono_version_valid?
+          if config[:mono] == 'latest' || MONO_VERSION_REGEXP === config[:mono]
+            if !is_mono_2? || (is_mono_2? && config[:mono] == '2.10.8')
+              true
+            end
+          end
+        end
```
akoeplinger: I think this whole method can be simplified to:

```ruby
return true if config[:mono] == 'latest'

if match = config[:mono].match(MONO_VERSION_REGEXP)
  major = match.captures
  return true if major >= 3 || (major == 2 && config[:mono] == '2.10.8')
end
```

do we want to factor the version check into a separate method?

lib/travis/build/script/csharp.rb
```diff
@@ -20,19 +22,30 @@ def configure
...
-              sh.cmd 'mozroots --import --sync --quiet', timing: true
+              if !is_mono_2?
+                sh.cmd 'sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF', echo: false
+                sh.cmd "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian #{mono_repo} main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", echo: false
+                sh.cmd "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian wheezy-libtiff-compat main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", echo: false
+                sh.cmd 'sudo apt-get update -qq', timing: true
+                sh.cmd 'sudo apt-get install -qq mono-complete referenceassemblies-pcl nuget mono-vbnc fsharp', timing: true
+                sh.cmd 'mozroots --import --sync --quiet', timing: true
+              else
+                sh.cmd 'sudo apt-get install -qq mono-complete mono-basic', timing: true
```
akoeplinger: I'm curious why we have the `mono-basic` vs `mono-vbnc` as in the later versions?

I'd also switch the if condition, so `if is_mono2? ... ` to avoid negation.

akoeplinger: Ah, and we should definitely do `mozroots` here as well :)

lib/travis/build/script/csharp.rb
akoeplinger: This assumes you change the regex to contain the `()` for the capturing groups again ;)

lib/travis/build/script/csharp.rb
```diff
@@ -48,7 +61,7 @@ def export
         end
 
         def install
-          sh.cmd "nuget restore #{config[:solution]}", retry: true if config[:solution]
+          sh.cmd "nuget restore #{config[:solution]}", retry: true if config[:solution] && !is_mono_2?
         end
 
         def script
```
akoeplinger: unrelated to this PR, but just stumled over this: the xbuild line checks `if config[:solution]` again, this is redundant as we do it right above in the if.

lib/travis/build/script/csharp.rb
Joshua-Anderson: There isn't a mono-vbnc package as far as I can tell in Ubuntu, but the [mono-basic package](https://launchpad.net/ubuntu/+source/mono-basic) seems to contain the same things.

If we are not installing nuget than what is the point of running `mozroots`?

akoeplinger: There is: http://packages.ubuntu.com/precise/mono-vbnc

mozroots is important for any SSL connections, not just nuget :)

Joshua-Anderson: Thanks. That explains a lot, mono-basic only exists for newer versions of Ubuntu.

People who are fetching nuget as part of their buildpacks are probably running mozroots already and it wastes time for those who aren't.

akoeplinger: there's no `mono-basic` package, the one on launchpad you're referring too is just the name of the source, the package name is still `mono-vbnc`: http://bazaar.launchpad.net/~ubuntu-branches/ubuntu/vivid/mono-basic/vivid/view/head:/debian/control#L70

> People who are fetching nuget as part of their buildpacks are probably running mozroots already and it wastes time for those who aren't.

You could apply the same argument to the newer version as well, we install nuget+mozroots even though not everyone might be using it. I'd prefer playing safe and make it consistent in favor of saving a few seconds of build time.

lib/travis/build/script/csharp.rb
Joshua-Anderson: We can't do this because some projects might override the script but not install anything.

akoeplinger: If they override `script`, then this code won't get executed anyway, or am I missing something?

Joshua-Anderson: This code is executed unless they override install. The other code is executed unless they override script.

akoeplinger: Are we talking about the same thing? My comment was targeted at https://github.com/Joshua-Anderson/travis-build/blob/8a124e510cd40242649fe895a0f0a3a64e8e8d24/lib/travis/build/script/csharp.rb#L70 but GitHub wouldn't let me comment on that line in this PR :)

lib/travis/build/script/csharp.rb
```diff
@@ -20,19 +22,31 @@ def configure
...
-            if config[:csharp] == 'mono'
+            if mono_version_valid?
               sh.echo 'Installing Mono', ansi: :yellow
 
-              sh.cmd 'sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF', echo: false
-              sh.cmd "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian wheezy main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", echo: false
-              sh.cmd "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian wheezy-libtiff-compat main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", echo: false
-              sh.cmd 'sudo apt-get update -qq', timing: true
-              sh.cmd 'sudo apt-get install -qq mono-complete referenceassemblies-pcl nuget mono-vbnc fsharp', timing: true
+              if !is_mono_2?
```
akoeplinger: How do you feel about getting rid of the negation and doing:
```ruby
if is_mono_2?
  ...
else
  ...
end
```

I'd say this reads better :)

lib/travis/build/script/csharp.rb
```diff
@@ -59,6 +73,31 @@ def script
...
         end
+
+        def mono_repo
+          if config[:mono] == 'latest'
+            'wheezy'
+          else
+            if MONO_VERSION_REGEXP === config[:mono]
+              'wheezy/snapshots/' << config[:mono]
+            else
+              'wheezy'
```
akoeplinger: I think this else can never be hit?

lib/travis/build/script/csharp.rb
Joshua-Anderson: I see what you are saying now. Fixed it.

lib/travis/build/script/csharp.rb
Joshua-Anderson: I wanted to put the most common situation first so it is easier to find but you have a point.

lib/travis/build/script/csharp.rb
```diff
@@ -48,17 +62,38 @@ def export
         end
 
         def install
-          sh.cmd "nuget restore #{config[:solution]}", retry: true if config[:solution]
+          sh.cmd "nuget restore #{config[:solution]}", retry: true if config[:solution] && !is_mono_2?
```
akoeplinger: very minor nit: `&& !is_mono_2?` --> `unless is_mono_2?`

lib/travis/build/script/csharp.rb
```diff
@@ -48,17 +62,38 @@ def export
...
+          else
+            'wheezy/snapshots/' << config[:mono]
+          end
+        end
+
+        def mono_version_valid?
+          return true if config[:mono] == 'latest'
+          return false unless MONO_VERSION_REGEXP === config[:mono]
+
+          return false if MONO_VERSION_REGEXP.match(config[:mono])[1] == '2' && !is_mono_2?
```
akoeplinger: hmmm, wouldn't this accept `1.0.0` as a valid version?

Joshua-Anderson: Yeah. I guess we could blacklist mono 1, but I figured it wouldn't be a problem if we document the fact that mono 1 isn't supported. Who wants to test with mono 1 anyway? :smile:

akoeplinger: Yeah of course :) I just figured we can catch some dumb mistakes. Can we check that major >= 2?

Joshua-Anderson: Done.

akoeplinger:
Looks good to me! Awesome work :heart: 

@BanzaiMan can we get this deployed to staging?

BanzaiMan:

akoeplinger:
Soo, this seems to have worked quite well: https://staging.travis-ci.org/akoeplinger/travis-mono-test

The only issue is that the PCL assemblies package is not in the 3.8 snapshot repo...
I'll ask around tomorrow if we can get this added or if we need to workaround that.

Joshua-Anderson:
Thanks for testing that. Awesome!

akoeplinger:
Btw. we should look into adding the [PPA](https://launchpad.net/~directhex/+archive/ubuntu/ppa) so we can install Mono 3.2.8 as well. I know the Pash guys want to test against it and it would be great if we could offer it from the get go.

akoeplinger:
@Joshua-Anderson ping on the updates we discussed on slack :) if you don't have time, just let me know.

Joshua-Anderson:
Added support for mono 3.2.8 and errored the build when the install fails.

nterry:
We really should consider creating a set of class variables to hold these versions:

    MONO_3_10 = 3.10.0
    MONO_3_12 = 3.12.0
    MONO_2_10 = 2.10.0

etc... as well as storing some of those magic strings of repo keys in variables as well....

Joshua-Anderson:
The reason I didn't create variables for repo keys is that they are only used once and I figured it would be easiest to find them where they are used, rather than track down where a global is defined. I didn't do that for the version numbers because those variables aren't shorter than the string itself, which we only need to check if it is mono 2.10.8 or 3.2.8

akoeplinger:
@Joshua-Anderson I think it might be valuable to create methods for the whole version comparison, as we only ever use the result of that, e.g.:

``` ruby
def is_mono_2_10_8
  config[:mono] == '2.10.8'
end

def is_mono_3_2_8
  config[:mono] == '3.2.8'
end

def is_mono_3_8_0
  config[:mono] == '3.8.0'
end
```

this would simplify the checks in a few places and also reduce the chance of accidentally typing a different version when adding another check somewhere in the future.

Joshua-Anderson:
I don't think it makes any sense as `is_mono_3_2_8` isn't any significantly shorter than say, `config[:mono] == '3.8.0'` and we use the checks twice at the most.

akoeplinger:
It's not about being shorter but about encapsulation. Other language scripts do similar things, e.g. [python](https://github.com/travis-ci/travis-build/blob/ee61d0d60120cc30ef1073620110145461328e7d/lib/travis/build/script/python.rb#L51), [node.js](https://github.com/travis-ci/travis-build/blob/ee61d0d60120cc30ef1073620110145461328e7d/lib/travis/build/script/node_js.rb#L65), [rust](https://github.com/travis-ci/travis-build/blob/ee61d0d60120cc30ef1073620110145461328e7d/lib/travis/build/script/rust.rb#L60).

However I don't feel too strongly about it, so if you want to keep it as is I'm not going to complain ;)

PR #395, opened 2015-02-28
Better checking for failures in our R scripts.

craigcitro:
We use inline R scripts for several steps (eg package installation), but don't
convert script failures into a nonzero exit status. This diff fixes this for
all `Rscript` calls.

PTAL @hadley and @cbarraford for the R/ruby bits, and then @BanzaiMan for the merge once they OK.

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -131,9 +131,9 @@ module Travis
           sh.echo "Building with: R CMD build ${R_BUILD_ARGS}"
           sh.cmd "R CMD build #{config[:r_build_args]} ."
           tarball_script = [
-            'pkg <- devtools::as.package(".")',
-            'cat(paste0(pkg$package, "_", pkg$version, ".tar.gz"))',
-          ].join('; ')
+            'pkg <- devtools::as.package(".");',
+            'cat(paste0(pkg$package, "_", pkg$version, ".tar.gz"));',
+          ].join(' ')
           sh.export 'PKG_TARBALL', "$(Rscript -e '#{tarball_script}')"
 
           # Test the package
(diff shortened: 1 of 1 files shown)


hadley:
Did you consider setting the global error option to `q(status = 1)`?

craigcitro:
that's a good idea -- it would mean that i could drop the various `tryCatch` blocks, right?

i'd still need to do the manual checking on package installs, since `install.packages` doesn't signal an error if something goes wrong. (maybe i could also do `options(warn = 2)`, but i suspect that would lead to other warnings on package install spuriously becoming errors.)

hadley:
Maybe littler already has the capability to do that? Worth checking with @eddelbuettel first

eddelbuettel:
littler lets you do what R already offers:

```{.bash}
edd@max:~$ r -e'q("no", 0)' && echo No Error                                                                                                                                                                         
No Error
edd@max:~$ r -e'q("no", 42)' || echo "Error $?"
Error 42
edd@max:~$ 
```

craigcitro:
@eddelbuettel but in particular, `littler` doesn't change the R defaults for things like setting an exit status on command error?

craigcitro:
@hadley the other question is about messaging: if i set the global `error` option, i don't get my hands on the error (to print it first), unless i'm missing something. (is there a simple way to ask R for "last error"?)

craigcitro:
@hadley any objections to this going in as-is, with the caveat that we may want to clean it up later?

(i know @BanzaiMan was looking to merge some things early Monday ...)

BanzaiMan:
Ping me when you are satisfied with this PR.

hadley:
LGTM, although I wish `install.packages()` failed with an error.

craigcitro:
@hadley me too.

@BanzaiMan i think we're good to go. (this works, there may be a cleaner version later.)

PR #398, opened 2015-03-01
Quiet the output on `brew update` for R builds.

craigcitro:
PTAL @BanzaiMan

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -84,7 +84,9 @@ module Travis
             sh.cmd 'sudo chmod 2777 /usr/local/lib/R /usr/local/lib/R/site-library'
 
           when 'osx'
-            sh.cmd 'brew update', retry: true
+            # We want to update, but we don't need the 800+ lines of
+            # output.
+            sh.cmd 'brew update >/dev/null', retry: true
 
             # Install from latest CRAN binary build for OS X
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/r.rb
```diff
@@ -84,7 +84,9 @@ def setup
             sh.cmd 'sudo chmod 2777 /usr/local/lib/R /usr/local/lib/R/site-library'
 
           when 'osx'
-            sh.cmd 'brew update', retry: true
+            # We want to update, but we don't need the 800+ lines of
+            # output.
+            sh.cmd 'brew update >/dev/null', retry: true
```
BanzaiMan: You can also use `echo: false` instead. Either is fine; I'm just pointing it out.

craigcitro: oh -- in the generated bash script, it looked like `echo` only controlled whether or not we printed the command before executing it. does it also prevent output from landing in the logs?

BanzaiMan: Ah. You are right about that. Carry on.

craigcitro: sweet ... then I think this one is ready to go ;)

BanzaiMan:
This, #397, #396, and #395, have been deployed to production.

PR #400, opened 2015-03-03
Implement transifex addon

meatballhat:
- [x] get a lot of feedback
- [x] change a lot of things
- [x] make it work
- [x] profit :money_with_wings:


lib/travis/build/addons/transifex.rb
```diff
@@ -0,0 +1,134 @@
...
+          end
+
+          def tx_init_options
+            @tx_init_options ||= [].tap do |o|
+              o << "--host=#{tx_config[:host]}"
+            end.join(' ')
+          end
+
+          def tx_set_source_options
+            @tx_set_source_options ||= [].tap do |o|
```
diegobz: Good stuff. We need to keep in mind that not all frameworks have the same file structure. So this can be an optional option, but people can actually have this configuration stored directly in the repo. Some resources might have multiple resource files too. Ex. https://github.com/bikalabs/Bika-LIMS/blob/master/bika/lims/locales/.tx/config

More reference: http://docs.transifex.com/developer/client/config

lib/travis/build/addons/transifex.rb
```diff
@@ -0,0 +1,134 @@
...
+            @validation_errors = []
+            warn_invalid unless @valid
+            @valid
+          end
+
+          def tx_config
+            @tx_config ||= config[:transifex].deep_symbolize_keys.reverse_merge(DEFAULTS)
+          end
+
+          def tx_init_options
```
diegobz: `tx init` expects a `username` and `passwd`. It basically creates a `~/.transifexrc`. Usually people just generate the file directly using environment vars for sensitive info. Ex. https://github.com/tendenci/tendenci-donations/blob/master/transifex.sh

lib/travis/build/addons/transifex.rb
meatballhat: :thumbsup: got it.  I admit this is primarily a port of the OpenStack implementation, which I'm sure you guessed :smiley:.  Would it be better to omit the setting of source options entirely in the initial implementation (on line 55)?

lib/travis/build/addons/transifex.rb
meatballhat: Ah, nice.  What if we did something similar such that all of the required options were translated directly from the YAML or env vars to `~/.transifexrc`?  a la:

Set all options in YAML:
```
travis setup transifex
```
which results in:
``` yaml
addons:
  transifex:
    username: abcd1234
    password:
      secure: ABCD...ZYXW
```

**OR**

Set username and password in env vars
``` bash
travis env set TRANSIFEX_USERNAME abcd1234
travis env set TRANSIFEX_PASSWORD secretsecret
```
with the complimentary YAML being:
``` yaml
addons:
  transifex: true
```

The idea being that both of these would result in the generation of a `~/.transifexrc` that looks like:

``` ini
[https://www.transifex.com]
hostname = https://www.transifex.com
password = secretsecret
token =
username = abcd1234
```

lib/travis/build/addons/transifex.rb
diegobz: I think so, that probably makes more sense.

lib/travis/build/addons/transifex.rb
diegobz: I'm using the second approach at the moment here at Transifex, however I don't set the `passwd` in the YAML file. I actually do it on the repo settings, so it doesn't get propagated to all repositories (clones) our devs have.

lib/travis/build/addons/transifex.rb
```diff
@@ -0,0 +1,62 @@
...
+              echo "[#{tx_config[:host]}]
+              hostname = #{tx_config[:host]}
+              username = ${TX_USER:-#{tx_config[:username]}}
+              password = ${TX_PASSWD:-#{tx_config[:password]}}
+              token = ${TX_TOKEN:-#{tx_config[:token]}}" > #{Travis::Build::HOME_DIR}/.transifexrc
+            EOF
+          end
+
+          def update
+            sh.cmd "tx push --source --translations --no-interactive", echo: true
```
diegobz: @meatballhat We should drop the `--translations` here now. This should push only source content in general.

meatballhat: :thumbsup:

lib/travis/build/addons/transifex.rb
```diff
@@ -0,0 +1,62 @@
...
+            sh.cmd <<-EOF.gsub(/^ {14}/, ''), echo: false
+              echo "[${TX_HOSTNAME:-#{tx_config[:hostname]}}]
+              hostname = ${TX_HOSTNAME:-#{tx_config[:hostname]}}
+              username = ${TX_USERNAME:-#{tx_config[:username]}}
+              password = ${TX_PASSWORD:-#{tx_config[:password]}}
+              token = ${TX_TOKEN:-#{tx_config[:token]}}" > #{Travis::Build::HOME_DIR}/.transifexrc
+            EOF
+          end
+
+          def update
```
diegobz: How about calling this `source_push` or something, instead of 'update'?

meatballhat: :+1: can do

lib/travis/build/addons/transifex.rb
```diff
@@ -0,0 +1,86 @@
...
+              echo "[${TX_HOSTNAME:-#{tx_config[:hostname]}}]
+              hostname = ${TX_HOSTNAME:-#{tx_config[:hostname]}}
+              username = ${TX_USERNAME:-#{tx_config[:username]}}
+              password = ${TX_PASSWORD:-#{tx_config[:password]}}
+              token = ${TX_TOKEN:-#{tx_config[:token]}}" > #{tx_rc_path}
+            EOF
+          end
+
+          def source_push
+            sh.raw "echo $TRAVIS_JOB_NUMBER | grep -q -E '\\.#{tx_config[:auto_push][:job]}$'",
```
diegobz: @meatballhat @emdantrim I'm not sure what's the idea behind this line here. Dan, maybe you can elaborate a bit?

meatballhat: That's the line meant to ensure the automatic push only runs once per build.  It's fairly brittle and hacky, and really should be implemented as a first class feature, but such a thing isn't likely to land until we roll out support for customizable build pipelines.

lib/travis/build/addons/transifex.rb
```diff
@@ -0,0 +1,83 @@
...
+        CLIENT_VERSION = '>=0.11'
+        SUPER_USER_SAFE = true
+        DEFAULTS = {
+          hostname: 'https://www.transifex.com',
+          username: '',
+          password: '',
+          token: '',
+          auto_push: {
+            enabled: true,
+            job: 1,
```
diegobz: @meatballhat what did you have in mind with the property `jobs` here?

meatballhat:
@BanzaiMan 👋

BanzaiMan:
This si superseded by https://github.com/travis-ci/dpl/pull/277.

PR #404, opened 2015-03-06
Feature: Allow testing with Python 3.5-dev

BanzaiMan:
This PR allows testing with Python 3.5-dev with ease.

The binary archive is created by https://github.com/travis-ci/cpython-builder.

It's been tested on staging: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/428817#L85


BanzaiMan:
Oh. Yes. It needs specs.

BanzaiMan:
Note that, unlike the PHP process, this Python process is tied to `python-build`'s way of building `3.5-dev`. So the `nightly` alias we use for the PHP counterpart should not apply.

BanzaiMan:
We decided to only support `nightly` as the name.

PR #410, opened 2015-03-20
Add customizable 'before_cache' stage

BanzaiMan:
Often users want to perform cleanup tasks before the new cache
is uploaded; for example, deleting a new marker file which
fools casher into thinking that the entire cache is new.

Also, `after_result` stage is renamed to `cache`, which correctly
describes the purpose of this stage.
Note that this is now a non-critical stage, since failure to update
cache should not fail the entire job.


BanzaiMan:
Tested in staging: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/429632#L97

lib/travis/build/stages.rb
```diff
@@ -10,7 +10,8 @@ class Stages
       STAGES = [
         :builtin,     [:header, :configure, :checkout, :prepare, :disable_sudo, :export, :setup, :announce],
         :custom,      [:before_install, :install, :before_script, :script],
-        :builtin,     [:after_result],
+        :custom,      [:before_cache],
```
BanzaiMan: Maybe I should tack this onto the line above.

joshk: sounds good

PR #415, opened 2015-03-28
Add Perl6 build support

paultcochrane:
This attempts to address issue https://github.com/travis-ci/travis-ci/issues/2978, however the build
process is *very* slow, such that even though it works, it is basically unusable.  This change is intended generate feedback/discussion about the build script code so as to improve it before finally closing issue https://github.com/travis-ci/travis-ci/issues/2978.


joshk:
is sudo required?

maybe this is something we should bake into the Perl image?

paultcochrane:
Since the files are installed under `/usr`, yes, sudo is required.  One could build and install to a user-writable location, however then the paths would have to be set correctly, causing extra debugging and maintenance overhead.  I chose to install under `/usr` so as to simplify such issues.  What would be a better option as build/install location, such that sudo wouldn't be required?

I don't think this should be baked into the Perl image (assuming you mean Perl5) since Perl5 and Perl6 are two separate languages.  They are considered to be "sister" languages, however since Perl6 is a backwards incompatible development it would be fair to both the Perl5 and Perl6 communities to treat them separately on Travis.  See also the comments about this issue by @ugexe in https://github.com/travis-ci/travis-ci/issues/2978.

ugexe:
I think most people in both the Perl5 and Perl6 communities would agree they should be separate entities. What should be taken into consideration is that Perl5 is *required* to build Perl6 in many (most?) circumstances. So if you want to use a build newer than a month (or however often the packaged versions are created) you need to have Perl5. I'm guessing Perl5 is standard on most, if not all, of the travis builds so this is mostly a non-issue, although maybe Perl6 class could inherit from the Perl5 so that its Perl5 environment is the same for both Perl5 and Perl6? (this might not even be an actual problem, i'm just thinking about various Build.PLs, module P5, etc)

For the installation location my guess would be to wherever travis-ci puts the perls its creating via perlbrew. In Perl5 land you can use perl-travis-helpers for installing/testing against versions of Perl5 that travis-ci does not have, which just installs to the relative directory (like rakudobrew).

BanzaiMan:
Shouldn't the install path be configurable? Not installing into `/usr` is preferable, if Perl6 wants to be a different language and wants to support multiple versions.

BanzaiMan:
On a second thought, support for multiple versions doesn't require installation in different directories.

lib/travis/build/script/perl6.rb
```diff
@@ -0,0 +1,45 @@
...
+
+          sh.echo ''
+          sh.echo 'Perl6 support for Travis-CI is community maintained.', ansi: :red
+          sh.echo 'Please open any issues at https://github.com/travis-ci/travis-ci/issues/new', ansi: :red
+
+          sh.echo 'Installing Rakudo (MoarVM)', ansi: :yellow
+          sh.cmd 'git clone https://github.com/rakudo/rakudo.git'
+          sh.cmd 'cd rakudo'
+          sh.cmd 'sudo perl Configure.pl --backends=moar --gen-nqp --gen-moar --prefix=/usr'
+          sh.cmd 'sudo make install'
```
BanzaiMan: How long does this take?

ugexe: You can see build times listed in the original issue: https://github.com/travis-ci/travis-ci/issues/2978 which use perl.rb. I'd run your code but i've not been able to get local travis testing working.

Also shown in the original issue is the use of 'rakudobrew', a perl program much like perlbrew which allows installation to a custom directory. 
https://github.com/ugexe/zef/blob/master/.travis.yml#L14-L17 for instance builds/installs perl6 and a package manager to ~./rakudobrew/[backend name]/


For an empty module, using the MoarVM backend, it takes ~5-8 minutes to build rakudo perl6, the backend, nqp, and the empty module.

JVM takes significantly longer, but I think getting just MoarVM working first and primarily should be the main focus.

paultcochrane: Cloning the Rakudo repo and then building and installing MoarVM, NQP and finally Rakudo took 344s on the Travis infrastructure in a speed test today.  Using `rakudobrew` for the same process took 394s.  The variation could be purely statistical.  Thus one could say that a full Rakudo build and installation takes about 6 minutes.

BanzaiMan:
I added timing and folds to these commands and tested them on the real VM. On containers, it adds about 3 minutes total (90 seconds each for configuring and installing). On the standard infrastructure, it adds a bit more (almost 3 minutes each to configure and to install).

It is ideal if there is a pre-compiled binary package somewhere to pull and install, but it seems like acceptable cost to testing Perl 6.

paultcochrane:
Many thanks for looking into this!  I've updated the build script to use `rakudobrew`, since this allows one to use many specific Rakudo versions in a `.travis.yml` configuration (e.g. `2015.04`, `2014.12` etc.).  Thus one can now either use latest Rakudo (built on top of MoarVM and NQP) or any of the older versions.

A Debian package of MoarVM is available (however not yet in unstable); packages for NQP and Rakudo (built on MoarVM and NQP; there is a package built on Parrot, however this is obsolete) are in the works will become available at some stage (we're waiting for a new release of an upstream package).  As soon as these packages are available, then the cost to providing Perl6 will be much reduced.

lib/travis/build/script/perl6.rb
```diff
@@ -0,0 +1,60 @@
...
+
+        def export
+          super
+          sh.export 'TRAVIS_PERL6_VERSION', version, echo: false
+        end
+
+        def setup
+          super
+          if version == "latest"
+            sh.cmd 'rakudobrew build moar', assert: false
```
BanzaiMan: I suggest putting this in a fold.

Also, adding timing would be informative.

lib/travis/build/script/perl6.rb
```diff
@@ -0,0 +1,60 @@
...
+          else
+            sh.cmd "rakudobrew triple #{version} #{version} #{version}", assert: false
```
BanzaiMan: Same as above.

BanzaiMan:
Besides the 2 minor points, I think this is ready to roll out. Do you have 2 other volunteers to share the support?

We've published community-language development guide. Please take a look. http://docs.travis-ci.com/user/languages/community-supported-languages/

Also, if you have fun bits of information about you (and other volunteers) to share in the announcement blog post, I'd appreciate it!

hoelzro:
@BanzaiMan I would happily lend support to this!

ugexe:
I'll volunteer again :+1:

BanzaiMan:
Thank you! The volunteers' names should be announced in the build log.

BanzaiMan:
@paultcochrane @ugexe @hoelzro If you could update this PR and work on docs, that'll be great.

paultcochrane:
@BanzaiMan have updated the build script as requested and am in the process of creating the docs.

What kind of fun information is appropriate for the blog entry?

BanzaiMan:
Hello. I looked at the past blog entries, and saw that there weren't additional bits. So, it seems all right without.

paultcochrane:
@BanzaiMan: the build script has matured a bit after a couple of rounds of feedback from @ugexe and @hoelzro, and the docs have been added in https://github.com/travis-ci/docs-travis-ci-com/pull/253.  Let me know if there's anything else that needs doing before the build script can be rolled out.  Many thanks again for your help!  It's much appreciated :-)

BanzaiMan:
Hi, @paultcochrane. Did you have a chance to look at @ugexe's comment above?

paultcochrane:
Hi @BanzaiMan, yes I've discussed @ugexe's comment with him on irc.  I've also discussed the issue with @hoelzro and we both see @ugexe's point, however still think it would be better to keep the `panda`-based solution which is currently implemented.

It would be good to have the Perl6 build scripts available to us to use directly on Travis in order to iron out any issues before making the availablity of Perl6 as a build language more widely known.  I assume we can simply submit PRs to further adjust the Perl6 build scripts, is that right?

tony-o:
Can we please not do this?  I'm seconding @ugexe's concerns.

paultcochrane:
@BanzaiMan further discussion has ensued.  I'm going to change the current implementation and request further feedback from the Perl6 people.

paultcochrane:
@BanzaiMan: I have just pushed the latest set of updates.  I believe we are all in agreement with how dependencies are now managed and how the test suite is to be run.  The docs (https://github.com/travis-ci/docs-travis-ci-com/pull/253) have also been updated.  Unless there are major complaints, I think this is ready to be merged now.  Many thanks again for your help!

BanzaiMan:

BanzaiMan:
I plan to push this on 2015-06-29.

paultcochrane:
@BanzaiMan, great!  That's awesome news :-)

PR #414, opened 2015-03-27
update default D compiler (dmd) version to new 2.067 release

Extrawurst:

PR #418, opened 2015-03-30
get latest versions from compiler sites

MartinNowak:
- avoids updating the build script when a new version is released
- supports latest dmd, ldc, and gdc compilers without specifying a version
- fixes Issue https://github.com/travis-ci/travis-ci/issues/3472 and
  https://github.com/travis-ci/travis-build/pull/414

### DIFF
--- a/lib/travis/build/script/d.rb
+++ b/lib/travis/build/script/d.rb
@@ -7,7 +7,7 @@ module Travis
     class Script
       class D < Script
         DEFAULTS = {
-          d: 'dmd-2.066.1'
+          d: 'dmd'
         }
 
         def cache_slug
@@ -85,6 +85,11 @@ module Travis
 
         def compiler_url(compiler = config[:d])
           case compiler
+          # latest DMD
+          when 'dmd'
+            sh.cmd 'LATEST_DMD=$(curl http://ftp.digitalmars.com/LATEST)', echo: false
+            "http://downloads.dlang.org/releases/2.x/${LATEST_DMD}/dmd.${LATEST_DMD}.#{os}.zip"
+
           # dmd-2.062, dmd-2.065.0, dmd-2.066.1-rc3
           when /^dmd-2\.(?<maj>\d{3})(?<min>\.\d)?(?<suffix>-.*)?$/
             basename = "dmd.2.#{$~[:maj]}#{$~[:min]}#{$~[:suffix]}"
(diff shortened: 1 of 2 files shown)


MartinNowak:
Ping @BanzaiMan, we recently released a new compiler version and this updates the travis-ci build scripts.

MartinNowak:
Can we get this integrated soon @BanzaiMan?

BanzaiMan:
@MartinNowak this has been deployed to staging. Could you test it?

MartinNowak:
Triggered a job, let's see if it works. I already tested `travis compile`, so I'm pretty confident.
https://staging.travis-ci.org/MartinNowak/bloom/builds/429881

BanzaiMan:
Hmm. Looks like something is not right with the staging environment. I'll take a look at it tomorrow. Sorry for the delay.

MartinNowak:
Hm, I tried to restart the job, but it still seems to hang @BanzaiMan.

MartinNowak:
Did you have time to look at the staging server @BanzaiMan?

BanzaiMan:
Yes, I finally did. Thank you! The results look pretty good to me. If they look good to you, too, I'll merge this and deploy it Monday.

MartinNowak:
Thanks, results look fine.

MartinNowak:
Seeing the download times and the traffic, any suggestion how to cache the compiler downloads?

BanzaiMan:
How big are the files? I tend to think this is a necessary cost that must be incurred at some point, since these compilers must be downloaded somehow.

MartinNowak:
Around 20MB, it doesn't kill anyone's traffic budget, but it's simply wasteful, even more on your side.
I thought you guys might have a HTTP proxy cache setup, would save a considerable amount of traffic and tester time.

MartinNowak:
Thanks, I also made a documentation pull https://github.com/travis-ci/docs-travis-ci-com/pull/222.

braddr:
@BanzaiMan To add some data to Martin's point, here's the number of downloads via the travis-ci nat device so far this month:

2015-04-01:96
2015-04-02:228
2015-04-03:182
2015-04-04:34
2015-04-05:104
2015-04-06:111
2015-04-07:255
2015-04-08:114
2015-04-09:450

Not large, but MOST of those are downloading the exact same file, once for each build.  Consider the cost of downloading gcc or ruby or python each time a build was executed for one of those.  Individually not a big deal, but adds up rather fast.  Luckily they're coming from S3 (us-west-2) so are both fast and only cost the inter-region traffic since travis mostly lives in EC2 us-east-1.

MartinNowak:
Even more important [this](https://travis-ci.org/D-Programming-Language/dub/jobs/58277093#L94) job spends 50 out of 127 seconds downloading, for simple libraries, where testing only takes a few seconds, the ratio gets much worse.

PR #420, opened 2015-04-03
C#: add call to `super` in setup

akoeplinger:
should fix directory caching


Joshua-Anderson:
Speaking of this, we should work on getting NuGet cache support added as well.

PR #423, opened 2015-04-09
Improvements to Haxe support

andyli:
* `retry: true` for each `apt-get` cmd.
 * Use double quotes for hxmls when running the haxe compiler. Such that ppl can use env variables for the `hxml:` keys in yaml.
 * Fail the build if no test can be run, similar to the python builds.


BanzaiMan:
We'll deploy this on Monday. Thanks!

andyli:
Great! Thanks!

PR #426, opened 2015-04-15
Escape all values from users

drogus:
From commit message:

```
Before this commit we were not escaping env vars values from users. A
big drawback of that is that our secure env vars are not "secure by
default", becasue if an unaware user sets a password with special
characters, like foo&\, bash will throw an error possibly outputting the
password.

This commit escapes all of the env values with `shellescape`.
```

Additionally, I think that we will need to somehow announce this change if we decide to merge it. Some people may already escape their env vars, which will most likely going to break stuff for them.

/cc @svenfuchs

### DIFF
--- a/lib/travis/build/env/config.rb
+++ b/lib/travis/build/env/config.rb
@@ -17,7 +17,8 @@ module Travis
           def env_vars
             vars = Array(config[:global_env]) + Array(config[:env])
             vars = vars.compact.reject(&:empty?)
-            vars.flat_map { |var| Var.parse(var) }
+            vars = vars.flat_map { |var| Var.parse(var) }.
+              map { |var| var[1] = var[1].shellescape; var }
           end
       end
     end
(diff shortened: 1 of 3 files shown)


drogus:
I see there are some failures in the build. I'll look into these after we discuss if this PR is a good thing to do.

henrikhodne:
In addition to breaking for users who already escape things, this will also break for users who rely on the values not to be escaped (for example if they use shell substitutions). I don't think that's too common for secure env vars, but isn't uncommon for regular env vars. we might be able to work around that by only escaping secure env vars?

drogus:
> we might be able to work around that by only escaping secure env vars?

I don't really like inconsistency that it would introduce.

The other option would be to at least not show the var when the error happens, but that could be hard as this is usually due to syntax errors. With my quick testing I haven't found a way to escape it properly.

Another option is to try to be smart and detect if there're any special chars that are unescaped and then escape them.

das-g:
Another option would be to highlight more prominently in the documentation that (and what) values need to be escaped or quoted if expansion is not desired.

meatballhat:
Some of this and the concerns about specifics have since been addressed (I believe) via output filtering, plus it's pretty dusty, so closing!  Please holler if you believe this should be reopened.

PR #427, opened 2015-04-15
R-lang: Fail, but not abort the build if R CMD check fails

jimhester:
Previously failing `R CMD check` resulted in the entire travis build
aborting, which prevented any `after_failure` commands to be executed.

This change will automatically dump the check log files if the check
fails and will also fail rather than abort the build, so `after_failure`
commands will still occur.

Fixes travis-ci#3312

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -143,7 +143,13 @@ module Travis
           sh.echo 'Testing with: R CMD check "${PKG_TARBALL}" ' +
                   "#{config[:r_check_args]}"
           sh.cmd "R CMD check \"${PKG_TARBALL}\" #{config[:r_check_args]}",
-                 assert: true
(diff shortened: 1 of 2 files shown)


craigcitro:
LGTM -- did you get a chance to try testing this in a vagrant VM?

jimhester:
I did, all tests pass on my machine.

craigcitro:
awesome. the issue i sometimes hit is that the generated shell script either (1) has a syntax error or (2) doesn't fail as i expect -- i don't know how to write sane failure tests for things like `R CMD check failed` with the existing bits.

jimhester:
I would have liked to add a specific test case for an R CMD check failure.
I grepped through the existing tests and didn't see anything similar
however, and I don't know ruby or rails well enough to figure out how to do
it.

PR #431, opened 2015-04-20
Restart MySQL on each boot

BanzaiMan:
With the move to MySQL 5.6 packages from mysql.com, we noticed that
MySQL socket (`/var/run/mysqld/mysqld.sock`) is inexplicably lost (sometimes?).
The process is running, and users can connect with TCP
(e.g., `mysql -h 127.0.0.1`), but connexions over the socket are
unsuccessful, leading to lots of frustration.

Upon restarting, though, the socket is re-created, and things
work.

I have not been able to determine what is removing the socket.


BanzaiMan:
Tested on staging: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/430098#L127

lib/travis/build/appliances/restart_mysql.rb
```diff
@@ -0,0 +1,14 @@
+require 'travis/build/appliances/base'
+
+module Travis
+  module Build
+    module Appliances
+      class RestartMysql < Base
+        def apply
+          sh.cmd '(ls /var/run/mysqld/mysqld.sock >& /dev/null) || sudo service mysql restart'
```
meatballhat: Can this be `test -e /var/run/mysqld/mysqld.sock` instead of `ls`?  Using `ls` for noninteractive stuff is a recipe for sadness, IMHO.

/cc @tianon

BanzaiMan: For semantics, then, `test -S` should be more appropriate.

meatballhat: Agreed! :smiley_cat:

BanzaiMan:
Do not merge until MySQL 5.6 goes live.

meatballhat:
:ship: :it: (after MySQL 5.6 goes live)

BanzaiMan:
Things that might explain the loss of the socket:

1. The boot cleanup process in `/tmp`. The socket is in `/var/run`, so this does not apply.
1. SELinux. This is a plain Ubuntu, so this does not apply.
1. The path for the socket is correct; if you restart it, the file is recreated. Furthermore, `/var/run/mysqld/mysqld.sock` shows up in `lsof`, but is not marked DELETED.
1. `/var/run` is a symbolic link to `/run`, which is not a FS mount.

[This StackOverflow answer](http://askubuntu.com/a/595560) alludes to an issue on Trusty; maybe a similar problem exists for Precise, but I don't really know.

BanzaiMan:
There is at least one report of the missing socket even with this PR merged and deployed.

https://twitter.com/user/status/592736165465346049

Not clear what's happening with this yet.

BanzaiMan:
We've received a few more such reports. So far, the common factor has been that the failing builds use `apt-get` to manipulate packages.

BanzaiMan:
I'm leaning towards aliasing `mysql` to `mysql -h 127.0.0.1`.

BanzaiMan:
Or set `MYSQL_HOST` to `127.0.0.1`.

But any of these workarounds will not fix the issue if `config/database.yml` connects to MySQL over the socket instead of TCP.

PR #432, opened 2015-04-22
Allow to specify caching timeout

drogus:
This allows to set timeout for cacher with:

```
cache:
  timeout: 1000
```

### DIFF
--- a/lib/travis/build/script/shared/directory_cache.rb
+++ b/lib/travis/build/script/shared/directory_cache.rb
@@ -21,6 +21,9 @@ module Travis
         end
 
         def use_directory_cache?
+          if data.cache[:timeout]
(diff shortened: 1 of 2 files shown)


svenfuchs:
:+1: 

Saw this on support before, so I'm all for merging this, and adding it to the docs.

lukeyeager:
Is that in minutes?

BanzaiMan:
@lukeyeager Seconds.

lukeyeager:
That's what I thought. Can you explain why I have caches that are months old even though I set the timeout to be 1 week?

https://github.com/NVIDIA/DIGITS/blob/v4.0.0-rc.3/.travis.yml#L18

PR #435, opened 2015-04-27
Hotfix for Mono 4.0 crash during installation on Travis

akoeplinger:
There's a fix pending but we shouldn't break people in the meantime, so just pin to 3.12.0 for now.


Joshua-Anderson:
Looks good to me!

PR #437, opened 2015-04-29
Abort  hhvm-nightly builds with exit status 1

BanzaiMan:
See https://github.com/travis-ci/travis-ci/issues/3788
and https://github.com/facebook/hhvm/issues/5220


BanzaiMan:
There is no test. I can add one if it's really necessary.

meatballhat:
I wonder about the folks who are depending on a meaningful job status, though.  hhvm-nightly is currently broken on precise, so I would think the correct status would be nonzero, and let folks know that they should probably add it to the allowed failures.  Once we switch everyone to trusty, the failures go away.

PR #438, opened 2015-04-29
Ensure $PATH is prepended with a few entries

meatballhat:
to ease development in container-based infrastructure.

### DIFF
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -197,10 +197,18 @@ decrypt() {
 }
 
 # XXX Forcefully removing rabbitmq source until next build env update
-# XXX See http://www.traviscistatus.com/incidents/6xtkpm1zglg3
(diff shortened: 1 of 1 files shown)


sethvargo:
:+1:

BanzaiMan:
Could you motivate this a little more clearly? What issues were there without this? Do we have related GitHub issues?

meatballhat:
@BanzaiMan we get a decent amount of "support noise" about these two PATH entries, one of which has been partly addressed at the build env image layer [citation needed].  By adding these entries by default, it removes the need for folks in container-based infrastructure in particular from having to add the entries themselves given they aren't able to install into `/usr/bin` or `/usr/local/bin`.

BanzaiMan:
Ah, OK. The citation you seek is https://github.com/travis-ci/travis-cookbooks/blob/master/ci_environment/travis_build_environment/files/default/ci_user/xdg_path.sh.

PR #441, opened 2015-05-01
Run an Elixir test-env compile before running tests

aprescott:
This avoids cluttering `mix test` up with compilation output, and also folds the compile output into its own section.

# Before

screen shot 2015-05-01 at 10 38 05

# After

screen shot 2015-05-01 at 10 34 25

### DIFF
--- a/lib/travis/build/script/elixir.rb
+++ b/lib/travis/build/script/elixir.rb
@@ -30,6 +30,7 @@ module Travis
         def install
           sh.cmd 'mix local.hex --force'
           sh.cmd 'mix deps.get'
+          sh.cmd 'MIX_ENV=test mix compile'
         end
 
         def script


aprescott:
Actually I'm not sure if the folding will work now, just because I don't quite know how the `install` command's output gets folded.

When I worked around this in a `.travis.yml` for a specific Elixir project I had

```yaml
before_script:
  - MIX_ENV=test mix compile
script:
  - mix test
```

Which you can see in the "After" screenshot.

Will this change lead to the same folding behavior? Don't know how that works.

lib/travis/build/script/elixir.rb
```diff
@@ -30,6 +30,7 @@ def announce
         def install
           sh.cmd 'mix local.hex --force'
           sh.cmd 'mix deps.get'
+          sh.cmd 'MIX_ENV=test mix compile'
```
aprescott: Should I add `fold:` to these lines?

```shell
sh.cmd 'mix deps.get', fold: 'mix.deps.get'
sh.cmd 'MIX_ENV=test mix compile, fold: 'mix.compile'
```

BanzaiMan:
`before_*` phases are not meant for implementation in the default script. They are for user customization. `install` is put in a fold by default, and `script` is not.

aprescott:
Ah, so no `fold:` necessary? Sounds good!

BanzaiMan:
Conceptually, putting the "compile" step into "install" phase seems wrong, though.

aprescott:
@BanzaiMan the `mix test` in the [`script` method](https://github.com/travis-ci/travis-build/blob/9466a7bb90c7848d32cb6496b19e7099a5ff382a/lib/travis/build/script/elixir.rb#L36) will run a compile because it's necessary in order for the tests to run at all. Since this compilation step has to happen anyway, and compilation is, to me, part of installation/setup, it doesn't seem like the wrong place. And purely from a UX perspective, it's far clearer to not have hundreds of lines of compilation output masking the result of your tests.

BanzaiMan:
Then I think it should be a separate step (in a fold) in the `compile` section.

aprescott:
@BanzaiMan that makes sense. I don't see a `def compile` anywhere for Elixir, Erlang, or the general `script.rb`. Could you point me to the right place?

BanzaiMan:
Ah. Sorry. `compile` stage does not exist.

BanzaiMan:
I really don't think `mix compile` should run during the `install` stage. This change affects what users may be running in `before_script`.

BanzaiMan:
Thanks for opening the PR, but I think it is best to leave it as is.

PR #449, opened 2015-05-07
Update some tex-related paths on OSX R support.

craigcitro:
These commands need to run as root, so an explicit path is clearly the easiest
choice here.

PTAL @BanzaiMan


meatballhat:
:shipit:

PR #451, opened 2015-05-08
Set a reasonable initial max heap size

BanzaiMan:
Some users report problems with initializing JVM under
certain circumstances:

    Could not reserve enough space for object heap

(See travis-ci/travis-ci#3805)

We use `JAVA_TOOL_OPTIONS` to set the maximum heap size in the hopes
that this will eradicate the problem.

Notice that the command line options (e.g., `java -Xmx1g …`) will override `JAVA_TOOL_OPTIONS`.


meatballhat:
Will this also be applied for android builds?

BanzaiMan:
@meatballhat Yes. But for Android we would need additional tweaks, since `android-update-sdk` script would need to create JVM to install components, and this can also fail.

PR #454, opened 2015-05-19
allow domain-related arguments for sauce connect addon

ngokevin:
Hey from Mozilla,

We got a use case where we want to tunnel the ```localhost``` domain with Sauce Connect, but the Travis Sauce Connect addon doesn't allow for many arguments. 

To only pass the arguments in when they're defined, I had Ruby build the argstrings. If the arguments are not defined, the argstrings are empty strings and don't affect the Sauce Connect command. 

```bash
$ export TRAVISTEST=""
$ sc/bin/sc -u MYUSER -k MYKEY $TRAVISTEST
27 May 13:38:13 - Command line arguments: sc/bin/sc -u MYUSER -k ****

$ export TRAVISTEST="-B localhost"
$ sc/bin/sc -u MYUSER -k MYKEY $TRAVISTEST
27 May 13:38:36 - Command line arguments: sc/bin/sc -u MYUSER -k **** -B localhost
```

r? @meatballhat

### DIFF
--- a/lib/travis/build/addons/sauce_connect.rb
+++ b/lib/travis/build/addons/sauce_connect.rb
@@ -15,6 +15,18 @@ module Travis
           sh.export 'SAUCE_USERNAME', username, echo: false if username
           sh.export 'SAUCE_ACCESS_KEY', access_key, echo: false if access_key
 
+          if direct_domains
+            sh.export 'SAUCE_DIRECT_DOMAINS', "-D #{direct_domains}", echo: false
+          end
+
+          if no_ssl_bump_domains
+            sh.export 'SAUCE_NO_SSL_BUMP_DOMAINS', "-B #{no_ssl_bump_domains}", echo: false
+          end
+
+          if tunnel_domains
+            sh.export 'SAUCE_TUNNEL_DOMAINS', "-t #{tunnel_domains}", echo: false
+          end
+
           sh.fold 'sauce_connect.start' do
             sh.echo 'Starting Sauce Connect', echo: false, ansi: :yellow
             sh.cmd 'travis_start_sauce_connect', assert: false, echo: true, timing: true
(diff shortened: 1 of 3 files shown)


ngokevin:
Or r? @svenfuchs if you're not too busy :)

lib/travis/build/addons/sauce_connect/templates/sauce_connect.sh
```diff
@@ -53,7 +53,10 @@ function travis_start_sauce_connect() {
   ${sc_dir}/bin/sc \
     ${sc_tunnel_id_arg} \
     -f ${sc_readyfile} \
-    -l ${sc_logfile} &
+    -l ${sc_logfile} \
+    -B ${SAUCE_NO_SSL_BUMP_DOMAINS} \
```
BanzaiMan: What happens if `$SAUCE_NO_SSL_BUMP_DOMAINS` is empty?

ngokevin: I was hoping an empty argument would mean it wouldn't do anything differently. I will explore to make sure it doesn't break anything.

ngokevin:
Updated to work when the arguments are not defined. Check the PR description for more details.

ngokevin:
How does it look now?

lib/travis/build/addons/sauce_connect.rb
```diff
@@ -15,6 +15,15 @@ def before_before_script
           sh.export 'SAUCE_USERNAME', username, echo: false if username
           sh.export 'SAUCE_ACCESS_KEY', access_key, echo: false if access_key
 
+          direct_domains_arg = direct_domains ? "-D #{direct_domains}" : ''
+          sh.export 'SAUCE_DIRECT_DOMAINS', direct_domains_arg, echo: false if direct_domains_arg
```
BanzaiMan: Note that this line will export `SAUCE_DIRECT_DOMAINS` as an empty string, since `''` is a truth-y value. (This is not a problem here, I'm just pointing it out.)

Perhaps we can roll these 2 statements into one `if` statement to make the logic clearer:

```ruby
if direct_domains
  sh.export 'SAUCE_DIRECT_DOMAINS', "-D #{direct_domains}", echo: false
end
```

ngokevin: Thanks, simplified it to not worry about exporting empty strings.

ngokevin:
Thanks! :)

PR #455, opened 2015-05-19
Dump /etc/hosts content before and after applying addon

BanzaiMan:
https://github.com/travis-ci/travis-ci/issues/3922 describes reports of problems with empty `/etc/hosts`. We are trying to figure out if the hosts addon is to blame.


meatballhat:
My only nitpick is that I'd change the fold names and/or make the debug behavior optional (somehow?)  :sheep: :it:

meatballhat:
.... change the fold names to `hosts.before` and `hosts.after`, respectively, that is

PR #458, opened 2015-05-27
2-step hosts file modification for hosts addon

BanzaiMan:
This PR breaks down the hosts addon into 2 steps; one to construct the content and another to actually overwrite the content of `/etc/hosts`. My theory here is that the race condition exists in the current 2 single pipeline in stu editing of `/etc/hosts` that manifests in an empty file.

See https://github.com/travis-ci/travis-ci/issues/3922.


BanzaiMan:
Tested in staging: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/430599#L85

svenfuchs:
Looks good to me!

PR #459, opened 2015-06-01
add more explanatory output about guessed package name

tkelman:
and the condition when the default test script is empty

for https://github.com/travis-ci/travis-ci/issues/3988

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -59,6 +59,8 @@ module Travis
         def script
           sh.echo 'Executing the default test script', ansi: :green
           set_jl_pkg
+          sh.echo 'Package name determined from repository url to be ${JL_PKG}',
+            ansi: :green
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/julia.rb
```diff
@@ -71,6 +73,11 @@ def script
                 '-e "Pkg.test(\"${JL_PKG}\", coverage=true)"'
             end
           end
+          sh.else do
+            sh.echo '\`src/${JL_PKG}.jl\` not found, repository is not a '\
+              'valid Julia package so the default test script is empty',
+              ansi: :green
```
BanzaiMan: Perhaps another color (`:yellow`) may draw more attention here? (Not sure if this is necessarily an error condition, but it may not be something the user is expecting.)

tkelman: sure

PR #461, opened 2015-06-03
Change elixir installation process

BanzaiMan:
Resolves https://github.com/travis-ci/travis-ci/issues/3269.

If the desired Elixir version is not installed already,
pull down package from s3.hex.pm, and set up Elixir runtime
as `kiex` expects.

`kiex list` does not make network calls to GitHub, so that
we should not reach GitHub's API rate limit.


BanzaiMan:
https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/430703

PR #462, opened 2015-06-04
Set GOMAXPROCS if it's not set by config

BanzaiMan:
This reduces the chance of Go running completely wild

### DIFF
--- a/lib/travis/build/script/go.rb
+++ b/lib/travis/build/script/go.rb
@@ -20,7 +20,7 @@ module Travis
           super
           sh.export 'TRAVIS_GO_VERSION', go_version, echo: false
           sh.if '-z $GOMAXPROCS' do
-            sh.export 'GOMAXPROCS', '$(nproc 2>/dev/null || sysctl -n hw.physicalcpu 2>/dev/null || echo 2)', echo: true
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/go.rb
```diff
@@ -19,6 +19,9 @@ class Go < Script
         def export
           super
           sh.export 'TRAVIS_GO_VERSION', go_version, echo: false
+          sh.if '-z $GOMAXPROCS' do
+            sh.export 'GOMAXPROCS', '2', echo: true
```
meatballhat: I think this should be `GOMAXPROCS=$(nproc 2>/dev/null || echo 2)`

meatballhat: It could be made more cross-platform friendly a la:

```bash
nproc 2>/dev/null || sysctl -a | awk '/machdep\.cpu\.core_count/ { print $NF }' 2>/dev/null || echo 2
```

BanzaiMan: `sysctl -n hw.physicalcpu`

meatballhat: Yay way better!

BanzaiMan:
I wonder if this should be `echo`ed (since it's an unwieldy command that doesn't indicate concrete value). The user may think, "What is it set to?". If the user wants to set it to a different value, they'd know what they are up against. Maybe it's not necessary to `echo`.

meatballhat:
:ship: :it:

PR #466, opened 2015-06-13
Fix `edge: local`

jasny:
Ruby gems v1.x the `--local` flag would scan the current working directory for gems of any version. However ruby gems v2.x no longer does this and requires you to use the file name of the gem.

This change modifies the command to

    gem install dpl-*.gem --local --pre

For more information see the question [How can I run a custom version of dpl on Travis-CI?](http://stackoverflow.com/questions/30584467/how-can-i-run-a-custom-version-of-dpl-on-travis-ci) on Stack Overflow.


BanzaiMan:
This has been deployed to production. Please test it.

PR #468, opened 2015-06-26
Compress cache with gzip

BanzaiMan:
When uploading cache, use gzip as the compression algorithm instead of bzip2.

And when fetching, look for `.tgz` archives before falling back to `.tbz`.


lib/travis/build/script/shared/directory_cache/s3.rb
```diff
@@ -94,11 +93,11 @@ def push
           end
 
           def fetch_url(branch = group)
-            url('GET', prefixed(branch), expires: fetch_timeout)
+            url('GET', prefixed(branch, '.tbz'), expires: fetch_timeout)
```
meatballhat: I'm assuming the intent here is to migrate to `.tgz` along the way, so won't this mismatch of downloading `.tbz`, then uploading `.tgz` only work the first time?

BanzaiMan: See https://github.com/travis-ci/casher/pull/12#discussion_r33323761 for a little more context. I think you're right that we need more thought here; it would be great to test it on staging.

meatballhat: :+1: got it

BanzaiMan:
https://github.com/travis-ci/casher/pull/12 has been updated, so that it can be deployed first, without this PR being merged.

PR #469, opened 2015-06-29
Add support for .nvmrc

hkjorgensen:
If projects support ```.nvmrc``` make sure travis use it during setup. Instead of maintaining multiple manifests for nodejs/iojs projects.

if ```node_js``` property is specified in ```.travis.yml``` use that version.
if ```.nvmrc``` file is present use that version
fallback to ```nvm install stable```

This enhancement has been discussed here and accepted: https://github.com/travis-ci/travis-ci/issues/3644 - but it was never added :-(

cc: @BanzaiMan

### DIFF
--- a/lib/travis/build/script/node_js.rb
+++ b/lib/travis/build/script/node_js.rb
@@ -13,8 +13,9 @@ module Travis
 
         def setup
           super
-          sh.cmd "nvm install #{version}"
-          sh.cmd 'npm config set spin false', echo: false, timing: false
+          convert_legacy_nodejs_config
+          nvm_install
+          npm_disable_spinner
           npm_disable_strict_ssl unless npm_strict_ssl?
           setup_npm_cache if use_npm_cache?
         end
@@ -47,13 +48,50 @@ module Travis
 
         private
 
+          def convert_legacy_nodejs_config
+            # TODO deprecate :nodejs
+            # some old projects use language: nodejs. MK.
+            if config[:nodejs] && !config[:node_js]
+              config[:node_js] = config[:nodejs]
+            end
+          end
+
+          def node_js?
+            !!config[:node_js]
+          end
+
           def version
             @version ||= begin
-              # TODO deprecate :nodejs
-              version = config[:node_js] || config[:nodejs] # some old projects use language: nodejs. MK.
+              version = config[:node_js]
               version = version.first if version.is_a?(Array) # it seems travis-core does not exand on nodejs anymore so we end up with an array here?
               version == 0.1 ? '0.10' : version.to_s
             end
+          end
+
+          def nvm_install
+            if node_js?
+              use_nvm_version
+            else
+              use_nvm_default
+            end
+          end
+
+          def use_nvm_default
+            sh.if '-f .nvmrc' do
+              sh.echo "Using nodejs version from .nvmrc", ansi: :yellow
+              sh.cmd "nvm install"
+            end
+            sh.else do
+              sh.cmd "nvm install 0.10"
+            end
+          end
+
+          def use_nvm_version
+            sh.cmd "nvm install #{version}"
+          end
+
+          def npm_disable_spinner
+            sh.cmd 'npm config set spin false', echo: false, timing: false
           end
 
           def npm_disable_strict_ssl
--- a/spec/build/script/node_js_spec.rb
+++ b/spec/build/script/node_js_spec.rb
@@ -17,8 +17,21 @@ describe Travis::Build::Script::NodeJs, :sexp do
     should include_sexp [:export, ['TRAVIS_NODE_VERSION', '0.10']]
   end
 
-  it 'sets up the node version' do
-    should include_sexp [:cmd, 'nvm install 0.10', echo: true, assert: true, timing: true]
+  describe 'nvm install' do
+    it 'sets the version from config :node_js' do
+      data[:config][:node_js] = 0.9
+      should include_sexp [:cmd, 'nvm install 0.9', assert: true, echo: true, timing: true]
+    end
+
+    it 'sets the version from .nvmrc' do
+      sexp = sexp_filter(subject, [:if, '-f .nvmrc'], [:then])
+      expect(sexp).to include_sexp [:cmd, 'nvm install', assert: true, echo: true, timing: true]
+    end
+
+    it 'sets the version to 0.10 otherwise' do
+      sexp = sexp_filter(subject, [:if, '-f .nvmrc'], [:else])
+      expect(sexp).to include_sexp [:cmd, 'nvm install 0.10', assert: true, echo: true, timing: true]
+    end
   end
 
   it 'announces node --version' do


BanzaiMan:
Thanks for the PR. I think this is a bit confusing, though, since `.nvmrc` now *overrides* what's in the config.

```yaml
node_js:
  - 0.12
  - 0.10
```

would now execute the same build twice if `.nvmrc` exists, without meaningful feedback. Presumably, the user knows what they are doing if they add `.nvmrc`, s I think that the presence of this file should be prominently announced, very much like we do now with `.ruby-version` file. (https://github.com/travis-ci/travis-build/blob/838eaad86e1651ed7bcceaeb22445058a60c7284/lib/travis/build/script/shared/rvm.rb#L86)

joshk:
Just like `.ruby-version`, if `node_js` has been specified in the `.travis.yml`, we will use that, if nothing is specified we will use the `.nvmrc`, does that sound fair?

hkjorgensen:
@BanzaiMan thanks for feedback - I agree. Was not 100% what solution would be, it could easily add confusion.

hkjorgensen:
@joshk would you mind taking a look at it? :-)

hkjorgensen:
@BanzaiMan can you help? :-)

hkjorgensen:
Guys don't leave me hanging :-)
:+1: / :-1:  ?

joshk:
Hi @hkjorgensen 

I'm sorry but I'm not able to help with this. I lead Product Management here at Travis and rely on the awesomeness of others to help make this happen.

I am in support of this feature, but right now @BanzaiMan is busy with some improvements we are making to caching.

That said, you have changed a few extra things which it would be good to change back, I will leave comments shortly.

Thanks a bundle

Josh

lib/travis/build/script/node_js.rb
```diff
@@ -2,10 +2,6 @@ module Travis
   module Build
     class Script
       class NodeJs < Script
-        DEFAULTS = {
```
joshk: please add this back

lib/travis/build/script/node_js.rb
```diff
@@ -47,15 +44,47 @@ def cache_slug
...
+
+          def node_js?
+            !!config[:node_js]
+          end
+
           def version
             @version ||= begin
-              # TODO deprecate :nodejs
-              version = config[:node_js] || config[:nodejs] # some old projects use language: nodejs. MK.
+              version = config[:node_js] || 'stable'
```
joshk: I would prefer not to default to `stable` just yet. We need to discuss any implications this may have with the team.

lib/travis/build/script/node_js.rb
```diff
@@ -13,8 +9,9 @@ def export
 
         def setup
           super
-          sh.cmd "nvm install #{version}"
-          sh.cmd 'npm config set spin false', echo: false, timing: false
+          convert_legacy_nodejs_config if config[:nodejs] && !config[:node_js]
```
joshk: you can probably move the checks within the method

lib/travis/build/script/node_js.rb
hkjorgensen: Bringing this back means there is always a ```config[:node_js]``` property(?) so if the user didn't specify a version it will be set to:
```
node_js:
  - 0.10
```

So we will never end up searching for a ```.nvmrc``` file

hkjorgensen:
Thanks for the feedback @joshk

lib/travis/build/script/node_js.rb
hkjorgensen: if you didn't specify a version in the ```.travis.yml``` manifest, you probably don't care that much :-)

hkjorgensen:
bump! @BanzaiMan

lib/travis/build/script/node_js.rb
```diff
@@ -47,15 +44,47 @@ def cache_slug
...
-              # TODO deprecate :nodejs
-              version = config[:node_js] || config[:nodejs] # some old projects use language: nodejs. MK.
+              version = config[:node_js] || 'stable'
               version = version.first if version.is_a?(Array) # it seems travis-core does not exand on nodejs anymore so we end up with an array here?
               version == 0.1 ? '0.10' : version.to_s
             end
           end
 
+          def nvm_install
+            return use_nvm_version if node_js?
```
BanzaiMan: This doesn't read well. When I read this, I'd expect `use_nvm_version` to return a value that is usable by callers of `nvm_install`, but what we really want is just adding suitable call to `sh.cmd`.

Perhaps

```ruby
if node_js?
  use_nvm_version
else
  use_nvm_default
end
```

lib/travis/build/script/node_js.rb
BanzaiMan: What is supposed to happen if `.nvmrc` does not exist and no value is set in `node_js` in `.travis.yml`?

lib/travis/build/script/node_js.rb
BanzaiMan: Sorry, but we disagree. Please change this.

lib/travis/build/script/node_js.rb
```diff
@@ -47,15 +44,47 @@ def cache_slug
...
+            use_nvm_default
+          end
+
+          def use_nvm_default
+            sh.if '-f .nvmrc' do
+              sh.echo "Using nodejs version from .nvmrc", ansi: :yellow
+              sh.cmd "nvm install"
+            end
+            sh.else do
+              sh.cmd "nvm install stable"
```
BanzaiMan: Should not default to `stable` here.

hkjorgensen:
Thanks for feedback @BanzaiMan - appreciate it.
If your not ready to move from 0.10 yet, I'll add a version const for nodejs v0.10.

Will update PR asap

lib/travis/build/script/node_js.rb
BanzaiMan: `config[:node_js]` property is the value given in `.travis.yml`. If there is no `node_js` set in `.travis.yml`, it is `nil`.

I _think_ nvm would use whatever is the "default" it thinks if we drop this, but I am not sure. It is safe to have a default value, though, so please add this back.

hkjorgensen:
@BanzaiMan can you help me with the final tests, then it should be ready to go :-)

lib/travis/build/script/node_js.rb
```diff
@@ -47,15 +48,52 @@ def cache_slug
...
+          def convert_legacy_nodejs_config
+            # TODO deprecate :nodejs
+            # some old projects use language: nodejs. MK.
+            if config[:nodejs] && !config[:node_js]
+              config[:node_js] = config[:nodejs]
+            end
+          end
+
+          def node_js?
+            !!config[:node_js]
```
BanzaiMan: I think our insistence to reinstate the `DEFAULTS` hash is the source of the spec failures in https://travis-ci.org/travis-ci/travis-build/builds/70773325.

In the case of RVM and Ruby, we set the version in the `DEFAULTS` hash to be `default`, which RVM accepts; if that is the Ruby version we see in `travis-build` & the file `.ruby-version` exists, we use it to set the Ruby version.

Removing `DEFAULTS` will make the failing specs in https://travis-ci.org/travis-ci/travis-build/builds/70773325 pass.

The problem here is that, without the `DEFAULT` hash and the `node_js` value present in `.travis.yml`, the environment variable `$TRAVIS_NODE_VERSION` will not be set. And the spec checking that will fail. I am not sure yet how we can reconcile all of these issues.

(Incidentally, `$TRAVIS_RUBY_VERSION` will be set to `default` if `.travis.yml` does not define the Ruby version for Ruby builds. This, too, is a problem.)

BanzaiMan:
I think our insistence to reinstate the `DEFAULTS` hash is the source of the spec failures in https://travis-ci.org/travis-ci/travis-build/builds/70773325.

In the case of RVM and Ruby, we set the version in the `DEFAULTS` hash to be `default`, which RVM accepts; if that is the Ruby version we see in `travis-build` & the file `.ruby-version` exists, we use it to set the Ruby version.

Removing `DEFAULTS` will make the failing specs in https://travis-ci.org/travis-ci/travis-build/builds/70773325 pass.

The problem here is that, without the `DEFAULT` hash and the `node_js` value present in `.travis.yml`, the environment variable `$TRAVIS_NODE_VERSION` will not be set. And the spec checking that will fail. I am not sure yet how we can reconcile all of these issues.

(Incidentally, `$TRAVIS_RUBY_VERSION` will be set to `default` if `.travis.yml` does not define the Ruby version for Ruby builds. This, too, is a problem.)

joshk:
@hkjorgensen Thanks for all your work on this. I would be fine with merging this into a branch and making any of the required changes, if you prefer?

hkjorgensen:
That would be @joshk - thanks.

joshk:
@BanzaiMan would you mind taking care of this?

BanzaiMan:
@hkjorgensen This has been deployed to staging.

We will need a repo set up on https://staging.travis-ci.org. If you have one set up on production (https://travis-ci.org), you can switch it to staging, too (and switch it back to production when we are done testing).

hkjorgensen:
@BanzaiMan I don't have a public repository with this atm.

There should just be a .nvmrc with a node or iojs version to test :-)
```echo node v0.12 > .nvmrc```

BanzaiMan:
With https://github.com/travis-ci/travis-build/commit/2e824511bff3803543ec3cf5509378c8b830104e and the correct `.nvmrc` format ("v0.12"), we install Node.js 0.12.x correctly, but `$TRAVIS_NODE_VERSION` is unset: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/433815

Everything is working as expected. The question is, what is the desired behavior?

hkjorgensen:
> Everything is working as expected. The question is, what is the desired behavior?

- 1. If node version is defined in ```.travis.yml``` use that
- 2. If node version isn't defined in ```.travis.yml``` but there is a ```.nvmrc``` use that
- 3. If node version isn't defined in ```.travis.yml``` and there aren't any ```.nvmrc``` fall back to travis default. In this case ```node v0.10```

BanzaiMan:
See https://github.com/travis-ci/travis-build/commit/7a5396a1540adf6e90f8c6ce43acbf2986b76682

BanzaiMan:
The above commit reads the content of `.nvmrc` verbatim, which is probably not a good idea. We need to fix that.

BanzaiMan:
Actually, there is no need, since `nvm` will fail to install if the user puts extra stuff in `.nvmrc`.

BanzaiMan:
Deployed and documented. http://docs.travis-ci.com/user/languages/javascript-with-nodejs/#Using-.nvmrc

hkjorgensen:
Thanks guys!

PR #471, opened 2015-07-08
When Coverity Scan quota is not authorized, do not report build fail

aceofspades:


solarce:
Thanks, we'll review this and see about getting it deployed soon

SecUpwN:
Happy to know that you guys collaborate so well when it comes to improving your services. That is *exactly* how open source should work, I really enjoy having been in touch with both if you. Rock on! :+1:

lib/travis/build/addons/coverity_scan.rb
```diff
@@ -51,7 +51,7 @@ def authorize_quota
   else
     WHEN=`echo $AUTH_RES | ruby -e "require 'rubygems'; require 'json'; puts JSON[STDIN.read]['next_upload_permitted_at']"`
     echo -e "\033[33;1mCoverity Scan analysis NOT authorized until $WHEN.\033[0m"
-    exit 1
+    exit 0
```
BanzaiMan: Since we are exiting with `0` in both cases of the `if` statement, we should remove that logic from `src`, and add `assert: false` to `sh.raw` below.

This might require changes in the specs (untested).

aceofspades: @BanzaiMan per the nested if on L44 there's exit 1, fallthrough, and exit 0... are we looking at the same thing?

BanzaiMan: @aceofspades Ah. You are right.

PR #472, opened 2015-07-10
Export env vars before running addons

BanzaiMan:
Fixes https://github.com/travis-ci/travis-ci/issues/4173

### DIFF
--- a/lib/travis/build/stages.rb
+++ b/lib/travis/build/stages.rb
@@ -8,7 +8,7 @@ module Travis
   module Build
     class Stages
       STAGES = [
-        :builtin,     [:header, :configure, :checkout, :prepare, :disable_sudo, :export, :setup, :announce],
+        :builtin,     [:header, :configure, :checkout, :export, :prepare, :disable_sudo, :setup, :announce],
         :custom,      [:before_install, :install, :before_script, :script, :before_cache],
         :builtin,     [:cache],
         :conditional, [:after_success],


joshk:
I'm sorry but this is a security concern and can not be merged in. Someone can write an end var which executes a shell command which uses sudo before we disable it.

BanzaiMan:
Yeah, you're right. We can't do this.

phadej:
Couldn't env vars be sanitised? Haven't seen (or needed) any non alphanumeric+dots/dashes envs. And non-safe ones could be exported after `:disable_sudo` if someone really needs them?

couchand:
It looks like the env vars are already being sanitized here: https://github.com/travis-ci/travis-build/blob/1ac9f14d7f0f1139e7ccd6c0a32225134a08faea/lib/travis/build/env/var.rb#L5

    PATTERN = /(?:SECURE )?([\w]+)=(("|')(.*?)(\3)|\$\(.*?\)|[^"' ]+)/

Unfortunately that regex doesn't look like it's doing much cleaning.  Note the second case allows for a construct like `$(sudo something --dangerous)`, which is clearly not ideal.  What is the intended use-case for allowing shell-interpolation in environment variables at all?

It seems to me that using env vars in the apt setup (per travis-ci #4173 and #4291) is far more valuable than some hypothetical need to run a command at export time.  Am I missing something?

joshk:
I'm sorry but there is a real security risk for our infrastructure as someone escaping a container could in theory gain access to our underlying host, and possibly under running jobs on the host.

So for the time being at least, we will not be exporting env vars before running the apt addons.

Thank you for the comments and feedback on this though.

PR #473, opened 2015-07-13
Add directories to cache in one pass

BanzaiMan:
To avoid scanning the same archive multiple times to extract
directories, we pass directories to 'casher add' in one go.

### DIFF
--- a/lib/travis/build/script/shared/directory_cache/s3.rb
+++ b/lib/travis/build/script/shared/directory_cache/s3.rb
@@ -30,6 +30,10 @@ module Travis
                   time_total:  %{time_total} s
           EOF
 
+          # maximum number of directories to be 'added' to cache via casher
+          # in one invocation
+          ADD_DIR_MAX = 100
+
           KeyPair = Struct.new(:id, :secret)
 
           Location = Struct.new(:scheme, :region, :bucket, :path) do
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/shared/directory_cache/s3.rb
```diff
@@ -77,8 +77,8 @@ def install
             end
           end
 
-          def add(path)
-            run('add', path) if path
+          def add(*paths)
+            run('add', paths) if paths
```
meatballhat: What about command line arg limit?  Can you do batches of 200?

lib/travis/build/script/shared/directory_cache/s3.rb
```diff
@@ -77,8 +81,10 @@ def install
             end
           end
 
-          def add(path)
-            run('add', path) if path
+          def add(*paths)
+            if paths
+              paths.flatten.each_slice(ADD_DIR_MAX) { |dirs| run('add', dirs) }
```
meatballhat: :heart_eyes_cat:

BanzaiMan:
Casher change has been deployed to production. I plan to deploy this early tomorrow.

I'm a bit ambivalent about documenting this behavior; on one hand, it is good to speed up extraction via `cache.directories`. On the other hand, this makes other caching niceties (such as `cache: bundler`) less attractive if it is used in conjunction with `cache.directories`. (It is pure win, but it discourages `cache: bundler` to some extent.)

BanzaiMan:
This was merged with https://github.com/travis-ci/travis-build/commit/c09d12807c2f003cebebd5aebb9f8b0a84305552.

PR #474, opened 2015-07-14
Ensure "travis" user and db are present for all PostgreSQL versions

meatballhat:
The good news is that the existing addon Just Works™ even when installing PostgreSQL from the pgdg repo.  Cycling through both ports 5432 and 5433 is a bit of a hack, as installing previously absent versions of PostgreSQL while there's a server still running will cause the installer to select the next unused port, which is 5433.

Working build in staging: https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/builds/433553
Associated config: https://github.com/meatballhat/yolo-octo-adventure/blob/5b2db2f53e2a8dbd599f067b8dc85addaa26aae0/.travis.yml

### DIFF
--- a/lib/travis/build/addons/postgresql.rb
+++ b/lib/travis/build/addons/postgresql.rb
@@ -17,8 +17,8 @@ module Travis
             end
             sh.cmd "service postgresql start #{version}", assert: false, sudo: true, echo: true, timing: true
             %w(5432 5433).each do |pgport|
-              sh.raw "sudo -u postgres createuser -p #{pgport} travis", assert: false, echo: true, timing: true
-              sh.raw "sudo -u postgres createdb -p #{pgport} travis", assert: false, echo: true, timing: true
+              sh.raw "sudo -u postgres createuser -p #{pgport} travis &>/dev/null", assert: false, echo: true, timing: true
+              sh.raw "sudo -u postgres createdb -p #{pgport} travis &>/dev/null", assert: false, echo: true, timing: true
             end
           end
         end
--- a/lib/travis/build/addons/postgresql.rb
+++ b/lib/travis/build/addons/postgresql.rb
@@ -9,13 +9,17 @@ module Travis
 
         def after_prepare
           sh.fold 'postgresql' do
-            sh.export "PATH", "/usr/lib/postgresql/#{version}/bin:$PATH", echo: false
+            sh.export 'PATH', "/usr/lib/postgresql/#{version}/bin:$PATH", echo: false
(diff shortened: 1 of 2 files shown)


lib/travis/build/addons/postgresql.rb
```diff
@@ -9,13 +9,17 @@ class Postgresql < Base
...
-            sh.export "PATH", "/usr/lib/postgresql/#{version}/bin:$PATH", echo: false
+            sh.export 'PATH', "/usr/lib/postgresql/#{version}/bin:$PATH", echo: false
             sh.echo "Starting PostgreSQL v#{version}", ansi: :yellow
-            sh.cmd "service postgresql stop", assert: false, sudo: true, echo: true, timing: true
+            sh.cmd 'service postgresql stop', assert: false, sudo: true, echo: true, timing: true
             sh.if "-d /var/ramfs && ! -d /var/ramfs/postgresql/#{version}", echo: false do
               sh.cmd "cp -rp /var/lib/postgresql/#{version} /var/ramfs/postgresql/#{version}", sudo: true, assert: false, echo: false, timing: false
             end
             sh.cmd "service postgresql start #{version}", assert: false, sudo: true, echo: true, timing: true
+            %w(5432 5433).each do |pgport|
```
BanzaiMan: Why do we need 5433?

meatballhat: Take a peek at the PR description :wink:

I gave the ports names for clarity, too.

BanzaiMan:
Hmm. I am not quite sure this is the right thing to do, though. The problem is running the installer when an instance is already running. Isn't it more appropriate to figure out *why* we have PostgreSQL listening on a non-default port and eradicate that possibility?

Also, what stops a third PostgreSQL instance from being configured?

We can brute force `createuser` and `createdb` for these ports, but isn't it wiser to scan the configuration files and find the defined ports instead?

meatballhat:
@BanzaiMan The problem is doing this on container-based where we don't (and probably shouldn't) change the running state of services based on what packages are being installed.  The installation of PostgreSQL while an existing server is running causes the new installation to be configured with the next available port after the default of 5432.  This is expected behavior.  Nothing is stopping more PostgreSQL instances from being configured.  The intent of this PR is to cover a known use case of installing PostgreSQL 9.5.  Scanning config files and changing the running state of services is certainly possible, but this solution is what I know to work given the current PostgreSQL and APT addons on container-based.

BanzaiMan:
OK.

BanzaiMan:
This PR might be the cause of travis-ci/travis-ci#4570.

PR #476, opened 2015-07-16
Fixes to `fix_resolv_conf` appliance

sarahhodne:
- Remove OpenDNS
- Remove "dead" code testing for BB DNS servers (this was being run on Heroku, so it would never run)

Essentially we are now left with an addon that just overwrites with Google DNS. To my knowledge this appliance is only run on Blue Box VMs.


meatballhat:
No longer applicable?  Probably?

PR #478, opened 2015-07-17
Fix git tarball strategy for private repos

BanzaiMan:
See https://github.com/travis-ci/travis-ci/issues/3923


GrahamCampbell:
Downloading a tar to run tests is a bad idea.

GrahamCampbell:
People often exclude their tests folder from the archives.

joshk:
@GrahamCampbell this tar is from GitHub, and is just a shallow clone of the repo.

GrahamCampbell:
Normally when you download tars from github, things like tests are excluded. For example, try downloading laravel/framework, and you'll see what I mean.

GrahamCampbell:
https://github.com/laravel/framework/archive/5.1.tar.gz

BanzaiMan:
@GrahamCampbell Interesting. Is this behavior documented somewhere on GitHub? Do keep in mind that this is an opt-in, so we expect the user to know what they are opting in to.

GrahamCampbell:
Not sure if's documented, but it's done when users mark files as excluded to save on download size of archives. These archives are exclusively intended for distribution for use, which is why people exclude their tests. They're not at all meant for this purpose.

BanzaiMan:
So, maintainers of the repository *chooses* to exclude them from archive, right? I tend to think this is not a concern for the purpose of this PR.

GrahamCampbell:
The point I'm getting at is that you're misusing these archives if use them in this way.

BanzaiMan:
What purposes are these archives supposed to serve? It is for convenient distribution of the source code, no? It seems to me that we can use them for testing if the maintainers choose to do so.

GrahamCampbell:
> It is for convenient distribution of the source code, no?

For distribution for actual use, not to run it's own tests.

BanzaiMan:
@GrahamCampbell I am afraid that we are not going to agree on this.

I believe that running test constitutes "actual use". Whether or not the archive includes tests is completely up to the maintainers (who decides what's in `.gitconfigure`); if the tests are in the archive, I believe that the archive is fit for running tests. When the maintainers of the repository decides to use the git tarball strategy to run the tests on Travis, they have decided that it is a good idea to do so. If you, a laravel/framework maintainer, don't want to use the tarball strategy to get the source code for your tests on Travis CI, you don't have to.

GrahamCampbell:
> I believe that running test constitutes "actual use".

But it doesn't? That is literally the definition. The definition of that archive is to exlude the tests and anything unrelated to public use of the package.

BanzaiMan:
It is your archive's definition. Not someone else's. As I said before, this is an opt-in strategy.

GrahamCampbell:
It's the definition everyone uses. All package managers use that definition for example.

BanzaiMan:
Thanks for excluding me from "everyone".

Package managers most likely just need the binaries (some have "source" packages, yes), and their view is drastically different from CI platforms, which need to actually run tests. It stands to reason that the optimal "archives" for package managers and the CI platforms are different.

You are welcome to open an issue to discuss whether using tarball strategy to pull down source is a good idea, but that is not the purpose of this PR.

deckar01:
The documentation for [`git archive`](https://git-scm.com/docs/git-archive) does not mention any specific use cases for excluding files or archiving in general. Although GitHub has exposed this feature as a method for distributing releases, the git protocol allows an archive to be requested for any commit, which suggests a general design that does not exclude development work flows.

@BanzaiMan This seems like a straightforward bug fix. Is there anything that can be done here to get this merged or closed?

meatballhat:
Now obsolete!

PR #483, opened 2015-07-25
announce version with addon: mariadb

grooverdan:


grooverdan:
sorry, i'm a bit lost of writing ruby test cases.

BanzaiMan:
No worries. Addons don't get their `#announce` invoked, so we'll have to figure something out if we are to do something about it. It's getting late here now, but I'll take a look at this some time next week.

joshk:
Can we just announce the version installed when we install it?

BanzaiMan:
I added a command to announce this at the end of the addon invocation. See @940fd45.

grooverdan:
@940fd45 is incorrect as it announces the mysql client version not the server version.

PR #484, opened 2015-07-25
export TRAVIS_MARIADB_VERSION for addons: mariadb

grooverdan:
closes travis-ci/travis-ci/issues/4553


BanzaiMan:
We don't do this for other DBs. Should we?

(Needs a rebase.)

grooverdan:
> We don't do this for other DBs. Should we?

seems like an easy thing to do and yes I think its useful.

> (Needs a rebase.)

done

grooverdan:
grep -r TRAVIS_[A-Z]*_VERSION 

seems like all languages export a version. When there are multiple database version possible it makes sense that this is exported too.

PR #487, opened 2015-07-25
Add support for requesting more specific Julia versions

tkelman:
closes https://github.com/JuliaCI/travis-build/issues/2

I can tweak the docs accordingly if this looks good.

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -83,21 +83,29 @@ module Travis
         private
 
           def julia_url
+            case config[:os]
+            when 'linux'
+              status = 'linux-x86_64'
+              osarch = 'linux/x64'
+              ext = "#{status}.tar.gz"
+            when 'osx'
+              status = 'osx10.7+'
+              osarch = 'osx/x64'
+              ext = "#{status}.dmg"
+            end
             case config[:julia]
             when 'release'
-              version = 'stable'
+              url = "status.julialang.org/stable/#{status}"
             when 'nightly'
-              version = 'download'
+              url = "status.julialang.org/download/#{status}"
+            when /^(\d+\.\d+)\.\d+$/
+              url = "s3.amazonaws.com/julialang/bin/#{osarch}/#{$1}/julia-#{config[:julia]}-#{ext}"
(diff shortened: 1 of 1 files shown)


staticfloat:
This looks good to me!

tkelman:
Thanks @staticfloat!

There was some feedback from @ninjin on an initial version of this at https://github.com/JuliaCI/travis-build/commit/96f45aa434bd3a9f7c358ccdff2549fe8746fff4, though I've backed out the 32-bit-related parts of that change for the purposes of this PR (we should have more of a discussion on that with the Travis team first).

A few things that might be useful to have on the Julia side would be static URL's that tell us what the most recent release version is (so we could switch to s3.amazonaws.com/julialang for `release`, in case of any connectivity issues at status.julialang.org). And another for the version number of `nightly`, so we could allow users to ask for sha versions to get recent but not-latest nightlies from s3.amazonaws.com/julianightlies - though maybe https://raw.githubusercontent.com/JuliaLang/julia/master/VERSION would be good enough for that.

staticfloat:
What use cases are the static URLs you're looking for?  Something that can tell us the most recent release without having to know the major version number?  (Because if you know the major version number, we already have the `0.3-latest` and whatnot?)

tkelman:
Maybe it's not so important, the thought is to keep `release` and `nightly` working since lots of packages are already using those, but move the links to S3 if possible. We'd need to know the major.minor to do that right now. Since status.julialang.org has been pretty well-behaved recently, changing the links to S3 might not be necessary.

And since looking up what recent-but-not-latest nightlies are available on S3 already requires manual intervention, maybe it isn't worth the trouble of trying to make that work on Travis. Certainly could be left for later, the change made in this PR will be absolutely necessary before we release 0.4.0 so that people don't lose out on the ability to test with 0.3. There's even a chance this could resurrect testing on 0.2.0 and 0.2.1, though we only have OSX and Windows binaries for those versions.

BanzaiMan:
Is this ready for merge?

tkelman:
If it looks okay to you, then yes, thanks! I'll prepare the doc changes.

PR #488, opened 2015-07-25
Add OS X support for C#

Joshua-Anderson:
@akoeplinger @nterry

Some people have been requesting OS X support for C#, so here it is!

 Fixes travis-ci/travis-ci#3546

### DIFF
--- a/lib/travis/build/script/csharp.rb
+++ b/lib/travis/build/script/csharp.rb
@@ -24,22 +24,29 @@ module Travis
           sh.fold('mono-install') do
             if mono_version_valid?
               sh.echo 'Installing Mono', ansi: :yellow
-
-              if is_mono_2_10_8
-                sh.cmd 'sudo apt-get update -qq', timing: true, assert: true
-                sh.cmd 'sudo apt-get install -qq mono-complete mono-vbnc', timing: true, assert: true
-              elsif is_mono_3_2_8
-                sh.cmd 'sudo apt-add-repository ppa:directhex/ppa -y', assert: true # Official ppa of the mono debian maintainer
-                sh.cmd 'sudo apt-get update -qq', timing: true, assert: true
-                sh.cmd 'sudo apt-get install -qq mono-complete mono-vbnc fsharp', timing: true, assert: true
-              else
-                sh.cmd 'sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF', echo: false, assert: true
-                mono_repos.each do |repo|
-                  sh.cmd "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian #{repo} main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", echo: false, assert: true
+              case config[:os]
+              when 'linux'
+                if is_mono_2_10_8
+                  sh.cmd 'sudo apt-get update -qq', timing: true, assert: true
+                  sh.cmd 'sudo apt-get install -qq mono-complete mono-vbnc', timing: true, assert: true
+                elsif is_mono_3_2_8
+                  sh.cmd 'sudo apt-add-repository ppa:directhex/ppa -y', assert: true # Official ppa of the mono debian maintainer
+                  sh.cmd 'sudo apt-get update -qq', timing: true, assert: true
+                  sh.cmd 'sudo apt-get install -qq mono-complete mono-vbnc fsharp', timing: true, assert: true
+                else
+                  sh.cmd 'sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 3FA7E0328081BFF6A14DA29AA6A19B38D3D831EF', echo: false, assert: true
+                  mono_repos.each do |repo|
+                    sh.cmd "sudo sh -c \"echo 'deb http://download.mono-project.com/repo/debian #{repo} main' >> /etc/apt/sources.list.d/mono-xamarin.list\"", echo: false, assert: true
+                  end
+                  sh.cmd 'sudo apt-get update -qq', timing: true, assert: true
+                  sh.cmd "sudo apt-get install -qq mono-complete mono-vbnc fsharp nuget #{'referenceassemblies-pcl' if !is_mono_3_8_0}", timing: true, assert: true
+                                                                                        # PCL Assemblies only supported on mono 3.10 and greater
                 end
-                sh.cmd 'sudo apt-get update -qq', timing: true, assert: true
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/csharp.rb
```diff
@@ -24,22 +24,30 @@ def configure
...
+                  end
+                  sh.cmd 'sudo apt-get update -qq', timing: true, assert: true
+                  sh.cmd "sudo apt-get install -qq mono-complete mono-vbnc fsharp nuget #{'referenceassemblies-pcl' if !is_mono_3_8_0}", timing: true, assert: true
+                                                                                        # PCL Assemblies only supported on mono 3.10 and greater
                 end
-                sh.cmd 'sudo apt-get update -qq', timing: true, assert: true
-                sh.cmd "sudo apt-get install -qq mono-complete mono-vbnc fsharp nuget #{'referenceassemblies-pcl' if !is_mono_3_8_0}", timing: true, assert: true
-                                                                                      # PCL Assemblies only supported on mono 3.10 and greater
+              when 'osx'
+                sh.cmd 'mkdir "$HOME/mono-tmp"', assert: true, echo: false
```
akoeplinger: Can't we just use /tmp here?

martincostello:
*Like*

lib/travis/build/script/csharp.rb
```diff
@@ -105,12 +112,29 @@ def mono_repos
...
+        end
+
         def mono_version_valid?
           return true if ['latest', 'alpha', 'beta', 'nightly'].include? config[:mono]
           return false unless MONO_VERSION_REGEXP === config[:mono]
 
-          return false if MONO_VERSION_REGEXP.match(config[:mono])[1] == '2' && !is_mono_2_10_8
-          return false if MONO_VERSION_REGEXP.match(config[:mono])[1].to_i < 2
+          return false if MONO_VERSION_REGEXP.match(config[:mono])[1] == '2' && !is_mono_2_10_8 && config[:os] == 'linux'
+          return false if MONO_VERSION_REGEXP.match(config[:mono])[1].to_i < 2 && config[:os] == 'linux'
```
akoeplinger: Hmm, do we really want to allow using Mono < 2 on Mac? I can't see this being useful and will create just additional confusion.

spec/build/script/csharp_spec.rb
```diff
@@ -13,6 +13,7 @@
 
   describe 'configure' do
     it 'sets up package repository' do
+      data[:config][:os] = 'linux'
```
akoeplinger: these changes tell me that we need to add a default value for os=linux, otherwise every existing .travis.yml out there will break

lib/travis/build/script/csharp.rb
Joshua-Anderson: I figured we would let people work that out for themselves, since mono 1 & 2 are still on the archive and wouldn't take anything extra to install.

akoeplinger: Does Mono 1 even install/run on a recent version of OSX? I'd rather get rid of it tbh, but if it still works I'm fine with it too.

spec/build/script/csharp_spec.rb
Joshua-Anderson: I believe travis-builder sets a default `config[:os]` of `linux`.

Joshua-Anderson: Other languages depend on this behavior as well, such as [r](https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/r.rb).

akoeplinger:
Looks great, thanks :+1:

Do you think we should hold off on merging this until the symlinks are in place?

Joshua-Anderson:
That's fine with me. An alternative is we could merge this and test it but only document it when the all symlinks are setup.

spec/build/script/csharp_spec.rb
akoeplinger: Ok, so it's just the unit test environment that doesn't? If the unit tests set it too, we should revert these changes to ensure we're as close to the real world as possible ;)

akoeplinger:
>  An alternative is we could merge this and test it but only document it when the all symlinks are setup.

Sounds good.

akoeplinger:
@Joshua-Anderson there you go:

- [http://download.mono-project.com/archive/mdk-latest.pkg](http://download.mono-project.com/archive/mdk-latest.pkg)
- [http://download.mono-project.com/archive/mdk-latest-alpha.pkg](http://download.mono-project.com/archive/mdk-latest-alpha.pkg)
- [http://download.mono-project.com/archive/mdk-latest-beta.pkg](http://download.mono-project.com/archive/mdk-latest-beta.pkg)
- [http://download.mono-project.com/archive/mdk-latest-weekly.pkg](http://download.mono-project.com/archive/mdk-latest-weekly.pkg)

martincostello:
Happy to give this a try out with my repo once this is available to use.

akoeplinger:
@Joshua-Anderson merged the other PR, can you update this one with the "weekly" changes?

Joshua-Anderson:
@akoeplinger Done!

spec/build/script/csharp_spec.rb
```diff
@@ -125,4 +126,37 @@
...
+
+    it 'selects beta' do
+      data[:config][:os] = 'osx'
+      data[:config][:mono] = 'beta'
+      should include_sexp [:cmd, "curl -o \"/tmp/mdk.pkg\" -L http://download.mono-project.com/archive/mdk-latest-beta.pkg", timing: true, assert: true]
+    end
+
+    it 'selects weekly' do
+      data[:config][:os] = 'osx'
+      data[:config][:mono] = 'nightly' # TODO: Change to weekly once #482 is merged
```
akoeplinger: nit: this can be removed now

akoeplinger:
I just tried this and it works fine: https://travis-ci.org/akoeplinger/travis-mono-test :smile: :tada:  

The only small gotcha is that both Mono 3.2.8 and 2.10.8 don't exist as such on the download server, which is unfortunate since we explicitly call them out in the docs. I'll see if we can get them added to the server.

martincostello:
:+1:

I'm waiting for multi-OS to be enabled on my repo to try this out for myself.

martincostello:
Yep, working for me now too. I just had to force a commit to make it take effect as the "Rebuild commit" button on Travis doesn't take into account whether the multi-OS feature was enabled since the previous build.

lib/travis/build/script/csharp.rb
akoeplinger: @Joshua-Anderson I just tried this and it doesn't work since the folder structure is different. E.g. even for 2.10.1 it's http://download.mono-project.com/archive/2.10.1/macos-10-x86/3/ . I think we should match the Linux behavior of only allowing 2.10.8 for Mac as well once that package is on the download server.

PR #492, opened 2015-07-27
git: fix typo in rm_key

sarahhodne:
The key that we install is named ~/.ssh/id_rsa, not ~/.ssh/source_rsa.

### DIFF
--- a/lib/travis/build/git.rb
+++ b/lib/travis/build/git.rb
@@ -54,7 +54,7 @@ module Travis
         end
 
         def rm_key
-          sh.rm '~/.ssh/source_rsa', force: true, echo: false
+          sh.rm '~/.ssh/id_rsa', force: true, echo: false
         end
 
         def config
--- a/spec/build/git_spec.rb
+++ b/spec/build/git_spec.rb
@@ -5,7 +5,7 @@ describe Travis::Build::Git, :sexp do
   let(:script)  { Travis::Build::Script.new(payload) }
   subject       { script.sexp }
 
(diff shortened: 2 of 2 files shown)


joshk:
Please do not merge this as it is a change in behavior which some customers may be relying on.

henrikhodne:
If so we should probably just remove the step entirely?

Also, the key is still in the SSH agent, so it's still usable.

joshk:
Hmmmmmmmmm, ok, that is a good point that it is in the SSH agent.

@meatballhat @rkh @svenfuchs thoughts and opinions?

meatballhat:
I think we can safely remove this step.  Those users who really need the identity to be gone have to interact with `ssh-add` anyway.

joshk:
@meatballhat do you mean remove the code?

meatballhat:
@joshk sorry for ambiguity.  Yes, remove the code :smiley_cat:

joshk:
Sounds good to me.

deckar01:
For historical context, "source_rsa" is not a typo, it was the [original location that wrote the configured private key for git](https://github.com/travis-ci/travis-build/pull/10/files). At one point the [code to copy the source key was removed](https://github.com/travis-ci/travis-build/pull/13/files) then later [reimplemented using "id_rsa"](https://github.com/travis-ci/travis-build/commit/ade0d42e7b1d88fd596e34c788b52b1bebbe4bd1) without updating the key cleanup code.

joshk:
@henrikhodne can this PR be closed and a new one opened to remove the code which removes the ssh key?

meatballhat:
No longer valid!

PR #494, opened 2015-07-28
[WIP] chruby improvements

BanzaiMan:
This PR adds a few improvements in consistency when `ruby` key is set in `.travis.yml` (i.e., the user wishes to use chruby to mange Ruby runtimes), especially when both `ruby` and `rvm` keys are present.

The `rvm` key is removed from `DEFAULT`, so that the Ruby version is set up correctly when `ruby` is present.

Note also the new cache naming scheme: `chruby` is used instead of `rvm`.

### DIFF
--- a/lib/travis/build/script/ruby.rb
+++ b/lib/travis/build/script/ruby.rb
@@ -7,7 +7,6 @@ module Travis
     class Script
       class Ruby < Script
         DEFAULTS = {
-          rvm:     'default',
           gemfile: 'Gemfile'
         }
 
--- a/lib/travis/build/script/shared/rvm.rb
+++ b/lib/travis/build/script/shared/rvm.rb
@@ -20,7 +20,7 @@ module Travis
         def export
           super
           # for now, we take the rvm key if both are defined
-          # see also:`#version` and `#version_manager`
+          # see also:`#version` and `#cache_name_component`
           if rvm?
             sh.export 'TRAVIS_RUBY_VERSION', config[:rvm], echo: false
           elsif config[:ruby]
@@ -42,7 +42,7 @@ module Travis
         end
 
         def cache_slug
-          super.tap { |slug| slug << "--#{version_manager}-" << ruby_version.to_s }
+          super.tap { |slug| slug << "--#{cache_name_component}-" << ruby_version.to_s }
         end
 
         private
@@ -126,11 +126,11 @@ module Travis
             sh.cmd "rvm autolibs disable", echo: false, timing: false
           end
 
-          def version_manager
+          def cache_name_component
             if rvm?
               'rvm'
             elsif chruby?
-              'chruby'
+              'ruby'
             else
               'rvm'
             end
--- a/lib/travis/build/script/shared/rvm.rb
+++ b/lib/travis/build/script/shared/rvm.rb
@@ -42,7 +42,7 @@ module Travis
         end
 
         def cache_slug
-          super.tap { |slug| slug << "--#{version_manager}-" << ruby_version.to_s unless chruby? }
+          super.tap { |slug| slug << "--#{version_manager}-" << ruby_version.to_s }
         end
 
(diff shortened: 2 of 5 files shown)


lib/travis/build/script/ruby.rb
```diff
@@ -7,7 +7,6 @@ module Build
     class Script
       class Ruby < Script
         DEFAULTS = {
-          rvm:     'default',
```
henrikhodne: What will happen if neither `rvm` nor `ruby` is present in the .travis.yml?

BanzaiMan: It will pick up `DEFAULT_VERSION`, defined in `rvm.rb`.

lib/travis/build/script/shared/rvm.rb
```diff
@@ -15,30 +15,46 @@ module RVM
...
         def announce
           super
           sh.cmd 'ruby --version'
-          sh.cmd 'rvm --version' if rvm?
+          sh.cmd 'rvm --version' unless chruby?
         end
 
         def cache_slug
-          super.tap { |slug| slug << "--rvm-" << ruby_version.to_s if rvm? }
+          super.tap { |slug| slug << "--#{version_manager}-" << ruby_version.to_s }
```
joshk: i think we can just use `ruby` instead of `rvm` or `chruby`

lib/travis/build/script/shared/rvm.rb
```diff
@@ -15,30 +15,46 @@ module RVM
...
           super
-          sh.export 'TRAVIS_RUBY_VERSION', config[:rvm], echo: false if rvm?
+          # for now, we take the rvm key if both are defined
+          # see also:`#version` and `#version_manager`
+          if rvm?
+            sh.export 'TRAVIS_RUBY_VERSION', config[:rvm], echo: false
+          elsif config[:ruby]
+            sh.export 'TRAVIS_RUBY_VERSION', config[:ruby], echo: false
+          else
+            sh.export 'TRAVIS_RUBY_VERSION', DEFALUT_VERSION, echo: false
```
BanzaiMan: Here, the Ruby version is picked up as `default` when neither `rvm` nor `ruby` key is present.

lib/travis/build/script/shared/rvm.rb
BanzaiMan: That's reasonable, but it'll nullify all existing caches.

joshk:
Are there any feature or behavior changes in this PR that current users will notice?

BanzaiMan:
The current RVM users should not. The users of `chruby`, on the other hand, will see that the cache will be refreshed, because the cache has changed. (It used to be something like `cache--rvm-default--chruby-default--gemfile-Gemfile`; even when `ruby` was used)

lib/travis/build/script/shared/rvm.rb
joshk: hmmmmm, ok, we can leave this, but can we use `ruby` instead of chruby?

BanzaiMan: of course!

lib/travis/build/script/shared/rvm.rb
```diff
@@ -15,30 +15,46 @@ module RVM
...
         private
 
           def version
-            config[:rvm].to_s
+            if rvm?
+              config[:rvm].to_s
+            elsif chruby?
+              config[:ruby].to_s
+            else
+              DEFALUT_VERSION
```
henrikhodne: This has a typo, it should be `DEFAULT_VERSION`

henrikhodne: Looks like the typo is repeated in the rest of the file, so shouldn't cause errors.

BanzaiMan:
Became too stale now. Will revisit in the future.

PR #504, opened 2015-08-25
Install npm version specified by package.json

BanzaiMan:
Fixes https://github.com/travis-ci/travis-ci/issues/4653

`jq -e` exits with 1 if the desired element is `false` or `null`.

`jq -r` writes raw data to STDOUT; when the desired element is a string, print that (`STRING`) instead of formatting it in JSON (`"STRING"`).

If `npm install -g npm@$(jq -r .engines.npm package.json)` fails for whatever reason, the default version will be used.


masone:
I am `jq`-ing the npm version in my `before_install` hook manually and would welcome this feature! Can I help with anything to get this merged?

BanzaiMan:
I've reverted this PR. See https://github.com/travis-ci/travis-ci/issues/4653#issuecomment-169035161.

PR #508, opened 2015-09-02
Add a configurable "import path" for Go builds

tianon:
Wasn't sure what to name this new configuration option...  The `gobuild_args` has me confused because it seems like it should really be either `gobuildargs` or `go_build_args`.

I'm definitely more than happy to rebase/amend as desired though!

As for what this enables, it's most useful for "vanity import paths" (like [`pault.ag/go/debian`](http://pault.ag/go/debian) and `golang.org/x/net`, for example).  For repos that specify it, this will also smooth out forks running tests on the forked repo, since they'll be put into the proper import path so imports between packages in the same repo will work properly without modification.

This could also be expanded to use `go list` to check for https://golang.org/s/go14customimport and use that if no configuration is supplied (ie, something like `go list -e -f '{{.ImportComment}}'`).

### DIFF
--- a/lib/travis/build/script/go.rb
+++ b/lib/travis/build/script/go.rb
@@ -51,11 +51,11 @@ module Travis
           sh.export 'GOPATH', "#{HOME_DIR}/gopath", echo: true
           sh.export 'PATH', "#{HOME_DIR}/gopath/bin:$PATH", echo: true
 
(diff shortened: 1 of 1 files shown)


meatballhat:
:+1: I'm a fan!

/cc @mislav

PR #509, opened 2015-09-02
Use "gobuild_args" method consistently

tianon:
While working on #508 (and https://github.com/travis-ci/docs-travis-ci-com/pull/346) I noticed there was a method for `gobuild_args` that wasn't being used.  I've updated it here to be used instead of grabbing the `config` directly, but I'd also be happy to update this to just remove the method if you'd rather. :+1:


meatballhat:
:sheep: :it: LGTM

PR #511, opened 2015-09-08
Look for BUNDLE_GEMFILE if it's defined

BanzaiMan:
When looking for Gemfile, consult `$BUNDLE_GEMFILE` first,
which is an absolute path if it is set.

This sidesteps the issue if the build changes working directory
when we do not expect (e.g., in before_install).

A simple test case (assume files exist):

```yaml
language: ruby
gemfile: path/to/Gemfile
before_install: cd $(dirname $BUNDLE_GEMFILE)
script: ./test.sh
```

This has been tested on staging:

Before: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/435418
After: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/435420#L97 (Note that Gemfile is found, and `bundle install` is run.)

### DIFF
--- a/lib/travis/build/script/shared/bundler.rb
+++ b/lib/travis/build/script/shared/bundler.rb
@@ -46,11 +46,11 @@ module Travis
         private
 
           def gemfile?
-            "-f #{config[:gemfile]}"
+            "-f ${BUNDLE_GEMFILE:-#{config[:gemfile]}}"
           end
 
           def gemfile_lock?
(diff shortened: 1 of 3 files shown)


henrikhodne:
Should `config[:gemfile]` be shell-escaped?

BanzaiMan:
Should it be? It was not escaped before.

PR #512, opened 2015-09-09
Remove deb-src lines to support repositories without sources.

lrowe:
Closes travis-ci/travis-ci#4772.

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -93,7 +93,12 @@ module Travis
               sh.export 'DEBIAN_FRONTEND', 'noninteractive', echo: true
               whitelisted.each do |source|
                 sh.cmd "curl -sSL #{source['key_url'].untaint.inspect} | sudo -E apt-key add -", echo: true, assert: true, timing: true if source['key_url']
-                sh.cmd "sudo -E apt-add-repository -y #{source['sourceline'].untaint.inspect}", echo: true, assert: true, timing: true
+                if source['sourceline'].start_with? 'ppa:'
+                  sh.cmd "sudo -E apt-add-repository -y #{source['sourceline'].untaint.inspect}", echo: true, assert: true, timing: true
+                else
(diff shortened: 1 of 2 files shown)


meatballhat:
@lrowe thanks for the PR!  Can this be reworked so that the `deb-src` removal is optional?  Perhaps a config format like this, where `addons.apt.deb_src` defaults to `true`:

``` yaml
addons:
  apt:
    packages:
    - elasticsearch
    sources:
    - elasticsearch-1.7
    deb_src: false
```

lrowe:
Thinking this through again, removing all deb-src lines is probably a bit drastic (though it might be a good idea to remove them in the image itself for faster apt-get updates.)

I've updated the patch to simply append the sourceline to /etc/apt/sources.list when it is not a 'ppa:' link.

meatballhat:
@lrowe AFAIK if we're bypassing `apt-add-repository`, this means we're relying on the GPG keys already being present, no?

lrowe:
@meatballhat I don't think so, I think that's what the separate `apt-key add` is for a few lines above.

The elasticsearch apt repositories doc has steps for both `apt-key add` as well as the append to sources.list: https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-repositories.html#_apt

meatballhat:
@lrowe I'm silly and had the logic backward in my head.  The `ppa:` sources are special in that they do not contain a key, e.g.: https://github.com/travis-ci/apt-source-whitelist/blob/90716de807e7eca98d9f037ed02cf940ee7ef33e/ubuntu.json#L2-L6, which doesn't matter in this case since *only* `ppa:` sources would be added via `apt-add-repository` with your change.

BanzaiMan:
So, maybe this logic should live with `apt-source-whitelist`, rather than `travis-build`. That is, source should indicate where or not it adds src.

meatballhat:
:sheep: :it:

I don't think any further changes are needed to either this repo or the apt source repo.  Sorry for delay!

PR #465, opened 2015-06-08
Add Crystal

will:
In https://github.com/travis-ci/travis-ci/issues/4056 @BanzaiMan suggested this [community supported languages](http://docs.travis-ci.com/user/languages/community-supported-languages/) approach instead. 

This PR is an attempt at that.

/cc conscripted from IRC support group @asterite, @jhass, @waj

PR #514, opened 2015-09-20
Crystal: Install shards

jhass:
Starting with Crystal 0.8.0, Shards is the default package manager
for Crystal. However it is still a separate project and not distributed
alongside.

Since installing Shards on Travis will become very common for Crystal
projects, always do it.

ref #465 

/cc @asterite @waj @will @BanzaiMan

### DIFF
--- a/lib/travis/build/script/crystal.rb
+++ b/lib/travis/build/script/crystal.rb
@@ -5,6 +5,7 @@ module Travis
 
         def configure
           super
+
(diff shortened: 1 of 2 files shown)


will:
Looks good. Tried the new download step and it works, thanks for finding that.

asterite:
Nice! :+1:

jhass:
Anything else needed here?

PR #515, opened 2015-09-24
Update HHVM on Precise

BanzaiMan:
Since our support on Precise is ending, we do not have a chance to update HHVM to anything newer than 3.5 that we currently have.

HHVM has terminated [Precise support at 3.6.x](https://github.com/facebook/hhvm/wiki/Prebuilt-Packages-on-Ubuntu-12.04), so we handle this special case.

On future releases and OS X, we do not update HHVM, as we assume that it is reasonably up-to-date.

### DIFF
--- a/lib/travis/build/script/php.rb
+++ b/lib/travis/build/script/php.rb
@@ -74,12 +74,24 @@ module Travis
         end
 
         def configure_hhvm
-          install_hhvm_nightly if nightly?
+          if nightly?
+            install_hhvm_nightly
+          elsif hhvm?
+            update_hhvm
+          end
           fix_hhvm_php_ini
         end
 
+        def update_hhvm
+          sh.if '"$(lsb_release -sc &> /dev/null)" = "precise"' do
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/439173#L95-L96

meatballhat:
@BanzaiMan Does this also mean we (meaning myself and @solarce) can remove HHVM from the next precise image update?

BanzaiMan:
@meatballhat Is there a Precise update scheduled?

BanzaiMan:
@meatballhat Let's leave it there for now. If the update includes the latest 3.6.x HHVM, then this update step would be a no-op (though it adds a few seconds to the build). It is easy enough to revert it.

meatballhat:
@BanzaiMan There is a precise update in the works, so it's scheduled-ish, with a deadline of "before 2016".

PR #516, opened 2015-09-24
Show list of iOS simulators in announce section

henrikhodne:
Suggested in travis-ci/travis-ci#4749.


BanzaiMan:
Good idea. Do we want a spec for it?

henrikhodne:
I'm leaning towards not adding a spec, since the only way I could think of adding a spec for this would tie the spec pretty closely to the implementation.

PR #517, opened 2015-09-24
Introduce a new build stage "setup_cache"

BanzaiMan:
This dissociates setting up cache from the "install" step.
"install" is often overridden, which can confuse
users when "cache: bundler" (for example) does not work.

This resolves https://github.com/travis-ci/travis-ci/issues/3239

It is tested briefly: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/439189

### DIFF
--- a/lib/travis/build/script/c.rb
+++ b/lib/travis/build/script/c.rb
@@ -27,8 +27,7 @@ module Travis
           super << '--compiler-' << compiler
         end
 
-        def install
-          super
+        def setup_cache
           if data.cache?(:ccache)
             directory_cache.add('~/.ccache')
           end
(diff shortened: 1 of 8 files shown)


joshk:
This is very exciting! 

Is it possible to do this via feature flag for now? eg. support both ways to setup the cache.

BanzaiMan:
@joshk I looked at the making this into a feature, and I think it is a bit hard. Build stages are currently hard coded, which makes such an endeavor a little more invasive than I would like.

BanzaiMan:
Changing strategies to achieve what we want. Will close this, and open a new one.

PR #518, opened 2015-09-25
Install Perl on demand

BanzaiMan:
🐪

### DIFF
--- a/lib/travis/build/script/perl.rb
+++ b/lib/travis/build/script/perl.rb
@@ -13,6 +13,10 @@ module Travis
 
         def setup
           super
+          sh.if "! -x $HOME/perl5/perlbrew/perls/#{version}/bin/perl" do
+            sh.echo "#{version} is not installed; attempting download", ansi: :yellow
+            install_perl_archive(version)
+          end
           sh.cmd "perlbrew use #{version}", assert: false
         end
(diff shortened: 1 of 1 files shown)


meatballhat:
\*ports to cookbooks\*

meatballhat:
The namespace for the tarball appears to be `14.04`.  Could this be a wee bit more specific?  Perhaps `$(lsb_release -is | tr '[:upper:]' '[:lower:]')/$(lsb_release -rs)/$(arch)`, which would be `ubuntu/14.04/x86_64`?

BanzaiMan:
Sure thing! We need to standardize paths (hierarchy and names) across all languages.

meatballhat:
@BanzaiMan I'd love to get some standardization action going.  I'm assuming that anything that's dynamically linked will need to have distro and arch information, and namespacing by OS is cool.  This sounds related to the internal discussion of stacks, too, but then we're getting very Travis-specific.  Dunno.  Talking time.

BanzaiMan:
Rebased to use the standard archive URL scheme, based on #539.

BanzaiMan:
I've tested on staging this PR and the archiving process. It is ready to go, and I plan to deploy it next Monday.

PR #519, opened 2015-09-25
Include `apt-get update` in build output

dirk:
Redirecting output to a log file isn't useful for people trying to debug build failures.

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -126,7 +126,7 @@ module Travis
 
             unless whitelisted.empty?
               sh.export 'DEBIAN_FRONTEND', 'noninteractive', echo: true
-              sh.cmd "sudo -E apt-get -yq update &>> ~/apt-get-update.log", echo: true, timing: true
+              sh.cmd "sudo -E apt-get -yq update", echo: true, timing: true
               sh.cmd 'sudo -E apt-get -yq --no-install-suggests --no-install-recommends ' \
                 "--force-yes install #{whitelisted.join(' ')}", echo: true, timing: true
             end


jfirebaugh:
:+1: Yes please! I lost a lot of time to this over in https://github.com/travis-ci/apt-source-whitelist/issues/156.

BanzaiMan:
I am torn. I see the value in displaying, but it also tends to be long and could tip the log output over the 4 MB limit. It is easier to display what is hidden (with `cat ~/apt-get-update.log`) than hide what is shown.

dirk:
@BanzaiMan: Where/how are you getting giant 4 MB logs from `apt-get update`? My experience with the `update` command is that is has never shown more than 50-100 lines, even with a few extra PPAs and sources added to the default list.

jfirebaugh:
Ping @BanzaiMan -- could you please reconsider this PR? The current situation is a problem for multiple people, and apt output is not going to be a significant contributor to log size.

dirk:
Seconding @jfirebaugh in pinging @BanzaiMan.

I **highly** doubt this would tip anyone over the 4 MB limit, and if it did it would indicate they're trying to do some really wacky stuff with `apt-get` that they probably shouldn't be doing in the first place (especially for CI where you're trying to make the minimum possible configuration that users would have to have to use the given piece of software being CI'ed).

lib/travis/build/addons/apt.rb
```diff
@@ -126,7 +126,7 @@ def add_apt_packages
 
             unless whitelisted.empty?
               sh.export 'DEBIAN_FRONTEND', 'noninteractive', echo: true
-              sh.cmd "sudo -E apt-get -yq update &>> ~/apt-get-update.log", echo: true, timing: true
+              sh.cmd "sudo -E apt-get -yq update", echo: true, timing: true
```
BanzaiMan: Does `apt-get update` return non-zero if there is an error? If so, it might be better to dump the logs only when there is a problem.

jfirebaugh: I suspect not, unfortunately. That's another part of why https://github.com/travis-ci/apt-source-whitelist/issues/156 took me so long to figure out: `apt-get update` failed, and instead of failing the build immediately it continued on, to produce cryptic errors later.

deckar01: It does return a non-zero value if there is an error, but not if there are only warnings.

Failing to resolve the repository host only generates a warning (0).
 > W: Some index files failed to download. They have been ignored, or old ones used instead.

Providing an invalid distribution name to a reachable repository host results in an error (100).
 > E: Some index files failed to download. They have been ignored, or old ones used instead.

The both have the same final message, but one is a warning and the other is an error.

meatballhat:
This has been addressed via the `debug_apt` feature :metal:

PR #522, opened 2015-09-28
Crystal: Update shards install command

jhass:
Shards changed its release format, so the command to install it needs some updates.

/cc @asterite @waj @will @ysbaddaden @BanzaiMan


will:
:thumbsup:

PR #526, opened 2015-10-13
Crystal: Add support for nightly builds

will:
@jhass I noticed the recv-keys is the same on http://nightly.crystal-lang.org/, does this look correct?

### DIFF
--- a/lib/travis/build/script/crystal.rb
+++ b/lib/travis/build/script/crystal.rb
@@ -11,15 +11,12 @@ module Travis
 
             sh.cmd %q(sudo sh -c 'apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54')
 
-            if config[:crystal] == 'nightly'
-              sh.cmd %q(sudo sh -c 'echo "deb http://crystal-dist.aeshna.de/apt crystal main" > /etc/apt/sources.list.d/crystal-nightly.list')
-              sh.cmd %q(sudo sh -c 'apt-get update')
-              sh.cmd %q(sudo apt-get install crystal-nightly)
-            else
-              sh.cmd %q(sudo sh -c 'echo "deb http://dist.crystal-lang.org/apt crystal main" > /etc/apt/sources.list.d/crystal.list')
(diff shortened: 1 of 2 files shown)


jhass:
It should be, got a signing subkey for the main key.

jhass:
Should we perhaps explicitly validate for `default` (and maybe also allow `latest` or something)?

lib/travis/build/script/crystal.rb
```diff
@@ -10,10 +10,16 @@ def configure
             sh.echo 'Installing Crystal', ansi: :yellow
 
             sh.cmd %q(sudo sh -c 'apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54')
-            sh.cmd %q(sudo sh -c 'echo "deb http://dist.crystal-lang.org/apt crystal main" > /etc/apt/sources.list.d/crystal.list')
-            sh.cmd %q(sudo sh -c 'apt-get update')
 
-            sh.cmd %q(sudo apt-get install crystal)
+            if config[:crystal] == 'nightly'
+              sh.cmd %q(sudo sh -c 'echo "deb http://crystal-dist.aeshna.de/apt crystal main" > /etc/apt/sources.list.d/crystal-nightly.list')
```
jhass: This should actually use `nightly.crystal-lang.org`, gotta update the index, sorry.

will:
@jhass yeah I like `latest` better. 

Does the main apt let you install old versions with like `apt-get install crystal-0.7.3`?

jhass:
Good question, I don't think so.

spec/build/script/csharp_spec.rb
```diff
@@ -27,7 +27,7 @@
 
   describe 'version switching' do
     it 'throws a error with a invalid version' do
-      data[:config][:mono] = 'foo'
+      data[:config][:crystal] = 'foo'
```
jhass: Typo.

jhass:
LGTM (I hope :P) :+1:

PR #528, opened 2015-10-14
travis_buildtempl_header: Save full PIPESTATUS

Arthur2e5:
This commit saves the full `PIPESTATUS[]` array into a globally-accessible array called `TRAVIS_PIPESTATUS[]`, and is an extension of the edits involved in  infertux/bashcov#12.

Two doubts so far:
* Shall we consider chances of the `TRAVIS_PIPESTATUS` stuff being exported? If we consider those, we may need to save it like a something-delimited strings like a ling of `TRAVIS_PIPESTATUS=${TRAVIS_PIPESTATUS[*]}` right after the `result=` line, and users can split them later like `pstat=( $TRAVIS_PIPESTATUS )`.
* Is the bash version on travis new enough for negative indices to indicate the last elements? If not, we need to compute them using the length of the array ourselves.

### DIFF
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -40,7 +40,9 @@ function travis_cmd() {
   else
     eval "$cmd"
   fi
-  result=$?
+  TRAVIS_PIPESTATUS=( "${PIPESTATUS[@]}")
+  # in bash < 4, do like ${arr[$((${#arr[@]}-1))]} for the last element instead
+  result=${TRAVIS_PIPESTATUS[-1]}
 
   if [[ -n "$timing" ]]; then
     travis_time_finish
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -9,7 +9,7 @@ ANSI_CLEAR="\033[0K"
 TRAVIS_TEST_RESULT=
 TRAVIS_CMD=
 
-function travis_cmd() {
+travis_cmd() {
   local assert output display retry timing cmd result
 
   cmd=$1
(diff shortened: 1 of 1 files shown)


Arthur2e5:
I think I'll find out about doubt 2 after the CI check.

Arthur2e5:
Messing things up.. referenced the wrong issue in the first commit..

lib/travis/build/templates/header.sh
```diff
@@ -67,7 +67,7 @@ travis_time_finish() {
   return $result
 }
 
-function travis_nanoseconds() {
+travis_nanoseconds() {
   local cmd="date"
   local format="+%s%N"
   local os=$(uname)
```
Arthur2e5: TODO a pr to use `HOSTTYPE`..

lib/travis/build/templates/header.sh
```diff
@@ -168,14 +168,14 @@ travis_jigger() {
 travis_retry() {
   local result=0
   local count=1
-  while [ $count -le 3 ]; do
-    [ $result -ne 0 ] && {
+  while (( count < 3 )); do
+    (( result != 0 )) && {
       echo -e "\n${ANSI_RED}The command \"$@\" failed. Retrying, $count of 3.${ANSI_RESET}\n" >&2
```
Arthur2e5: TODO yet another PR to use `$(printf '%q ' "$@")`..

joshk:
Hi, thanks for the PR, could you please explain some more information on what the use case of this PR is?

Arthur2e5:
Forget it. travis-ci/travis-ci#3771 took me to some weird place and made me dizzy… Oh it might work still.

* * *

@joshk For pipe commands like `foo | bar | baz`, the `$?` would only consider the return value of `baz` which can be quite annoying.

Bash's `$PIPESTATUS[]` array will save all those three's return values, but like `$?` it will be modified immediately after another command ran. To ease users who want to do extra testing around it, we should save it somewhere.

BanzaiMan:
For complex builds that requires the use of `$?`, I tend to think writing a custom script is a cleaner solution.

Arthur2e5:
You are right.

PR #529, opened 2015-10-15
C#: only add the wheezy-libtiff-compat apt repository on Ubuntu 12.04

akoeplinger:
This allows the script to work on Ubuntu 14.04.


Joshua-Anderson:
It might be cleaner to check if `config['dist'] != 'trusty'` than running a check in the script itself.

akoeplinger:
That won't work with the docker service, see the conversation on Slack. This doesn't fully work yet, I forgot to mark as WIP :)

Joshua-Anderson:
Docker builds also run on trusty, so I believe dist should be set to  trusty when the job is processed.

akoeplinger:
It isn't, that's the problem.

Joshua-Anderson:
Just read slack, code LGTM! Sorry about that.

PR #525, opened 2015-10-09
[dart] handle tests written with the new test package

a14n:

PR #532, opened 2015-10-20
Improve dart support

svenfuchs:
* add specs to script/dart
* make sure we pass through a :raw option for :elif nodes
* make sure shell/generator/bash respects :raw for :elif nodes
* simplify the condition 

It seems there's no specs for shell/ast.rb ... we should probably add some at some point, since this would have caught the `:raw` option not being added in for `:elif` nodes. For now I've just added a respective spec to the dart script spec.

### DIFF
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -87,26 +87,26 @@ module Travis
 
         def script
           # tests with test package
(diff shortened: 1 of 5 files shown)


BanzaiMan:
This PR addresses https://github.com/travis-ci/travis-build/pull/525#issuecomment-149556699.

PR #534, opened 2015-10-23
Add travis support to Pharo (pharo.org) language

estebanlm:
We are a growing community and we started to use travis in a "manual" way... better like this, so we can provide managed containers (right now we are using legacy mode).

### DIFF
--- a/lib/travis/build/script.rb
+++ b/lib/travis/build/script.rb
@@ -37,6 +37,7 @@ require 'travis/build/script/ruby'
 require 'travis/build/script/rust'
 require 'travis/build/script/scala'
 require 'travis/build/script/smalltalk'
+require 'travis/build/script/pharo'
 require 'travis/build/script/shared/directory_cache'
 
 module Travis
--- a/lib/travis/build/script/pharo.rb
+++ b/lib/travis/build/script/pharo.rb
@@ -3,8 +3,8 @@ module Travis
     class Script
       class Pharo < Script
         DEFAULTS = {
-          version: 'stable',
-          sourceDir: '.'
+          pharo: 'stable',  # stable, alpha, 50, 40"
+          source_dir: '.'    # where the pharo sources actually are (usually root, ./mc or ./repository)
         }
         # baseline
         # test
@@ -23,6 +23,13 @@ module Travis
           end
         end
 
+        def announce
+          super
+          sh.echo ''
+          sh.cmd './pharo Pharo.image printVersion'
+          sh.echo ''
+        end
+
         def setup
           super
           sh.cmd "export PROJECT_HOME=\"$(pwd)\""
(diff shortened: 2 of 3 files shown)


timfel:
Hi @estebanlm, I would say that Pharo is still a Smalltalk distribution, and as such should go with the Smalltalk language as a version to select. That way, we can also include Gemstone, eventually

estebanlm:
H @timfel,
The problem is that Gemstone, Pharo, Squeak, etc. even if they are smalltalk distributions they have build process very different... In fact I tried to fork and update the one called "smalltalk", but it should be called just "Squeak": is too specific. 
We do not rely in all those scripts but command line options integrated in our language. Following Squeak path would create a dirty version and a maintenance problem for us. 
That's why I took this path... we cannot use the other :(

timfel:
I realize they are different in their use of commandline arguments. But not doing the work to abstract that in the language definition means that any project has to decide if it wants to support _only_ pharo or only _squeak_, which I think would just serve to further balkanize the smalltalk community.

fniephaus:
I do agree with @timfel.
Also, I started [filetreeCI](https://github.com/hpi-swa/filetreeCI), because I just wanted to see how much it takes to run Squeak on Travis' container-based infrastructure. builderCI does support Gemstone, Pharo and Squeak, but it would require lots of work to make it run on the new infrastructure.
It is nice to see that filetreeCI is now used to make Smalltalk support available on Travis-CI.
I am happy to accept PRs and add Pharo support to filetreeCI.
In fact, I was already looking into that as [I told Serge](https://twitter.com/fniephaus/status/657246079482777600).

estebanlm:
@fniephaus, but we do not need all that nor we can maintain it. 
Then we also have special requirements that of course, we could ask you to provide but as we continue to diverge, that will be harder and harder. 
We are even thinking on change the filetree source format.

@timfel,  please, notice also that even if all dialects has similar origines (Pharo being actually a fork of Squeak), they are not the same language. Projects made with Pharo, Squeak or Gemstone (or any other dialect) are not intended to be portable between them *by default*... is not a requirement for any of our languages... is just something some people desires. 

What will actually happen if we are asked to use filetreeCI is that we are going to continue using the legacy build/test process we use now (you can check an example here: https://github.com/estebanlm/ice) instead using filetreeCI, just because is a lot easier for us.

Sorry, but filetreeCI does not provides any advantage for us :(

fniephaus:
The way filetreeCI currently works is different to how you guys run Pharo builds on Travis.
So I understand your concern and I certainly don't want to force you into using filetreeCI.
I was suggesting to add the Pharo way of testing to filetreeCI, so that Travis supports both, Pharo and Squeak (and maybe even Gemstone), as part of their Smalltalk support.
Also, feel free to ignore the *filetree* in filetreeCI. It was/is just a working title. Maybe later we should call it something like *smalltalk-ci*.

Finally, I am not sure if we want to continue the discussion here.
Maybe we should use one of our mailing lists instead?

estebanlm:
well, I'm open to suggestions, then :)
yes, let's discuss in private so we do not annoy travis people anymore. 
I will close this pull request and will wait for any outcome ;)

(thanks @timfel for the feedback)

PR #539, opened 2015-10-29
Use standardized archive URL for on-demand language installations

BanzaiMan:
The scheme is: BUCKET/binaries/DIST/RELEASE/ARCH/LANGUAGE-VERSION.tar.bz2

Specs involving hard-coded download URLs are temporarily skipped

### DIFF
--- a/lib/travis/build/script.rb
+++ b/lib/travis/build/script.rb
@@ -2,6 +2,7 @@ require 'core_ext/hash/deep_merge'
 require 'core_ext/hash/deep_symbolize_keys'
 require 'core_ext/object/false'
 require 'erb'
+require 'rbconfig'
 
 require 'travis/build/addons'
 require 'travis/build/appliances'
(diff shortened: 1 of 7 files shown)


lib/travis/build/script.rb
```diff
@@ -78,6 +79,12 @@ def cache_slug
         'cache'
       end
 
+      def archive_url_for(bucket, version, ext = 'bz2')
+        lang = self.class.name.split('::').last.downcase
+
+        "https://s3.amazonaws.com/#{bucket}/binaries/#{host_os}/#{rel_version}/$(uname -m)/#{lang}-#{version}.tar.#{ext}"
```
meatballhat: Is `uname -m` the one that's reliably going to be `x86_64` for 64-bit, or is there a chance it'll be `amd64`?

BanzaiMan: On OS X and Ubuntu, this returns x86_64. It is entirely possible that it returns amd64 on other platforms that we may support in the future.

meatballhat: :+1: :grin:

PR #541, opened 2015-11-02
Support regular expressions in APT package whitelist

nickmarden:
This PR is an alternative solution to [the issue I opened](https://github.com/travis-ci/apt-package-whitelist/issues/1595) in the apt-package-whitelist repo, regarding my inability to install a specific version of the elasticsearch package.

After some further reflection I've decided that the problem with the elasticsearch case is a fairly repeatable problem that may occur with some semi-commercial or purely-commercial software packages, to wit:
- The package is released as binary-only (no source packages), so the standard apt-package-whitelist workflow can't be used
- The package is deemed sufficiently safe by a Travis CI maintainer that the repo and (non-version-specific) package are added to the respective whitelists
- The package maintainer periodically releases subsequent patch-level or minor-level releases which match the "generic" package name added by the Travis CI maintainer in the previous step

In this situation the Travis CI user is up the proverbial creek because they can't specify which exact package version they are installing - the apt-package-whitelist workflow is unavailable to them (no source packages) and the whitelisted package name ("elasticsearch") doesn't support version information.

My suggestion then is this PR: allow _a Travis CI maintainer_ to specify a package whitelist regex rather than a package whitelist name. This safely limits the subset of package strings that could be requested by Travis CI users, while still allowing them to pick a specific version if that matters to them.

For example, in the [existing ubuntu-precise whitelist](https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise) whitelist could be modified from this:

```
elasticsearch
elasticsearch:i386
```

to

```
^elasticsearch(-1\.[67]\.(\d+)(:i386)?)?$
```

which covers all of the packages currently whitelisted in the elasticsearch-1.6 and elasticsearch-1.7 repos, and also allows users to specify any of the following packages version strings:

```
elasticsearch
elasticsearch-1.6.2
elasticsearch-1.6.2:i386
elasticsearch-1.7.3
```

And, of course, if this gets merged, I'd love it if the regex shown above gets added to the apt-package-whitelist repo as well :-)

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -12,19 +12,35 @@ module Travis
 
         class << self
           def package_whitelist
-            @package_whitelist ||= load_package_whitelist
+            @package_whitelist ||= load_package_whitelist.reject { |w| w =~ regex_detecting_regex }
+          end
+
+          def package_whitelist_regexes
+            @package_whitelist_regexes ||= load_package_whitelist
+              .select { |w| w =~ regex_detecting_regex }
+              .map { |w| Regexp.compile(w) rescue nil }
+              .compact
           end
 
           def source_whitelist
             @source_whitelist ||= load_source_whitelist
           end
 
+          def reset_caches
+            @package_whitelist =
+              @package_whitelist_regexes =
+              @source_whitelist =
+              @loaded_package_whitelist = nil
+          end
+
           private
 
           def load_package_whitelist
             require 'faraday'
-            response = fetch_package_whitelist
-            response.split.map(&:strip).sort.uniq
+            @loaded_package_whitelist ||= begin
+              response = fetch_package_whitelist
+              response.split.map(&:strip).sort.uniq
+            end
           rescue => e
             warn e
             []
@@ -54,6 +70,11 @@ module Travis
 
           def source_whitelist_url
             ENV['TRAVIS_BUILD_APT_SOURCE_WHITELIST']
+          end
+
+          def regex_detecting_regex
+            # Detects only the subset of regexes supported in the apt packages whitelist
+            /[*?+()]/
           end
         end
 
@@ -110,7 +131,8 @@ module Travis
             disallowed = []
 
             config_packages.each do |package|
-              if package_whitelist.include?(package)
+              if package_whitelist.include?(package) ||
+                 package_whitelist_regexes.detect { |w| w.match(package) }
                 whitelisted << package
               else
                 disallowed << package
@@ -146,6 +168,10 @@ module Travis
 
           def package_whitelist
             ::Travis::Build::Addons::Apt.package_whitelist
+          end
+
+          def package_whitelist_regexes
+            ::Travis::Build::Addons::Apt.package_whitelist_regexes
           end
 
           def source_whitelist
--- a/spec/build/addons/apt_spec.rb
+++ b/spec/build/addons/apt_spec.rb
@@ -8,7 +8,7 @@ describe Travis::Build::Addons::Apt, :sexp do
   let(:addon)             { described_class.new(script, sh, Travis::Build::Data.new(data), config) }
   let(:config)            { {} }
   let(:source_whitelist)  { [{ alias: 'testing', sourceline: 'deb http://example.com/deb repo main' }] }
-  let(:package_whitelist) { %w(git curl) }
+  let(:package_whitelist) { %w(git curl ^elasticsearch(-1\.[67]\.(\d+)(:i386)?)?$) }
   subject                 { sh.to_sexp }
 
   before :all do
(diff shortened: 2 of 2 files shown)


nickmarden:
@BanzaiMan @meatballhat You guys seem to be two people who would care a lot about this, so I'd definitely welcome your feedback or alternative workaround to this issue (if such a workaround exists).

nickmarden:
Hi @BanzaiMan have I opened this PR the right way? Please let me know if there is a way forward on this issue.

meatballhat:
@nickmarden sorry for delay!  You opened it the right way.  We were all at an offsite when you opened it up, and are in various stages of getting back online this week.

meatballhat:
LGTM :ship: :it: pending @BanzaiMan :eyes: :+1:

nickmarden:
Thanks @meatballhat for getting back to me. I'll await feedback from @BanzaiMan and all-powerful :eyes:

BanzaiMan:
I understand the desire for the use case, but I am not sure if we support adopt it. This would allow potentially untested packages to be installed.

meatballhat:
@BanzaiMan imho we already run that risk with exact string matches given we allow custom sources to be added that we don't fully control.  I view this as an incremental change.  Certainly, we will have to be diligent about just how loose a given regex is.  Diligence everywhere.

nickmarden:
@BanzaiMan I completely agree with the concern but the current infrastructure allows for installing untested packages anyway. When `elasticsearch` was added to the apt package whitelist originally, the maintainers must have decided that future package versions within the same whitelisted repo were likely to be container-safe. That conclusion, combined with knowing the provider's version numbering scheme, is the due diligence that @meatballhat is referring to.
My goal here is not to be forced to use the latest patch level package until I am ready for it.

nickmarden:
@BanzaiMan is there anything I can do to further allay your concerns about the risk of the PR? I'm pretty sure the points mentioned above by @meatballhat and me explain why this isn't actually a security risk IRL. And I could *really* use this functionality, for all of the reasons I've given so far.

nickmarden:
@BanzaiMan sorry to be a pest, but my work is blocked on this PR and I don't know what to tell the team I'm working with.

This PR is so limited in scope - it's intended for use only by travis-build maintainers and even then only to allow the installation of specific packages _that are already implicitly whitelisted_ - that I can't see how this is a big security risk.

BanzaiMan:
Please request separate packages be whitelisted for now, while we come to a conclusion here.

nickmarden:
@BanzaiMan I have no mechanism for making that request because the [current whitelist approval process](https://github.com/travis-ci/apt-package-whitelist#package-approval-process) requires access to the Debian source. There is no source provided for Elasticsearch, so it is impossible for me to submit the request through normal channels.

Please see [my original attempt](https://github.com/travis-ci/apt-package-whitelist/issues/1564) and the issues that [referenced here](https://github.com/travis-ci/apt-package-whitelist/issues/1595) for more background.

nickmarden:
@meatballhat @BanzaiMan I'm sorry to be a pest but it appears that this PR is currently dead in the water because of a misunderstanding that I've tried very hard to clear up.

Can you please let me know if you intend to consider this PR?

nickmarden:
@BanzaiMan @meatballhat @joshk I feel that I've provided a very reasonable pull request, as well as lots of background information about why it's necessary. What I'm getting back is either silence, or responses that suggest that no one has read what I've written. It's pretty frustrating.

I'm OK with this PR being rejected for valid technical reasons, but as yet no discussion about those reasons has taken place. As I've asked before, would you please let me know if you intend to consider this PR?

nickmarden:
@BanzaiMan One last attempt at this before I give up. You asked me on Nov 24 to "[p]lease request separate packages be whitelisted for now," but as I've demonstrated that's physically impossible for me to do. It requires access to the elasticsearch source code, which is not public.

Can you please consider my PR or suggest a workable alternative? I'm having trouble understanding why this conversation has stopped completely - I've attempted to be extremely reasonable, do all of the heavy lifting, and explain myself very clearly.

BanzaiMan:
Sorry for the delay.

ElasticSearch sources are defined in https://github.com/travis-ci/apt-source-whitelist/blob/1f3c504a529a5c584dd8a581737b93caa476f5cc/ubuntu.json#L82-L101. None of them has what you are after?

If you need the additional APT package source, I must suggest that you use `sudo: required`, as the source whitelist requests are not moving quickly.

nickmarden:
@BanzaiMan we're specifically trying to get away from `sudo: required` so that we can use the faster containerized builds. That's the whole reason that I did this work in the first place.

The sources that you pointed to above aren't the source *code* for Elasticsearch, they are the APT repo sources from which an already-built .deb can be fetched. In order to whitelist a package to a specific version number using apt-package-whitelist [one must provide the source *code*](https://github.com/travis-ci/apt-package-whitelist#inspecting-packages), and in this case I cannot because ES doesn't offer it.

Or am I misunderstanding something?

meatballhat:
Now obsolete!  Please holler if you believe this has been closed in error.

PR #543, opened 2015-11-11
Always pass `-y` to `apt-get install` in R builds.

craigcitro:
It seems that this is necessary in trusty, but wasn't in precise. Fixes
travis-ci/travis-ci#5086.

PTAL @BanzaiMan @maverickg

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -75,7 +75,7 @@ module Travis
             # Install an R development environment. qpdf is also needed for
             # --as-cran checks:
             #   https://stat.ethz.ch/pipermail/r-help//2012-September/335676.html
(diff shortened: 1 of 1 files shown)


maverickg:
Thanks. Looks good to me.

meatballhat:
An earlier version of the trusty image(s) needed this, but it's since been addressed (I thought).  Is this still a problem in production?

maverickg:
here is an example https://travis-ci.org/stan-dev/rstan/jobs/90447999

meatballhat:
@maverickg Thank you!  Working on a fix.  Either way, I think this patch is :+1:.  No harm in passing `-y`.

meatballhat:
Awaiting @BanzaiMan and/or @sinthetix sign-off

craigcitro:
thanks @meatballhat and @maverickg!

PR #552, opened 2015-11-19
Also install libmariadbclient for the mariadb plugin

sodabrew:
Installing the MariaDB client library and dev package would allow me to avoid a sudo invocation here:
https://github.com/brianmario/mysql2/blob/master/.travis_setup.sh#L12

/cc @grooverdan


BanzaiMan:
Please fix the specs, and we'll merge.

sodabrew:
Fixed, thank you!

sodabrew:
:trophy:

BanzaiMan:
Where is this package supposed to come from? https://travis-ci.org/BanzaiMan/travis_production_test/builds/92713047#L94

sodabrew:
Oh, it looks like there isn't a non-version-suffixed libmariadbclient package here: http://sfo1.mirrors.digitalocean.com/mariadb/repo/10.1/ubuntu/pool/main/m/mariadb-10.1/
http://sfo1.mirrors.digitalocean.com/mariadb/repo/10.0/ubuntu/pool/main/m/mariadb-10.0/

So maybe specifying only `apt-get install libmariadbclient-dev` and allowing it to pull in an appropriate "libmariadbclientXX" library would be better?

PR #554, opened 2015-11-23
Let versioned libmariadbclient package be determined automatically.

peplin:
There is no "libmariadbclient" package, only one with a version in the
name ("libmariadbclient18"). Installing libmariadbclient-dev will
automatically pull in the proper version.

This fixes the change from #552.


daspecster:
:+1:

BanzaiMan:
Thanks for the PR. I see that this is an identical PR as #555, which I'd just merged.

PR #557, opened 2015-11-29
Adjust Julia release and nightly URL's to download from S3

tkelman:
This removes the dependency on status.julialang.org, which is another point of failure for DNS outages etc.
Downside is we'll need a tiny PR to change the `0.4` in `release` every time we have a new minor release, which should only be once or twice a year.

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -85,19 +85,20 @@ module Travis
           def julia_url
             case config[:os]
             when 'linux'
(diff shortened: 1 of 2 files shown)


tkelman:
cc @staticfloat and @ninjin, this is a bit urgent since status.julialang.org is down right now, we're in the middle of some messy DNS migrations and builds are failing.

staticfloat:
+1

ninjin:
Looks good to me.

PR #558, opened 2015-12-01
remove .ruby-version beta warning

whit537:
The [docs](https://docs.travis-ci.com/user/languages/ruby/#Using-.ruby-version) give no indication that version detection via `.ruby-version` is a beta feature. Seems like either the docs should be updated, or the beta warning should be removed from build output. This PR does the latter.

### DIFF
--- a/lib/travis/build/script/shared/rvm.rb
+++ b/lib/travis/build/script/shared/rvm.rb
@@ -5,8 +5,7 @@ module Travis
         include Chruby
 
         MSGS = {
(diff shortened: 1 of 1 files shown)


BanzaiMan:
We should do the former, IMO.

whit537:
@BanzaiMan Okay! https://github.com/travis-ci/docs-travis-ci-com/pull/437 :-)

joshk:
@BanzaiMan do you think this should still be a beta feature?

PR #559, opened 2015-12-02
Optionally separate cache setup process from install

BanzaiMan:
With `cache.custom_install: true`, other caching directives
such as `cache.bundler: true` will be respected, and the
overridden `install` section will not negate `cache.bundler: true`

Caveat: There are no specs yet.


BanzaiMan:
The fold now looks like this:

build__792_-_banzaiman_travis_staging_test_-_travis_ci

PR #569, opened 2015-12-15
Remove libzmq3-dev custom PPA source

BanzaiMan:
Remove custom PPA source that we used for adding libzmq3-dev.
We are not updating the source any more, and it only prevents
other attempts to install zmq-related packages.

Fixes https://github.com/travis-ci/travis-ci/issues/3106


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/465672#L140

PR #571, opened 2015-12-17
HHVM on Trusty

BanzaiMan:
This PR adds support for HHVM on Trusty, which is at 3.11 as of this writing.

A single build test is here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/465695

And a matrix build, along with other PHP versions, is here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/465699


BanzaiMan:
This has been reverted.

PR #574, opened 2015-12-19
Export $SMALLTALK and $BASELINE before custom env variables

fniephaus:
@BanzaiMan Unfortunately, the change from yesterday broke all Smalltalk builds, because the value for `baseline` always overrides the custom env variable which everyone used before.

This change should make yesterday's change backwards-compatible by exporting the `$SMALLTALK` and `$BASELINE` before any other env variable.

Sorry about the hassle, but can we also get this in soon? That'd be awesome!

/cc @marceltaeumel

### DIFF
--- a/lib/travis/build/script/smalltalk.rb
+++ b/lib/travis/build/script/smalltalk.rb
@@ -20,9 +20,9 @@ module Travis
         end
 
         def export
-          super
           sh.export 'SMALLTALK', config[:smalltalk], echo: false
           sh.export 'BASELINE', config[:baseline], echo: false
+          super
          end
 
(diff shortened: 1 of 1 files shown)


fniephaus:
We were able to implement a workaround on our side, so this isn't time critical anymore.

Happy holidays! :christmas_tree:

joshk:
Did you not want this merged at all?

fniephaus:
I wanted to wait until we have discussed https://github.com/hpi-swa/smalltalkCI/issues/20 (you might want to take a look as well).
Our workaround does the trick for now...

PR #576, opened 2015-12-29
Refactor node.js installation logic

BanzaiMan:
Also addresses https://github.com/travis-ci/travis-ci/issues/5333


BanzaiMan:
When nodejs.org is unreachable, we would get [an error like this](https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/466032):

```
$ nvm install 5.3
Version '5.3' not found - try `nvm ls-remote` to browse available versions.
Remote repository may not be reachable
$ nvm use 5.3
N/A: version "v5.3" is not yet installed
The command "nvm use 5.3" failed and exited with 1 during .
```

PR #580, opened 2016-01-05
Limit hostname to a reasonable length

BanzaiMan:
OpenJDK 6 and 7 runs into a problem if the host name is 64 characters or longer. This manifests as https://github.com/travis-ci/travis-ci/issues/5227.

This PR aims to reduce such friction by shortening the host name when `sudo` is available. (Within a Docker container, we cannot change the host name.)

This would print log messages like [this](https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/465962#L106-L108):

```
Truncating hostname. See https://github.com/travis-ci/travis-ci/issues/5227.
$ hostname
testing-gce-job-465963
```

I am a little concerned that this might have unintended consequences for builds that do not need this fix.


lib/travis/build/appliances/limit_hostname_length.rb
```diff
@@ -0,0 +1,22 @@
...
+module Travis
+  module Build
+    module Appliances
+      class LimitHostnameLength < Base
+        def_delegators :data, :job
+
+        def apply
+          sh.echo "Truncating hostname. See https://github.com/travis-ci/travis-ci/issues/5227.", ansi: :yellow
+          sh.raw 'sudo hostname "$(hostname 2>/dev/null | cut -d. -f1 | cut -d- -f1-2)-job-' + job[:id].to_s + '" 2>/dev/null', echo: false
+          sh.raw 'sed -e "s/^\\(127\\.0\\.0\\.1.*\\)/\\1 $(hostname | cut -d. -f1 | cut -d- -f1-2)-job-' + job[:id].to_s + '/" /etc/hosts | sudo tee /etc/hosts >&/dev/null', echo: false
```
meatballhat: I'm not sure there's a potential bug here, but can the new hostname be defined once and used in both the `hostname <value>` and `/etc/hosts` munging operations?

meatballhat:
I'm generally against having this applied to all scripts.  I'd much rather see it implemented as an addon so that users can opt in, e.g.:

``` yaml
addons:
  hostname: whatever-hostname-i-want-to-use-over-8000
```

``` yaml
addons:
  hostname: java6rulz
```

BanzaiMan:
@meatballhat That's a valid concern. But at that point, I would think that it is as good as (or as bad as) putting the `sudo hostname` and `echo | sudo tee`.

meatballhat:
@BanzaiMan I don't follow.   My proposal is to encapsulate the same commands in an addon that's opt-in rather than applying this change to all builds.

BanzaiMan:
@meatballhat The addon is logically encapsulating, but the amount of work that user needs to put forth is roughly equivalent. I am not fond of either solution, to be honest. Maybe there is another that I'm not seeing.

joshk:
I'm not sure how I feel about having the user opt in to change the hostname. I feel like things should 'just work' and not have to require the user to fix things.

I also understand that this isn't our 'fault', but well, thats the joy of providing a build service, right?

I'm not sure what the best solution is.

I guess I would agree that doing the `addon` could be a good first step.

dan-blanchard:
FYI, this doesn't just impact Java.  I've also experienced this issue with Grid Engine.  I'm sure there's plenty of old software that doesn't like long hostnames.

BanzaiMan:
What is Grid Engine?

dan-blanchard:
It's a family of old systems for doing distributed batch computation:
-  [Sun Grid Engine](https://en.wikipedia.org/wiki/Oracle_Grid_Engine)
-  [Son of Grid Engine](https://arc.liv.ac.uk/trac/SGE)
-  [Univa Grid Engine](http://www.univa.com/products/)

BanzaiMan:
Closed in favor of #620.

PR #582, opened 2016-01-05
Install Crystal via homebrew in OS X

porras:
I couldn't test it _for real_ because I don't have a working Travis environment but the generated script is correct and according to [this](https://docs.travis-ci.com/user/osx-ci-environment/) calling `brew` is possible.

Note that from some versions ago installing Crystal via `brew` includes `shards` so it's not needed (I think that might also be the case with `apt-get` but I preferred not to change that because I'm not sure).

/cc @asterite @jhass @waj @will

### DIFF
--- a/lib/travis/build/script/crystal.rb
+++ b/lib/travis/build/script/crystal.rb
@@ -9,22 +9,29 @@ module Travis
           sh.fold 'crystal_install' do
             sh.echo 'Installing Crystal', ansi: :yellow
 
-            sh.cmd %q(sudo sh -c 'apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54')
-
             version = select_version
             return unless version
 
-            sh.cmd %Q(sudo sh -c 'echo "deb #{version[:url]} crystal main" > /etc/apt/sources.list.d/crystal-nightly.list')
-            sh.cmd %q(sudo sh -c 'apt-get update')
-            sh.cmd %Q(sudo apt-get install #{version[:package]})
+            case config[:os]
+            when 'linux'
+              sh.cmd %q(sudo sh -c 'apt-key adv --keyserver keys.gnupg.net --recv-keys 09617FD37CC06B54')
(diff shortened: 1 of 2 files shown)


jhass:
I think we should install the latest shards not the shipped one in either case for now, that's why they're separate projects still.

porras:
@jhass mmmm then this PR is not complete still, I'll update it with that. Do you think it could be installed with `brew`, or should it use the same github release download?

and thanks for the very fast answer!

jhass:
I don't know

porras:
maybe @ysbaddaden has an advice on this?

Current situation:

* brew: 0.5.3
* shipped with crystal 0.10.0 via brew (latest crystal release): 0.5.4
* shipped with crystal HEAD via brew: 0.5.4
* github release 0.5.4

With the current situation and the PR as is, we would be getting the lastest shards, but of course that doesn't mean that's guaranteed to be always like that.

ysbaddaden:
We don't have to install Shards anymore, it's already bundled. In my projects that run on Travis I'm just using "crystal deps install". If a project needs a more recent version, it can install it manually via its travis.yml

I vote for skipping the (re)installation of Shards.

jhass:
I don't bundle it with the nightly packages yet iirc

porras:
Thanks Julien,

It's definitely present in the brew packages (both stable and head), so at
least for osx it can be safely skipped (i.e. as is in the PR).

jhass:
My biggest point of always fetching the latest version is a more consistent build environment across all variants.

ysbaddaden:
Shards is also bundled in the Debian packages. I dont know about the Linux nightly.

kepler0:
bump

BanzaiMan:
Hello. Is there a consensus on this PR?

jhass:
I didn't see a convincing counter argument here yet, so my vote stays at keeping the build environments more consistent and always installing the latest shards.

ysbaddaden:
OSX binaries of Shards are lagging behind. Shards is bundled with all Crystal releases. I usually release a new version shortly before a Crystal release (so the latest version is bundled). I can't see any reason to keep installing Shards anymore —unless we're using nightly releases.

PR #585, opened 2016-01-07
Ensure github.com SSH host key exists in ~/.ssh/known_hosts

BanzaiMan:
This host key and the IP address is current as of
Thu Jan  7 16:51:28 UTC 2016

### DIFF
--- a/lib/travis/build/appliances/add_github_host_key.rb
+++ b/lib/travis/build/appliances/add_github_host_key.rb
@@ -1,0 +1,13 @@
+require 'travis/build/appliances/base'
+
+module Travis
+  module Build
+    module Appliances
+      class AddGithubHostKey < Base
+        def apply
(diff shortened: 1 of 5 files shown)


lib/travis/build/appliances/add_github_host_key.rb
```diff
@@ -0,0 +1,13 @@
+require 'travis/build/appliances/base'
+
+module Travis
+  module Build
+    module Appliances
+      class AddGithubHostKey < Base
+        def apply
+          sh.raw %(grep '^github\\.com,' ~/.ssh/known_hosts >&/dev/null || echo 'github.com,204.232.175.90 ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ==' >> ~/.ssh/known_hosts)
```
meatballhat: Can you use `ssh-keyscan` instead of a hardcoded key?

meatballhat:
@BanzaiMan Is the hardcoded key necessary?  I would prefer using `ssh-keyscan` if possible.

meatballhat:
@BanzaiMan said another way, why not use the same strategy as https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/addons/ssh_known_hosts.rb ?

BanzaiMan:
@meatballhat I really prefer fixing the image.

BanzaiMan:
This approach does not work, unless we enumerate all the IP addresses that `github.com` resolves to and add them all.

BanzaiMan:
keyscan could work, I believe. I'll open a separate PR for that.

meatballhat:
@BanzaiMan I don't understand what you mean by "I really prefer fixing the image".  Is this a problem that's being tracked elsewhere?

BanzaiMan:
@meatballhat My initial guess (which may still be true) was that, since OS X and Linux behave differently for the same configuration, there must be something that the former missed. I don't know if it's true or not.

PR #587, opened 2016-01-08
Install PostgreSQL on demand

BanzaiMan:
See https://github.com/travis-ci/travis-ci/issues/4264.

With 9.3 (already installed, so nothing happens): https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/469792

With 9.5: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/469801#L78

Notice that 9.5 is configured on port 5433, which might be a problem.

Note that we do not install PostGIS.

### DIFF
--- a/lib/travis/build/addons/postgresql.rb
+++ b/lib/travis/build/addons/postgresql.rb
@@ -11,6 +11,7 @@ module Travis
         DEFAULT_FALLBACK_PORT = 5433
 
         def after_prepare
+          install(version)
           sh.fold 'postgresql' do
             sh.export 'PATH', "/usr/lib/postgresql/#{version}/bin:$PATH", echo: false
             sh.echo "Starting PostgreSQL v#{version}", ansi: :yellow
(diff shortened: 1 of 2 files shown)


joshk:
should we install postgis?
9.5 uses a different port?

BanzaiMan:
When it starts, though, it seems to bind to port 5432. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/469801#L136

So this is probably fine.

BanzaiMan:
PostGIS is not available from the usual PPA. https://launchpad.net/~ubuntugis/+archive/ubuntu/ppa

BanzaiMan:
https://github.com/travis-ci/travis-ci/issues/5445 might also be relevant (re: PostGIS).

gildegoma:
I jump into the discussion to put my "two" cents:

* By default the pgdg apt packages sequentially assign a specific tcp port for each postgresql installations (e.g. 5432 for the first installations, then 5433, etc.)
* As decided with @joshk, the travis `postgresql` chef cookbook has been "designed" to make all preinstalled versions using the default port (5432), and ensure that only one postgresql server can be started at the same time. We also override some postgresql settings for CI context. See https://github.com/travis-ci/travis-cookbooks/pull/221 for (many) more details.
* It is still unsure that PostGIS 2.1 will be officially supported for 9.5. As mentioned in https://github.com/travis-ci/travis-cookbooks/pull/651, upgrading to PostGIS 2.2 might be a good choice when introducing the preinstalled PostgreSQL 9.5

With all the above elements, I am not totally sure whether it is a good idea to "officially" help PostgreSQL installations on demand via `travis-build` as it could lead to new sources of confusion. "Warned" users can run the `apt-get install` commands they want, but then they are aware that the behaviour is a bit "out of travis support control".

gildegoma:
@BanzaiMan I think that your 9.5 instance only listened to 5433. It works because you connect via this port actually: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/469801#L138-L139

BanzaiMan:
Going to close this for now. There are some issues raised by @gildegoma that needs addressed.

PR #588, opened 2016-01-08
#Stop overriding travis-rubies URL

sikachu:
This patch makes RVM on Travis CI to successfully install binary rubies from travis-rubies server given only the partial version string. (For example, "2.3" now correctly use "2.3.0", "2.2" uses "2.2.4", etc.)

This override were causing RVM to unable to see which binary rubies are available from travis-rubies, as the old travis-ruby URL does not contain the necessary index file[[1]]. RVM already properly configured the
new travis-ruby URL[[2]] which host the necessary index file[[3]] on it, so this override is no longer needed.

[1]: https://s3.amazonaws.com/travis-rubies/binaries/index.txt
[2]: https://github.com/rvm/rvm/blob/1.26.11/config/db#L87
[3]: https://rubies.travis-ci.org/index.txt

### DIFF
--- a/lib/travis/build/script/shared/rvm.rb
+++ b/lib/travis/build/script/shared/rvm.rb
@@ -9,12 +9,6 @@ module Travis
           ruby_version_file: 'BETA: Using Ruby version from .ruby-version. This is a beta feature and may be removed in the future.'
         }
 
-        CONFIG = %w(
-          rvm_remote_server_url3=https://s3.amazonaws.com/travis-rubies/binaries
-          rvm_remote_server_type3=rubies
-          rvm_remote_server_verify_downloads3=1
-        )
-
         def export
           super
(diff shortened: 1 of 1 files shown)


joshk:
nice!
@rkh can I have your thoughts on this?
@sikachu could this change have any adverse effects?

sikachu:
@joshk only if RVM decides to remove Travis rubies from it source in the future. :smile: 

I doubt that will happen, but I could also update this patch to just point to https://rubies.travis-ci.org as a precaution. What do you think?

rkh:
Looks good. [Index file link](https://s3.amazonaws.com/travis-rubies/binaries/index.txt) does not resolve properly, but if it works for rvm, it works for me.

sikachu:
@rkh RVM has https://rubies.travis-ci.org as the `url3` setting, which would then try to look for the index at https://rubies.travis-ci.org/index.txt, which resolves correctly.

PR #590, opened 2016-01-11
Add GemStone support and refactor smalltalk.rb

fniephaus:
We are trying to get [GemStone running on Travis' container-based infrastructure](https://github.com/hpi-swa/smalltalkCI/issues/28).
Unfortunately, GemStone requires some complex OS prereqs.
This PR is based on the official [prereqs script](https://github.com/GsDevKit/GsDevKit_home/blob/dev/bin/utils/installOsPrereqs).

We have used the chance to remove `baseline:` again, since we are working on a more Travis-independent solution to configure Smalltalk builds. Also, we updated the announcement and the tests.

We do understand if reviewing this PR takes more time, but thanks for the awesome work!

Please let us know if there's any way we can test this (e.g. locally or in staging) before it's getting merged.

Many thanks,
Fabio

/cc @dalehenrich, @bahnfahren, @chistopher, @jchromik, and @Nef10

### DIFF
--- a/lib/travis/build/script/smalltalk.rb
+++ b/lib/travis/build/script/smalltalk.rb
@@ -6,7 +6,8 @@ module Travis
     class Script
       class Smalltalk < Script
         DEFAULTS = {
-          gemstone_hostname: 'travis.dev'
+          gemstone_hostname: 'travis.dev',
+          release_version: '#{rel_version}'
         }
         HOSTS_FILE = '/etc/hosts'
         TEMP_HOSTS_FILE = '/tmp/hosts'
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/smalltalk.rb
```diff
@@ -1,39 +1,69 @@
+# Copyright (c) 2015-2016 Software Architecture Group (Hasso Plattner Institute)
+# Copyright (c) 2015-2016 Fabio Niephaus, Google Inc.
+
 module Travis
   module Build
     class Script
       class Smalltalk < Script
-        DEFAULTS = {}
+        DEFAULTS = {
+          gemstone_hostname: 'travis.dev',
```
BanzaiMan: Is this a reasonable default?

fniephaus: @dalehenrich works a lot with GemStone and he has always used that hostname.
AFAIK, it only needs to be a predictable hostname, because Travis builds usually have a randomized one.

BanzaiMan: OK!

fniephaus:
Let me know when this is in staging, so I can play around with it? Thanks!

BanzaiMan:
@fniephaus It's deploy to staging and production.

PR #593, opened 2016-01-13
Add Jim Hester to the list of people to contact with r-travis issues

jimhester:
We are transitioning r-travis maintenance from @craigcitro to me, so I should be added to the list of people to contact with build issues.


BanzaiMan:
Super! And welcome!

PR #597, opened 2016-01-15
PhantomJS addon

drogus:
Recently we get more reports of problems with PhantomJS installations and there're multiple ways to install PhantomJS on different architectures. Eventually we plan to support some PhantomJS versions, but it's unclear when will it happen and even then we will most likely require users to change their configs to be able to benefit from that change.

This PR adds a PhantomJS addon, which deals with all of the platform differences in a transparent way. It would work similarly to Firefox addon, for example:

```
addons:
  phantomjs: 2.0.0
```

With such an addon we could avoid situations like described in ember-cli/ember-cli#3895


drogus:
@BanzaiMan I would like to push this thing forward, could you let me know what you think? When I have a green light I can test all of the scenarios and fix any issues (so far I just tested the commands usded in the addon during regular builds).

BanzaiMan:
I am a bit hesitant to move forward with this.

For one thing, we are not completely committed to building PhantomJS binary for every release. I built 2.0.0 because there was no official static binary was available.

For another, only supporting 1.9.8 and 2.0.0 seems arbitrary. (Although, we might do well to fall back to the official archive, since 2.1.x has Linux static binary archives.)

lib/travis/build/addons/phantomjs.rb
```diff
@@ -0,0 +1,72 @@
...
+                  sh.cmd "tar -xvf /tmp/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2 -C /tmp/travis-phantomjs", echo: true, timing: true, retry: false
+                  sh.export "PATH", "/tmp/travis-phantomjs:$PATH"
+                else
+                  # no need to do anything for 1.9.8 at the moment, it's the
+                  # default on precise
+                end
+              end
+
+              sh.else do
+                if version == '1.9.8'
```
BanzaiMan: Since we have 1.9.8 pre-installed currently, there is no need to do anything on this logical branch.

lib/travis/build/addons/phantomjs.rb
```diff
@@ -0,0 +1,72 @@
...
+          end
+        end
+
+        private
+          def raw_version
+            config.to_s.strip.shellescape
+          end
+
+          def sanitize(input)
+            if m = /\A(1.9.8|2.0.0)\z/.match(input.chomp)
```
BanzaiMan: We can probably accept arbitrary `/\A(\d+\.\d+\.d+)\z/`.

lib/travis/build/addons/phantomjs.rb
```diff
@@ -0,0 +1,72 @@
...
+
+            unless version
+              sh.echo "Invalid version '#{raw_version}' given. Only 1.9.8 and 2.0.0 are supported at the moment", ansi: :red
+              return
+            end
+
+            sh.echo "Installing PhantomJS #{version}", ansi: :yellow
+
+            sh.if "$(uname) = 'Linux'" do
+              sh.if "$(lsb_release -c) = 'precise'" do
```
BanzaiMan: The 2.0.0 binary exists for Trusty. https://s3.amazonaws.com/travis-phantomjs/phantomjs-2.0.0-ubuntu-14.04.tar.bz2

lib/travis/build/addons/phantomjs.rb
```diff
@@ -0,0 +1,72 @@
...
+                if version == '2.0.0'
+                  sh.mkdir "/tmp/travis-phantomjs"
+                  sh.cmd "wget https://s3.amazonaws.com/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2 -O /tmp/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2", echo: true, timing: true, retry: true
+                  sh.cmd "tar -xvf /tmp/travis-phantomjs/phantomjs-2.0.0-ubuntu-12.04.tar.bz2 -C /tmp/travis-phantomjs", echo: true, timing: true, retry: false
+                  sh.export "PATH", "/tmp/travis-phantomjs:$PATH"
+                else
+                  # no need to do anything for 1.9.8 at the moment, it's the
```
BanzaiMan: We should support 2.1.1.

lib/travis/build/addons/phantomjs.rb
```diff
@@ -0,0 +1,72 @@
...
+        def after_prepare
+          sh.fold 'install_phantomjs' do
+            sanitize(raw_version)
+
+            unless version
+              sh.echo "Invalid version '#{raw_version}' given. Only 1.9.8 and 2.0.0 are supported at the moment", ansi: :red
+              return
+            end
+
+            sh.echo "Installing PhantomJS #{version}", ansi: :yellow
```
BanzaiMan: We should check the current version with `phantomjs --version` (which takes about a whole second) and bail out of this addon if it matches the desired one.

drogus: Good point.

drogus:
> For one thing, we are not completely committed to building PhantomJS binary for every release. I built 2.0.0 because there was no official static binary was available.

Sure, that's why I wanted a green light - if we decide to roll this addon out, we will need to support it somehow (either by creating our own binary releases or depending on third party releases).

The problem is that by not doing that, we're basically pushing that pain to our users (which you can see here: https://twitter.com/mixonic/status/696373804210221057 or here: https://github.com/ember-cli/ember-cli/pull/3895 and in many other places). I'm not saying that this means that we must fix it, but I feel that PhantomJS support is just broken at the moment.

A big advantage of doing this ourselves is that we know what changes we ship to our VMs, so we can update the addon accordingly. Our users are usually left in the blue, because they notice only when things break (for example when we update or need to roll back an image).

> For another, only supporting 1.9.8 and 2.0.0 seems arbitrary. 

Sure, it is arbitrary. I'm all for adding 2.1 here too. I haven't done it originally, because I was implementing this thing in first half of January and 2.1 was officially released on 23rd of January. That said, I think that even supporting 3 arbitrary chosen versions is better than having only partial support for some of the versions that depends mostly on which infrastructure you use.

backspace:
Any thoughts on where to go with this, @BanzaiMan? In my personal use of Travis before I joined the team, I went through [much frustration](https://github.com/backspace/adventure-gathering/commit/554774336bb80bcc7f0b6757bb79c08c736b7655#diff-354f30a63fb0907d4ad57269548329e3) getting tests to pass on Travis that were passing locally, mostly due to PhantomJS. It’s confusing for the unfamiliar to find obscure and possibly-untrustworthy links to download prebuilt versions of PhantomJS that are compatible with what you’re using locally. Having explicit support for particular versions would smooth this out for so many people.

PR #598, opened 2016-01-17
Miscellaneous shell script improvements

BanzaiMan:
These are some of the suggestions by http://www.shellcheck.net/.


BanzaiMan:
Too stale now.

PR #601, opened 2016-01-20
Install R and dependencies in the configure block rather than setup

jimhester:
This should allow using `language: r` on the container based infrastructure.

From looking at the [csharp builder](https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/csharp.rb#L16-L58) it looks like `sudo` commands can be used in the `configure()` block for language implementations. This PR simply moves the R `setup()` commands to `configure()` and puts them in a travis fold so the output log is easier to browse.

@BanzaiMan if I am mistaken about this or if there is a more preferred method please let me know.

cc @craigcitro @hadley

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -227,7 +227,7 @@ module Travis
         def r_binary_install(packages)
           return if packages.empty?
           if config[:os] == 'linux'
-            if config[:sudo] == 'false'
+            if config[:sudo] == false
               sh.echo "R binary packages not supported with 'sudo: false', " +
                 ' falling back to source install'
               return r_install packages
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -47,8 +47,6 @@ module Travis
         def configure
           super
 
-          # TODO(craigcitro): Confirm that these do, in fact, print as
-          # green. (They're yellow under vagrant.)
           sh.echo 'R for Travis-CI is not officially supported, ' +
                   'but is community maintained.', ansi: :green
           sh.echo 'Please file any issues using the following link',
@@ -60,7 +58,6 @@ module Travis
 
           sh.fold("R-install") do
             sh.with_options({ assert: true,  echo: true,  timing: true  }) do
-              # TODO(craigcitro): python-software-properties?
               sh.echo 'Installing R', ansi: :yellow
               case config[:os]
               when 'linux'
@@ -169,7 +166,7 @@ module Travis
               sh.failure "Found warnings, treating as errors (as requested)."
             end
           end
-          
+
           # Check revdeps, if requested.
           if config[:r_check_revdep]
             sh.echo "Checking reverse dependencies"
@@ -226,7 +223,7 @@ module Travis
           ].join(' ')
           sh.cmd "Rscript -e '#{install_script}'"
         end
-        
+
         def r_binary_install(packages)
           return if packages.empty?
           case config[:os]
@@ -297,7 +294,7 @@ module Travis
           )
           sh.export 'RCHECK_DIR', "$(Rscript -e '#{pkg_script}')"
         end
-        
+
         def dump_logs
           export_rcheck_dir
           ['out', 'log', 'fail'].each do |ext|
@@ -312,7 +309,7 @@ module Travis
             sh.cmd cmd
           end
         end
-        
+
         def setup_bioc
           unless @bioc_installed
             sh.echo 'Installing BioConductor'
@@ -391,7 +388,7 @@ module Travis
           pandoc_srcdir = "pandoc-#{config[:pandoc_version]}/#{os_path}"
           pandoc_destdir = '${HOME}/opt/pandoc'
           pandoc_tmpfile = "/tmp/pandoc-#{config[:pandoc_version]}.zip"
-          
+
           sh.mkdir pandoc_destdir, recursive: true
           sh.cmd "curl -o #{pandoc_tmpfile} #{pandoc_url}"
           ['pandoc', 'pandoc-citeproc'].each do |filename|
@@ -400,10 +397,9 @@ module Travis
                    "-d #{pandoc_destdir}"
             sh.chmod '+x', "#{File.join(pandoc_destdir, filename)}"
           end
-          
+
           sh.export 'PATH', "$PATH:#{pandoc_destdir}"
         end
-        
       end
     end
   end
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -29,6 +29,7 @@ module Travis
           bioc: 'http://bioconductor.org/biocLite.R',
           bioc_required: false,
           bioc_use_devel: false,
+          R: 'release'
         }
 
         def initialize(data)
@@ -41,7 +42,8 @@ module Travis
 
         def export
           super
-          sh.export 'TRAVIS_R_VERSION', 'release', echo: false
+          sh.export 'TRAVIS_R_VERSION', version, echo: false
+          sh.export 'R_LIBS_USER', '~/R/Library', echo: false
         end
 
         def configure
@@ -121,6 +123,9 @@ module Travis
 
         def install
           super
+          unless setup_cache_has_run_for[:R]
+            setup_cache
+          end
 
           # Install any declared packages
           apt_install config[:apt_packages]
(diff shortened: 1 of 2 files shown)


BanzaiMan:
In general, this approach should work.

jimhester:
Ok great, I will get the tests working and ping you.

jimhester:
Tests now pass https://github.com/jimhester/travis-build/commit/83926b6d949f2ab012a2123d092bed0937f73fd5

lib/travis/build/script/r.rb
```diff
@@ -53,57 +58,61 @@ def setup
...
-
-            sh.echo 'Installing OS X binary package for R'
-            sh.cmd 'sudo installer -pkg "/tmp/R-latest.pkg" -target /'
-            sh.rm '/tmp/R-latest.pkg'
-
-          else
-            sh.failure "Operating system not supported: #{config[:os]}"
+          sh.fold("R-install") do
+            sh.with_options({ assert: true,  echo: true,  timing: true  }) do
+              # TODO(craigcitro): python-software-properties?
```
craigcitro: You can probably delete this -- it was a reminder to myself to confirm we didn't need the `python-software-properties` apt package in order to use `add-apt-repository`. (I guess it's also now available in `software-properties-common` or somesuch.)

lib/travis/build/script/r.rb
```diff
@@ -39,7 +44,7 @@ def export
           sh.export 'TRAVIS_R_VERSION', 'release', echo: false
         end
 
-        def setup
+        def configure
```
craigcitro: will this break folks who are currently overriding `setup` in a `language: r` build, or have their own `configure` step?

craigcitro:
LGTM, just a few questions inline

lib/travis/build/script/r.rb
jimhester: I don't actually think it is possible for users to override `setup` or `configure` (see https://travis-ci.org/jimhester/travis-test/builds/103650273#L86 for an example of trying to do so). Neither is one of the [normal build steps](https://docs.travis-ci.com/user/customizing-the-build/#The-Build-Lifecycle).

lib/travis/build/script/r.rb
jimhester: Removed in https://github.com/jimhester/travis-build/commit/03bd60b44505daffbda461d73b0cac00e62b0811

lib/travis/build/script/r.rb
craigcitro: ok, phew. ;)

craigcitro:
LGTM -- ship it!

BanzaiMan:
This has been deployed to https://staging.travis-ci.org/. Please test it.

jimhester:
@BanzaiMan could you deploy https://github.com/jimhester/travis-build/commit/8c0a9b395feadb286b396a70f9f374dcb3449576 to staging, it contains a small tweak to fallback to R source installs if not using `sudo: true`.

BanzaiMan:
@jimhester It's been deployed to staging.

jimhester:
@BanzaiMan could you deploy https://github.com/travis-ci/travis-build/commit/058bdecd523051c364711443721bb8383a125b92 to staging. It contains a bugfix for checking `sudo: false` (https://github.com/jimhester/travis-build/commit/d5a5da07e5bde79bf909d80d20f73f4f2935f558) and switching to installing pandoc from the official releases rather than rstudio's mirrors.

BanzaiMan:
@jimhester It's been deployed to stating.

jimhester:
https://github.com/jimhester/travis-build/commit/4af3cd0dd2996d4785e19488b028dab1009ab769 Adds support for dependency caching and lays some groundwork for supporting multiple R versions (to be done later). If it could be deployed to staging that would be great!

jimhester:
@BanzaiMan Could you deploy https://github.com/jimhester/travis-build/commit/4af3cd0dd2996d4785e19488b028dab1009ab769 to staging when you have time. Thanks!

BanzaiMan:
Sorry for the delay. This has been deployed to staging.

jimhester:
@hadley, @craigcitro https://github.com/jimhester/travis-build/commit/4af3cd0dd2996d4785e19488b028dab1009ab769 supports caching, but I am not sure what the best syntax we should use is. Currently you turn it on by using a `.travis.yml` file like

```r
language: r
sudo: false
cache: R
```
See [jimhester/travis-test/builds/475863](https://staging.travis-ci.org/jimhester/travis-test/builds/475863) for example build output. (There is network congestion today on travis staging so the build time is much longer for that build than it would be normally)

Other languages use syntax like `cache: bundler` or `cache: ccache` or `cache: pip` (see https://docs.travis-ci.com/user/caching/ for more examples). Any thoughts on what to call Rs version? If we stick with using `cache: R` should we make it 
- Upper case like the `R` executable
- Lower case like `language: r`
- Just make it case insensitive

craigcitro:
We could use `cache: cran` if we want a name that suggests which dependencies we're caching?

I'm also totally fine with `r` -- it would feel weird to have them with different case, like `R CMD BUILD` vs. `R CMD check`. :grin: Could we make the `language` key case-insensitive, too?

jimhester:
I just verified `language:` is case insensitive. Hadley suggested `cache: packages` which is what I am going with. It makes it explicit what is being cached and allows for extensibility later if we want to cache something else, the knitr cache for knitr documents for example...

lib/travis/build/script/r.rb
```diff
@@ -115,6 +124,9 @@ def announce
 
         def install
           super
+          unless setup_cache_has_run_for[:packages]
```
BanzaiMan: This seems a bit unconventional for Ruby code. There is no need for a hash here. For example, an instance variable (e.g., `@cache_setup_has_run`) should do.

lib/travis/build/script/r.rb
```diff
@@ -217,11 +249,15 @@ def r_github_install(packages)
...
           sh.cmd "Rscript -e '#{install_script}'"
         end
-        
+
         def r_binary_install(packages)
           return if packages.empty?
-          case config[:os]
-          when 'linux'
+          if config[:os] == 'linux'
+            if config[:sudo] == false
```
BanzaiMan: This is not an entirely accurate way to check if this build can use `sudo` or not. This is mostly for the historical reason; older repositories defaults to `sudo: required`, while new ones do to `sudo: false`. In other words, the build can lack `sudo` without `sudo: false`.

I am not sure if precise detection is possible from here, unfortunately.

I also believe that the absence of `sudo` should result in `config[:sudo]` being `nil`, so this clause would be skipped.

Perhaps you can approximate by `unless config[:sudo]` (with some false positives related to the default behavior depending on the age of the repository).

lib/travis/build/script/r.rb
jimhester: I was just following the convention for this I saw in other projects ([c](https://github.com/travis-ci/travis-build/blob/8fb10b052ebf0aa23218931acb0728df4e22978f/lib/travis/build/script/c.rb#L32), [objective_c](https://github.com/travis-ci/travis-build/blob/acca7c1638e46324ec81e6e3efdab7b824b22f31/lib/travis/build/script/objective_c.rb#L44), [python](https://github.com/travis-ci/travis-build/blob/df534f3d3cb79d3dc3761e2259da478f8f1b8f74/lib/travis/build/script/python.rb#L42), [bundler](https://github.com/travis-ci/travis-build/blob/8fb10b052ebf0aa23218931acb0728df4e22978f/lib/travis/build/script/shared/bundler.rb#L25), [cpp](https://github.com/travis-ci/travis-build/blob/8fb10b052ebf0aa23218931acb0728df4e22978f/lib/travis/build/script/cpp.rb#L30)). I guess that convention exists if you want to have multiple caches.

In our case we actually do want to have more than one type of cache in the future, so I actually think it makes sense to keep the hash, but if there is a better way to implement this in ruby I am happy to change it.

lib/travis/build/script/r.rb
jimhester: That is a good point. Do you know the approximate date when the default changed from `sudo: required` to `sudo: false`?

`unless config[:sudo]` seems a fine compromise. All `language: r` projects currently have to have sudo enabled (either by default if they are old or with an explicit `sudo: required`). It is also not really a problem if the build runs _with_ sudo, they just won't be able to take advantage of the container based infrastructure and caching.

Thank you for the review and suggestion.

lib/travis/build/script/r.rb
BanzaiMan: Ah. Right. I will have to reread the code I wrote a while back.

jimhester:
I think https://github.com/jimhester/travis-build/commit/61c8b815bf3bb47d64da9a1889e8af6963aeddd4 is probably ready to merge unless you see anything else that needs to be changed.

jimhester:
@BanzaiMan this can be merged unless you have any further review comments.

PR #600, opened 2016-01-20
Retry the apt-get commands to overcome network issues.

andyli:
The apt repos are unstable from time to time, leading to builds failing (e.g. https://travis-ci.org/HaxeFoundation/haxe/builds/103538766).

This PR add retry to the apt-get commands, hopefully will reduce the impact of network issues.

PR #605, opened 2016-01-21
Split apt-get install command into "download" and "install" stages.

andyli:
Turn on assertion to stop proceeding when any of the stages fail.
Retry the "download" stage to overcome network issues.

See #600

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -126,9 +126,11 @@ module Travis
 
             unless whitelisted.empty?
               sh.export 'DEBIAN_FRONTEND', 'noninteractive', echo: true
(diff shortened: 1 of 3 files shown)


BanzaiMan:
I've tested this logical split after resolving conflicts (both textual and logical) in the [ha-feature-split-apt-get-install](https://github.com/travis-ci/travis-build/tree/ha-feature-split-apt-get-install) branch, but it seems to cause some display issue. See https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/491729.

Not clear what the issue is, but it appears to me that some `apt-get` commands keep running in the background and mess up the display.

BanzaiMan:
We've deployed #1343, which adds a different way to add timeouts and retries.

Thanks!

PR #606, opened 2016-01-22
Set PIP_DISABLE_PIP_VERSION_CHECK=1 after showing pip version

BanzaiMan:
Resolves https://github.com/travis-ci/travis-ci/issues/5499

By setting this value, we avoid making extra network calls
for each invocation of `pip`.
We do this *after* announcing the pip version, so that the user
may become aware.


BanzaiMan:
PR build failed because I merged it before test ran.

PR #607, opened 2016-01-24
Fix GemStone-related problems [Smalltalk]

fniephaus:
Unfortunately, we still haven't been able to run GemStone tests successfully on a container due to a couple of problems.
Since Docker doesn't support `hostname` changes within the container, we now just add the `hostname` to `/etc/hosts` which should work just fine as well.
This PR also fixes some bugs that keep the script from installing GemStone dependencies and it removes some redundant folds.

/cc @dalehenrich

### DIFF
--- a/lib/travis/build/script/smalltalk.rb
+++ b/lib/travis/build/script/smalltalk.rb
@@ -5,10 +5,7 @@ module Travis
   module Build
     class Script
       class Smalltalk < Script
-        DEFAULTS = {
-          gemstone_hostname: 'travis.dev',
-          release_version: '#{rel_version}'
-        }
+        DEFAULTS = {}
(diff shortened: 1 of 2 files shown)


BanzaiMan:
@fniephaus Is this ready for merge?

fniephaus:
Yes, I think so. Unfortunately, there is no way to test/debug this, but I hope GemStone builds will finally work after this.
Thanks!

BanzaiMan:
@fniephaus Let's test this next week.

fniephaus:
No worries, I don't have any time to play around with it this week anyway.

fniephaus:
Can we get this in before the weekend? That'd be awesome!

BanzaiMan:
Sorry for the delay. This has been deployed to production.

PR #608, opened 2016-01-25
Updated Texlive support for `language: R`

jimhester:
This PR includes https://github.com/travis-ci/travis-build/pull/601, so once that is merged the diff will be much smaller.

I am submitting it as a separate pull request because those features should be ready to merge, but I would like to test this on staging some, It likely has bugs that will need to be worked out.

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -44,7 +44,7 @@ module Travis
           super
           sh.export 'TRAVIS_R_VERSION', version, echo: false
           sh.export 'R_LIBS_USER', '~/R/Library', echo: false
-          sh.export 'TEXMFHOME', '~/texmf'
+          sh.export 'TEXMFHOME', '~/texmf', echo: false
         end
 
         def configure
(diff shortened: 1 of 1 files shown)


jimhester:
Wanted to note the reason we have to go this route and cannot use the backports is because the official backports (https://launchpad.net/~texlive-backports/+archive/ubuntu/ppa) do not include the `tlmgr` program to allow users to install packages individually. Therefore you have to install some very large dependencies (texlive-fonts-extra in particular is quite large) without being able to pick and choose only the packages needed by your R package.

This approach should result in faster build times for R packages while allowing users more flexibility to install packages that are not included in the default texlive apt packages.

cc @craigcitro @hadley

hadley:
LGTM

craigcitro:
LGTM (I basically just skimmed the last commit)

jimhester:
@BanzaiMan It would be great to have this pushed to staging so I can test it, thanks!

BanzaiMan:

PR #609, opened 2016-01-26
Fix tar extraction

jimhester:
This is a continuation of #601

The tar command has incorrectly specified causing the build to hang waiting for input.


jimhester:
@BanzaiMan if this could get pushed to staging that would be great.

BanzaiMan:
Deployed to staging.

PR #611, opened 2016-01-28
Changes for travis-enterprise

tyranja:
I wanted to test the changes we did for travis-enterprise on staging.org and staging.com. 
I create a commit and checked the logs and the heroku console and the web interface.

On staging.com evertything  works fine. But on staging.org the worker does not seem to process the jobs.
In the rabbit queue dashboard I can't see any docker.builds even through there should be one.

### DIFF
--- a/lib/travis/api/build/sentry.rb
+++ b/lib/travis/api/build/sentry.rb
@@ -1,4 +1,5 @@
 require 'raven'
+require 'raven/integrations/rack'
 require 'sinatra/base'
 
 module Travis


BanzaiMan:
Is there a problem with org staging?

tyranja:
@BanzaiMan Thanks for looking at that. Org staging was broken but its already fixed :+1:

joshk:
@BanzaiMan can you review this when you have a moment

PR #613, opened 2016-02-01
Create RVM aliases to allow short Ruby version

sikachu:
These aliases will overrides image's RVM aliases, and allow user to
specify MRI version using `MAJOR.MINOR` version number.

### DIFF
--- a/lib/travis/build/script/shared/rvm.rb
+++ b/lib/travis/build/script/shared/rvm.rb
@@ -15,6 +15,14 @@ module Travis
           rvm_remote_server_verify_downloads3=1
         )
 
+        # This list should contain the latest patch version of Ruby
+        ALIASES = {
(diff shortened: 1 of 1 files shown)


BanzaiMan:
I will test this shortly.

BanzaiMan:
One concern is that this would add some extra time to builds when the aliased runtime is not pre-installed. Whether or not this is acceptable behavior is up for debate.

sikachu:
@BanzaiMan thanks for testing it out.

> One concern is that this would add some extra time to builds when the aliased runtime is not pre-installed.

It won't be worse than current situation, given the fact that most Ruby projects I seen are specify latest patch version of Ruby that requires installation (2.2.4 and 2.3.0 is not available in the VM, for example).

arthurnn:
❤️

PR #620, opened 2016-02-09
Add hostname addon

BanzaiMan:
#580 sought to implement this functionality as an appliance. Instead of shortening the host name behind the back, make this choice an explicit option.


BanzaiMan:
See https://github.com/travis-ci/travis-ci/issues/5227

BanzaiMan:
@joshk Review, please. This will be useful on on non-container builds, where `sudo hostname` is effective. On containers, the host name is short enough, so this addon is not needed.

PR #630, opened 2016-02-15
Fix build failure on check failure

jimhester:
There was a bug in the R travis script that was failing to fail the travis build when there was a check failure.

This should fix the issue.


jimhester:
It would be great to deploy this fix sooner rather than later, current code does not signal build errors for any build that actually has an package check error.

PR #632, opened 2016-02-16
manually set OS X mountpoint to allow for different volume names

simonbyrne:
Recent changes in Julia (https://github.com/JuliaLang/julia/pull/14470) mean that the OS X dmg volume name is no longer "Julia". This instead manually sets the mountpoint.

cc: @tkelman @staticfloat

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -41,8 +41,9 @@ module Travis
               '| tar -C ~/julia -x -z --strip-components=1 -f -'
           when 'osx'
             sh.cmd %Q{curl -s -L -o julia.dmg '#{julia_url}'}
-            sh.cmd 'hdiutil mount -readonly julia.dmg'
-            sh.cmd 'cp -a /Volumes/Julia/*.app/Contents/Resources/julia ~/'
+            sh.cmd 'mkdir juliamnt'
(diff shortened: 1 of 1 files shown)


tkelman:
Ah, great, I see what you meant by https://github.com/JuliaLang/julia/pull/14470#issuecomment-183379928 then. Thanks!

staticfloat:
The only thing I am concerned about is whether the current working directory is writable.  Should we instead use an absolute path for `juliamnt`, e.g. `~/juliamnt`?

simonbyrne:
We already download `julia.dmg` into the current directory, so it must be writeable.

staticfloat:
Ah, great. No worries then!

tkelman:
Thanks @BanzaiMan!

tkelman:
Given we haven't heard as much from @ninjin these days, I'd be cool with adding @simonbyrne to our list of Julia language community maintainers if he would. Along with this file I think that would also need to be updated in the docs here: https://github.com/travis-ci/docs-travis-ci-com/blob/gh-pages/user/languages/julia.md

ninjin:
@tkelman has an excellent point, given a lot of changes over the last year I have had to focus more on code directly related to my line of research and I am not as active any more when it comes to CI.  @simonbyrne is a stand up guy, do add him.

PR #638, opened 2016-02-20
Allow C++ compiler versions to be specified in yaml-file

jpauwels:
Using a specific compiler version required up to now to manually export $CC and $CXX (a hack that is very common in the wild, see linked issues). This changeset should solve this so you can simply do `compiler: gcc-4.9`. Closes [#3871](https://github.com/travis-ci/travis-ci/issues/3871), partly implements [#979](https://github.com/travis-ci/travis-ci/issues/979), replaces [PR#542](https://github.com/travis-ci/travis-build/pull/542) (which accurately describes the problem, but just makes the hack easier).

Additionally, this solves the minor problem that `announce` displays the given `compiler: anystring` without sanitising the string, such that it prints a different version (the specified `anystring`) than the one that is actually used (the system default if not either `g++` or `clang++` is specified). Finally, this changeset also sets the default OS X compiler to `clang` instead of `gcc`.

### DIFF
--- a/lib/travis/build/script/c.rb
+++ b/lib/travis/build/script/c.rb
@@ -3,7 +3,7 @@ module Travis
     class Script
       class C < Script
         DEFAULTS = {
-          compiler: 'gcc'
+          compiler: ''
         }
 
         def export
@@ -43,7 +43,13 @@ module Travis
         private
 
           def compiler
-            config[:compiler].to_s
+            case config[:compiler].to_s
+            when /^gcc/i, /^clang/i
+              config[:compiler].to_s
+            else
+              { 'linux' => 'gcc', 
+                'osx' => 'clang' }[config[:os]]
+            end
           end
       end
     end
--- a/lib/travis/build/script/cpp.rb
+++ b/lib/travis/build/script/cpp.rb
@@ -3,7 +3,7 @@ module Travis
     class Script
       class Cpp < Script
         DEFAULTS = {
-          compiler: 'g++'
+          compiler: ''
         }
 
         def export
@@ -41,28 +41,30 @@ module Travis
         private
 
           def compiler
-            config[:compiler].to_s
+            cxx
           end
 
           def cxx
-            case compiler
-            when /^gcc/i, /^g\+\+/i then
-              'g++'
-            when /^clang/i, /^clang\+\+/i then
-              'clang++'
+            case config[:compiler].to_s
+            when /^gcc(.*)$/i, /^g\+\+(.*)$/i
+              'g++' + $1
+            when /^clang([^+]*)$/i, /^clang\+\+(.*)$/i
+              'clang++' + $1
             else
-              'g++'
+              { 'linux' => 'g++', 
+                'osx' => 'clang++' }[config[:os]]
             end
           end
 
           def cc
-            case compiler
-            when /^gcc/i, /^g\+\+/i then
-              'gcc'
-            when /^clang/i, /^clang\+\+/i then
-              'clang'
+            case config[:compiler].to_s
+            when /^gcc(.*)$/i, /^g\+\+(.*)$/i
+              'gcc' + $1
+            when /^clang([^+]*)$/i, /^clang\+\+(.*)$/i
+              'clang' + $1
             else
-              'gcc'
+              { 'linux' => 'gcc', 
+                'osx' => 'clang' }[config[:os]]
             end
           end
       end


jpauwels:
Okay, seems like the `spec` needs to be updated for the CI check to pass, but I could use some help for that. I basically learned a bit of Ruby just to fix this issue and I'm not even 100% sure what a `spec` is :-) (some sort of Ruby testing/coverage framework is what I can deduce).

BanzaiMan:
Sorry for the delay.

The failure is coming from here: https://github.com/travis-ci/travis-build/blob/bbe7c12b6f2c8bdc6cd9a7d3e839a729048648ae/spec/build/script/cpp_spec.rb#L88

This is testing whether `gcc --version` should be run when `compiler: gcc`. And I think it should.

lib/travis/build/script/c.rb
```diff
@@ -3,7 +3,7 @@ module Build
     class Script
       class C < Script
         DEFAULTS = {
-          compiler: 'gcc'
+          compiler: ''
```
BanzaiMan: I don't think this is correct.

When `compiler` is not given in `.travis.yml`, the announcement will be ` --version`, which results in "`command not found`" (though not critical), and the cache slug will lack this information (also not critical).

lib/travis/build/script/cpp.rb
```diff
@@ -41,28 +41,30 @@ def use_directory_cache?
         private
 
           def compiler
-            config[:compiler].to_s
+            cxx
```
BanzaiMan: This is causing the spec to fail.

BanzaiMan:
Well, maybe it is legitimate to show `g++ --version` output when we have `compiler: gcc`. Then the spec should be changed.

lib/travis/build/script/c.rb
BanzaiMan: Or maybe not. We should test it.

lib/travis/build/script/c.rb
```diff
@@ -43,7 +43,13 @@ def use_directory_cache?
         private
 
           def compiler
-            config[:compiler].to_s
+            case config[:compiler].to_s
+            when /^gcc/i, /^clang/i
+              config[:compiler].to_s
+            else
+              { 'linux' => 'gcc',
```
BanzaiMan: This construct is rather unsightly.

A more straightforward

```ruby
case config[:os]
when 'osx'
  'clang'
when 'linux'
  'gcc'
end
```

is preferred.

ldionne:
I would love it if we could go forward with this. https://github.com/travis-ci/travis-ci/issues/3871 is very annoying.

lib/travis/build/script/c.rb
jpauwels: What I wanted to achieve was to use different default compilers for Linux (gcc) and Mac (clang). It could be due to my limited knowledge of Ruby and/or Travis, but I failed to setup an if with `:os` at this point, so I kept the default blank and overwrite it later.

lib/travis/build/script/cpp.rb
jpauwels: That doesn't surprise me. The spec should be adapted to follow the changes to `cpp.rb`, but I don't know enough about Ruby to do so.

lib/travis/build/script/c.rb
jpauwels: Your suggestion does not what I want to do :-) Maybe there's a prettier `begins_with` function in Ruby that does what I want, but essentially, I want to check if the string begins with gcc or clang. If so, then leave the suffix (the compiler version) intact, otherwise replace it with a platform-dependent default.

lib/travis/build/script/c.rb
```diff
@@ -43,7 +43,13 @@ def use_directory_cache?
...
           def compiler
-            config[:compiler].to_s
+            case config[:compiler].to_s
+            when /^gcc/i, /^clang/i
+              config[:compiler].to_s
+            else
+              { 'linux' => 'gcc', 
+                'osx' => 'clang' }[config[:os]]
+            end
           end
```
jpauwels: Google taught me there is a `start_with`, so this should be a prettier version.

    if config[:compiler].to_s.start_with?('gcc', 'clang')
      config[:compiler].to_s
    else
      case config[:os]
      when 'osx'
        'clang'
      when 'linux'
        'gcc'
      end
    end

Also, at this point the blank default gets overwritten by the platform specific default. It would be even nicer and maybe necessary for the cache slug generation to move the platform-specific default generation to line 6, and then call `DEFAULTS.compiler` or something, but it isn't required for the platform announcement. And as said before, I don't know how to do that anyway :-)

lib/travis/build/script/cpp.rb
```diff
@@ -41,28 +41,30 @@ def use_directory_cache?
...
+              'gcc' + $1
+            when /^clang([^+]*)$/i, /^clang\+\+(.*)$/i
+              'clang' + $1
             else
-              'gcc'
+              { 'linux' => 'gcc', 
+                'osx' => 'clang' }[config[:os]]
             end
           end
       end
```
jpauwels: Same questions here as for the C case, just a bit more complicated.

jpauwels:
@BanzaiMan I could use some more input on best Ruby/Travis practices, before starting to update the `spec`. Notably, how to set platform specific `DEFAULTS`.

ldionne:
Actually, I don't understand why it's not possible to simply avoid modifying the `compiler` in the `cxx` and `cc` methods, i.e. instead of 

```ruby
def cxx
    case compiler
    when /^gcc/i, /^g\+\+/i then
        'g++'
    when /^clang/i, /^clang\+\+/i then
        'clang++'
    else
        'g++'
    end
end
```

just use

```ruby
def cxx
    compiler
end
```

This way, whatever is being set in `compiler: xxx` in the script will be used as `CXX`, which I think is what we want.

ldionne:
It seems like we don't need all the logic added in this PR (and the removal of the default compiler); we just need to stop overwriting the compiler set by the user.

jpauwels:
@ldionne It's more nice to have than strictly necessary. It's just some string sanitising to prevent users from making mistakes and to make matching pairs of C and C++ compilers.
Anyway, I managed to make the spec pass now. I guess I should add some new test cases too and then wait for @BanzaiMan's input before I squash the commits to merge.

lib/travis/build/script/c.rb
ldionne: Actually, you probably want to avoid setting `compiler` at all, instead of setting it to `''`?

ldionne:
@jpauwels Ok, I agree. For what it's worth, this LGTM. The only thing that I can imagine could break builds is the fact that you changed the default to `clang++` on OS X, but that should be fine since Clang has been the default compiler on OS X for a while and `g++` is even aliased to `clang++`.

brettwooldridge:
Is this pull request going anywhere?

AraHaan:
yeah, it would be nice if travis could automatically switch to whatever compiler version I want without manually having to do ``sudo apt-get install <insert compiler here>`` manually in my ``.travis.yml``.

CJKay:
Bump. This is becoming increasingly necessary for those of us who want to build only with the supported GCC versions (currently 6 and 7).

AraHaan:
I was thinking of this and well what if one wants gcc 8.x but travis has up to 7.x. This has so know if it should install gcc 8 on it’s own or whatever the current version is until travis provides that version too.

PR #651, opened 2016-02-24
Configure /etc/hosts on OS X just like on Linux [Smalltalk]

fniephaus:
Hi @BanzaiMan,

We finally have [GemStone running on containers](https://travis-ci.org/hpi-swa/smalltalkCI/jobs/111612181)!
There still is a `hostname`-related issue that keeps GemStone from also running on OS X.
This PR should fix it (https://github.com/hpi-swa/smalltalkCI/issues/68).

Best,
Fabio


BanzaiMan:
Would you like to test this in staging, or is it good to go now?

fniephaus:
This shouldn't break GemStone on Linux builds, so it can only fix GemStone on OS X builds. Good to go I'd say :)

PR #653, opened 2016-02-25
Patches travis to use bundle exec for cocoapods when it is running ge…

jcampbell05:
…ms from a podfile.

This should fix https://github.com/travis-ci/travis-ci/issues/5694 - When a `Gemfile` is detected we now run `bundler exec pod install` instead of just `pod install` so that the gem is correctly located. When there is no `Gemfile` we default to just `pod install`.

### DIFF
--- a/lib/travis/build/script/objective_c.rb
+++ b/lib/travis/build/script/objective_c.rb
@@ -59,9 +59,19 @@ module Travis
           sh.if podfile? do
             sh.if "! ([[ -f #{pod_dir}/Podfile.lock && -f #{pod_dir}/Pods/Manifest.lock ]] && cmp --silent #{pod_dir}/Podfile.lock #{pod_dir}/Pods/Manifest.lock)", raw: true do
               sh.fold('install.cocoapods') do
-                sh.echo "Installing Pods with 'pod install'", ansi: :yellow
+
                 sh.cmd "pushd #{pod_dir}"
-                sh.cmd 'pod install', retry: true
+
+                sh.if gemfile? do
+                  sh.echo "Installing Pods with 'bundle exec pod install'", ansi: :yellow
+                  sh.cmd 'bundle exec pod install', retry: true
+                end
+
+                sh.elif gemfile? do
+                  sh.echo "Installing Pods with 'pod install'", ansi: :yellow
+                  sh.cmd 'pod install', retry: true
+                end
+
                 sh.cmd 'popd'
               end
             end
--- a/lib/travis/build/script/objective_c.rb
+++ b/lib/travis/build/script/objective_c.rb
@@ -67,7 +67,7 @@ module Travis
                   sh.cmd 'bundle exec pod install', retry: true
                 end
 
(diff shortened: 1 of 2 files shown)


jcampbell05:
@BanzaiMan can you have a look ?

lib/travis/build/script/objective_c.rb
```diff
@@ -59,9 +59,19 @@ def install
...
+
                 sh.cmd "pushd #{pod_dir}"
-                sh.cmd 'pod install', retry: true
+
+                sh.if gemfile? do
+                  sh.echo "Installing Pods with 'bundle exec pod install'", ansi: :yellow
+                  sh.cmd 'bundle exec pod install', retry: true
+                end
+
+                sh.elif gemfile? do
```
BanzaiMan: This `elif` condition seems identical to the immediately preceding `if` condition. I don't think it works.

jcampbell05: What would you suggest changing it to?

BanzaiMan: I think it is just `sh.elf do`. If there is no `Gemfile`, we just run `pod install`, right?

jcampbell05: Exactly

lib/travis/build/script/objective_c.rb
```diff
@@ -59,9 +59,19 @@ def install
...
+
                 sh.cmd "pushd #{pod_dir}"
-                sh.cmd 'pod install', retry: true
+
+                sh.if gemfile? do
+                  sh.echo "Installing Pods with 'bundle exec pod install'", ansi: :yellow
+                  sh.cmd 'bundle exec pod install', retry: true
+                end
+
+                sh.elf do
```
BanzaiMan: Missed an `i`.

jcampbell05:
@BanzaiMan Updated, please let me know if it looks good.

jcampbell05:
@BanzaiMan Hey this is ready.

BanzaiMan:
Sorry about the delay. We need specs for this.

jcampbell05:
@BanzaiMan  how would I spec this ?

BanzaiMan:
Hmm. The spec fails when I run it on my machine. I'll investigate later.

BanzaiMan:
OK. I've figured out the failure. I'll merge this, and add a fix to the spec failure.

BanzaiMan:
I reverted this due to reports of syntax error. For example:

```
/Users/travis/build.sh: line 571: syntax error in conditional expression
/Users/travis/build.sh: line 571: syntax error near `:echo=>'
/Users/travis/build.sh: line 571: ` elif [[ {:assert=>true, :echo=>true, :timing=>true} ]]; then'
/Users/travis/build.sh: line 571: shell_session_update: command not found
```

I'll take another look soon.

jcampbell05:
@BanzaiMan cheers, it will be a big help!

PR #657, opened 2016-02-26
Use sudo for all tlmgr calls

jimhester:
Changing permissions (even recursively) still results in errors
unfortunately.


jimhester:
ping @BanzaiMan, OS X builds are currently fail with `language: r` due to this bug, so getting it merged would be great!

PR #658, opened 2016-02-27
Check for Go 1.5 vendoring before Godeps init.

paddyforan:
If we are using Go 1.5's vendoring method, we could have a
Godeps/Godeps.json file and _not_ want to run `godep restore`. Godep is
used as a tool to manage dependencies under Go 1.5's vendoring method,
and `godep restore` is _not_ supposed to be run in those cases.

This fixes that by adding a check to see if we're using Go 1.5's
vendoring method before we check if godep is being used, and skipping
the godep section if we're using 1.5's vendoring method instead.

See travis-ci/travis-ci#5702.

### DIFF
--- a/lib/travis/build/script/go.rb
+++ b/lib/travis/build/script/go.rb
@@ -67,12 +67,12 @@ module Travis
           sh.if uses_15_vendoring do
             sh.echo 'Using Go 1.5 Vendoring, not checking for Godeps'
           end
-	  sh.else do
-	    sh .if '-f Godeps/Godeps.json' do
+          sh.else do
+            sh .if '-f Godeps/Godeps.json' do
               sh.export 'GOPATH', '${TRAVIS_BUILD_DIR}/Godeps/_workspace:$GOPATH', retry: false
               sh.export 'PATH', '${TRAVIS_BUILD_DIR}/Godeps/_workspace/bin:$PATH', retry: false
 
-              if go_version >= '1.1'
+              if go_version != 'go1' && comparable_go_version >= Gem::Version.new('1.1')
                 sh.if '! -d Godeps/_workspace/src' do
                   sh.cmd "#{go_get_cmd} github.com/tools/godep", echo: true, retry: true, timing: true, assert: true
                   sh.cmd 'godep restore', retry: true, timing: true, assert: true, echo: true
@@ -107,12 +107,13 @@ module Travis
           def uses_make
             '-f GNUmakefile || -f makefile || -f Makefile || -f BSDmakefile'
           end
-
-	  def uses_15_vendoring
-            if (go_version < '1.5' || go_version == 'go1')
-              'false'
+          
+          # see https://golang.org/doc/go1.6#go_command
+          def uses_15_vendoring
+            if (go_version == 'go1' || comparable_go_version < Gem::Version.new('1.5'))
+              return 'false'
             end
-	    (go_version < '1.6') ? '$GO15VENDOREXPERIMENT == 1' : '$GO15VENDOREXPERIMENT != 0'
+            (comparable_go_version < Gem::Version.new('1.6')) ? '$GO15VENDOREXPERIMENT == 1' : '$GO15VENDOREXPERIMENT != 0'
           end
 
           def gobuild_args
@@ -140,8 +141,15 @@ module Travis
             end
           end
 
+          def comparable_go_version
+            if !go_version[/^[0-9]/] # if we don't have a semver version that Gem::Version can read
+              return Gem::Version.new('0.0.1') # return a consistent result
+            end
+            Gem::Version.new(go_version)
+          end
+
           def go_get_cmd
-            (go_version < '1.2' || go_version == 'go1') ? 'go get' : 'go get -t'
+            (go_version == 'go1' || comparable_go_version < Gem::Version.new('1.2')) ? 'go get' : 'go get -t'
           end
 
           def ensure_gvm_wiped
--- a/lib/travis/build/script/go.rb
+++ b/lib/travis/build/script/go.rb
@@ -64,14 +64,19 @@ module Travis
         end
 
         def install
-          sh.if '-f Godeps/Godeps.json' do
-            sh.export 'GOPATH', '${TRAVIS_BUILD_DIR}/Godeps/_workspace:$GOPATH', retry: false
-            sh.export 'PATH', '${TRAVIS_BUILD_DIR}/Godeps/_workspace/bin:$PATH', retry: false
+          sh.if uses_15_vendoring do
+            sh.echo 'Using Go 1.5 Vendoring, not checking for Godeps'
+          end
+	  sh.else do
+	    sh .if '-f Godeps/Godeps.json' do
+              sh.export 'GOPATH', '${TRAVIS_BUILD_DIR}/Godeps/_workspace:$GOPATH', retry: false
+              sh.export 'PATH', '${TRAVIS_BUILD_DIR}/Godeps/_workspace/bin:$PATH', retry: false
 
-            if go_version >= '1.1'
-              sh.if '! -d Godeps/_workspace/src' do
-                sh.cmd "#{go_get_cmd} github.com/tools/godep", echo: true, retry: true, timing: true, assert: true
-                sh.cmd 'godep restore', retry: true, timing: true, assert: true, echo: true
+              if go_version >= '1.1'
+                sh.if '! -d Godeps/_workspace/src' do
+                  sh.cmd "#{go_get_cmd} github.com/tools/godep", echo: true, retry: true, timing: true, assert: true
+                  sh.cmd 'godep restore', retry: true, timing: true, assert: true, echo: true
+                end
               end
             end
           end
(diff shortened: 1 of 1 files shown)


dougnukem:
:+1: this would be nice to use currently I'm having to disable the install step:

```yml
# Skip godep restore during install because we vendor our dependencies using godep
install:
  - true
```

BanzaiMan:
Hello, is anyone available to test this on staging? https://staging.travis-ci.org

lib/travis/build/script/go.rb
```diff
@@ -103,6 +108,13 @@ def uses_make
             '-f GNUmakefile || -f makefile || -f Makefile || -f BSDmakefile'
           end
 
+	  def uses_15_vendoring
+            if (go_version < '1.5' || go_version == 'go1')
+              'false'
+            end
+	    (go_version < '1.6') ? '$GO15VENDOREXPERIMENT == 1' : '$GO15VENDOREXPERIMENT != 0'
```
tianon: IMO (not maintainer, just an outside observation) wouldn't it make sense to include a link to https://golang.org/doc/go1.6#go_command here? (so it's more clear why this is ternary based on whether we're `> '1.6'`)

Also, will this `go_version` string comparison work if/when Go reaches 1.10?  Is there some stdlib Ruby version library that could be used instead?

tianon:
Also worth noting that whitespace here appears to be a fun mix of spaces and tabs in a few places. :smile:

tianon:
(The general intent of the logic seems sound IMO though.)

lib/travis/build/script/go.rb
```diff
@@ -103,6 +108,13 @@ def uses_make
             '-f GNUmakefile || -f makefile || -f Makefile || -f BSDmakefile'
           end
 
+	  def uses_15_vendoring
+            if (go_version < '1.5' || go_version == 'go1')
```
sirsean: This isn't a bug _yet_, but it will be once there's a version 1.10 (I don't know Go's plans for versioning that far into the future but this is waiting to blow up in a few years.)

```ruby
'1.10' < '1.5'
# => true
```

May want to write a version parsing function?

lib/travis/build/script/go.rb
```diff
@@ -103,6 +108,13 @@ def uses_make
...
+              'false'
+            end
```
sirsean: Also if this doesn't become an else, that `'false'` will never be used. (The return value of this function will always be the line below.)

sirsean: Which notably does what you might not expect when `go_version == 'go1'` as you've called out in the `if` above.

lib/travis/build/script/go.rb
tianon: http://stackoverflow.com/a/2051253/433558 is a solution that requires a gem (http://github.com/dazuma/versionomy), but http://stackoverflow.com/a/2051296/433558 has a pretty simple solution that gives simple comparability via `<=>` (split on `.`, `.to_i` each part, use `<=>` for comparing the arrays)

tianon: There's also `Gem::Version` as explained in http://stackoverflow.com/a/3064161/433558 that's likely viable.

BanzaiMan: Indeed, we already use `Gem::Version` in Elixir. See https://github.com/travis-ci/travis-build/blob/c4c406ff259c47fc5d60a20a1c55482ab3b49727/lib/travis/build/script/elixir.rb#L67-L69

paddyforan: I was trying to match the checks already present in the file, but am happy to amend this to use `Gem::Version`. :)

lib/travis/build/script/go.rb
paddyforan: Oops! Good catch. Sorry about that, I'll fix it up.

lib/travis/build/script/go.rb
paddyforan: Good idea with the docstring. `Gem::Version` was discussed above, so I'll give that a shot and see if I can't make it work. I'm not actually a Rubyist, but this was straightforward enough I took a crack at it anyways.

paddyforan:
> Hello, is anyone available to test this on staging? https://staging.travis-ci.org

@BanzaiMan I'll be happy to run some tests against this in staging to exercise a few different workloads (different versions of Go, vendored, non-vendored, not-officially-supported vendoring) and see if any of them have issues. I'm just not going to have the time or bandwidth to tackle this for a few days, unfortunately.

> Also worth noting that whitespace here appears to be a fun mix of spaces and tabs in a few places. :smile:

@tianon Sorry! Improper vim setup, apparently. I'll take a look at it and see if I can't resolve that.

lib/travis/build/script/go.rb
BanzaiMan: Thanks for the attention you gave to this PR. If you could start here (and maybe fix other places later), that'll be great.

BanzaiMan:
@paddyforan Has the remaining issues been addressed?

paddyforan:
Hey there! Apologies on the long delay here; Things kept me busy for a couple of weeks. I've made the requested changes, and am running the test suite on them now. If it all looks good, I'll push the changes and be ready for another round of review.

I'll also attempt to get the following projects set up in Github, so that they can be tested against staging:

- A go1.6 project using godep
- A go1.6 project not using godep
- A go1.5 project with GO15VENDOREXPERIMENT set to 1 using godep
- A go1.5 project with GO15VENDOREXPERIMENT set to 1 not using godep
- A go1.5 project with GO15VENDOREXPERIMENT set to 0 using godep
- A go1.5 project with GO15VENDOREXPERIMENT set to 0 not using godep
- A go1 project using godep
- A go1 project not using godep

I _believe_ if we run it against all those projects and get the expected results, it _should_ be safe to deploy.

paddyforan:
OK, just pushed changes. Should be ready for another round of review, which I will hopefully respond to faster this time. :)

paddyforan:
I also updated all version comparisons to use `Gem::Version`, for consistency's sake. Also, we were failing a test when forcing `Gem::Version` about passing through random values, so I had non-semver versions return a version of `0.0.1`. Hopefully this makes sense, and is at least a consistent result. If another approach is desired, I'm happy to do that, too.

njern:
Looking forward to see this make it into production 👍

BanzaiMan:
I'll take another look soon.

BanzaiMan:
This has been deployed to .org production. Could someone confirm that it is working to their satisfaction?

paddyforan:
I'll run some tests tonight. Thanks!

lib/travis/build/script/go.rb
```diff
@@ -64,14 +64,19 @@ def setup
...
-            if go_version >= '1.1'
-              sh.if '! -d Godeps/_workspace/src' do
-                sh.cmd "#{go_get_cmd} github.com/tools/godep", echo: true, retry: true, timing: true, assert: true
-                sh.cmd 'godep restore', retry: true, timing: true, assert: true, echo: true
+          sh.if uses_15_vendoring do
+            sh.echo 'Using Go 1.5 Vendoring, not checking for Godeps'
+          end
+          sh.else do
+            sh .if '-f Godeps/Godeps.json' do
+              sh.export 'GOPATH', '${TRAVIS_BUILD_DIR}/Godeps/_workspace:$GOPATH', retry: false
```
freeformz: FWIW: There is no reason to do this if you issue a `godep restore` as `godep restore` checks out the revs recorded in Godeps.json and installs them into your $GOPATH. So either do this *OR* do a restore. Also `godep go <cmd>` does this for you.

PS: Hopefully this will all be moot at some point anyway since vendor.

freeformz:
Is this deployed to .com ?

BanzaiMan:
@freeformz Not to .com yet.

BanzaiMan:
@paddyforan Have you had time to test this?

@freeformz Do you think your concern is serious enough that we should revert this?

freeformz:
I'm just pointing out that it's not that straight forward. I personally wouldn't bother with a `godep restore` at all, but people use godep in wild ways, so this seems fine. I'm just wanting this on .com ATM.

BanzaiMan:
@freeformz Thanks for the info. I've deployed this to .com production.

lib/travis/build/script/go.rb
paddyforan: Should be noted: I didn't add this code; it's the same from what was line 68 previously.

paddyforan:
> @paddyforan Have you had time to test this?

Apologies, I have not. It's the first item on my to-do list, though comprehensive tests will probably have to wait until tomorrow night due to some health stuff.

paddyforan:
I've created the paddyforan/travis-test-godep-vendored, paddyforan/travis-test-godep-unused, paddyforan/travis-test-godep-workspace, and paddyforan/travis-test-godep-workspace-rewrite repositories, which all import paddyforan/travis-test-dependency. I think that represents every possible variation of using godep:

- not using it
- using it with the new vendor folder
- using it with the Godeps/_workspace folder
- using it with the Godeps/_workspace folder and rewrites

I'll get each repo set up in Travis next, and have them run against go1, 1.1, 1.2, 1.3, 1.4, 1.5, 1.6, and tip. I'll also use GO15VENDOREXPERIMENT set to 1 and 0 for each of those variations.

That should exercise pretty much every permutation of the code, as integration tests.

PR #662, opened 2016-03-02
Set shmall kernel variable correctly [Smalltalk]

fniephaus:
Hi @BanzaiMan,

Once this is merged, GemStone will finally also run on OS X.

Thanks,
Fabio

This fixes https://github.com/hpi-swa/smalltalkCI/issues/68


fniephaus:
Mh, this is merged but doesn't seem to be in production. Are you sure it is, @BanzaiMan?

PR #665, opened 2016-03-04
Crystal: Install gmp

jhass:
So users don't have to in case they want to use the standard library
binding to it.

closes travis-ci/travis-ci#5392

/cc @asterite @waj @will


will:
Good call, looks good

asterite:
Nice. What about other libraries like libxml2 and libyaml?

jhass:
I think they're already in the image, given needed to compile Ruby and nokogiri.

dkhofer:
Sweet, thanks for adding gmp!

PR #666, opened 2016-03-04
Crystal: Fetch nightlies over HTTPS

jhass:
Since at least they are available over HTTPS now, make use of that.

/cc @asterite @waj @will


will:
Nice

asterite:
:+1:

PR #652, opened 2016-02-25
Install mariadb-server and mariadb-server-specific-version

sodabrew:
This should resolve the errors I'm seeing here, consistent for all three major MariaDB versions:

https://travis-ci.org/brianmario/mysql2/jobs/111612273
https://travis-ci.org/brianmario/mysql2/jobs/111612274
https://travis-ci.org/brianmario/mysql2/jobs/111612275

```
0.25s$ sudo apt-get install -o Dpkg::Options::='--force-confnew' mariadb-server libmariadbclient-dev
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
The following packages have unmet dependencies:
 mariadb-server : Depends: mariadb-server-5.5 (= 5.5.48+maria-1~trusty) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```

PR #667, opened 2016-03-04
Add -y to apt-get install mariadb so it does't hang with a prompt

sodabrew:
PR #652 fixed MariaDB installation and broke MariaDB installation.


BanzaiMan:
Fix the spec, and I can merge it.

sodabrew:
Fixed!

PR #668, opened 2016-03-07
Supply target parameter for OS X builds

jimhester:


jimhester:
Found another issue with OS X builds with `language: r`.

Not having a VM for this is painful, I am sorry for the inconvenience :(

PR #676, opened 2016-03-09
RethinkDB addon (WIP)

danielmewes:
This adds a RethinkDB addon.

The supplied test is currently failing as follows, and I'm not sure what that means:
```
Failures:

  1) Travis::Build::Addons::Rethinkdb 
     Failure/Error: raise TaintedOutput if !ignore_taint && script.tainted?

     Travis::Shell::Generator::TaintedOutput:
       Travis::Shell::Generator::TaintedOutput
     # ./lib/travis/shell/generator.rb:16:in `generate'
     # ./lib/travis/shell.rb:10:in `generate'
     # spec/spec_helpers/sexp.rb:55:in `store_example'
     # ./spec/build/addons/rethinkdb_spec.rb:12:in `block (2 levels) in <top (required)>'

  2) Travis::Build::Addons::Rethinkdb behaves like compiled script can be compiled
     Failure/Error: expect { subject }.to_not raise_error

       expected no Exception, got #<Travis::Shell::Generator::TaintedOutput: Travis::Shell::Generator::TaintedOutput> with backtrace:
         # ./lib/travis/shell/generator.rb:16:in `generate'
         # ./lib/travis/shell.rb:10:in `generate'
         # ./lib/travis/build/script.rb:72:in `compile'
         # spec/build/script/shared/script.rb:7:in `block (2 levels) in <top (required)>'
         # spec/build/script/shared/script.rb:14:in `block (3 levels) in <top (required)>'
         # spec/build/script/shared/script.rb:14:in `block (2 levels) in <top (required)>'
     Shared Example Group: "compiled script" called from ./spec/build/addons/rethinkdb_spec.rb:14
     # spec/build/script/shared/script.rb:14:in `block (2 levels) in <top (required)>'

  3) Travis::Build::Addons::Rethinkdb behaves like compiled script includes the expected travis_cmds
     Failure/Error: raise TaintedOutput if !ignore_taint && script.tainted?

     Travis::Shell::Generator::TaintedOutput:
       Travis::Shell::Generator::TaintedOutput
     Shared Example Group: "compiled script" called from ./spec/build/addons/rethinkdb_spec.rb:14
     # ./lib/travis/shell/generator.rb:16:in `generate'
     # ./lib/travis/shell.rb:10:in `generate'
     # ./lib/travis/build/script.rb:72:in `compile'
     # spec/build/script/shared/script.rb:7:in `block (2 levels) in <top (required)>'
     # spec/build/script/shared/script.rb:25:in `block (3 levels) in <top (required)>'
     # spec/build/script/shared/script.rb:24:in `each'
     # spec/build/script/shared/script.rb:24:in `block (2 levels) in <top (required)>'

Finished in 16.45 seconds (files took 1.77 seconds to load)
3090 examples, 3 failures, 11 pending

Failed examples:

rspec ./spec/build/addons/rethinkdb_spec.rb:12 # Travis::Build::Addons::Rethinkdb 
rspec ./spec/build/addons/rethinkdb_spec.rb[1:2:1] # Travis::Build::Addons::Rethinkdb behaves like compiled script can be compiled
rspec ./spec/build/addons/rethinkdb_spec.rb[1:2:3] # Travis::Build::Addons::Rethinkdb behaves like compiled script includes the expected travis_cmds
```

Mentioning @joshk who suggested opening the pull request despite the remaining failures.


lib/travis/build/addons/rethinkdb.rb
```diff
@@ -0,0 +1,39 @@
...
+    class Addons
+      class Rethinkdb < Base
+        SUPER_USER_SAFE = true
+
+        RETHINKDB_GPG_KEY = '0x3A8F2399'
+
+        def after_prepare
+          sh.fold 'rethinkdb' do
+            sh.echo "Installing RethinkDB version #{rethinkdb_version}", ansi: :yellow
+            sh.cmd "service rethinkdb stop", sudo: true
```
BanzaiMan: We have not been very vigilant about generating multi-platform scripts, but I think we should bail out if this addon runs on OS X.

lib/travis/build/addons/rethinkdb.rb
```diff
@@ -0,0 +1,39 @@
...
+            sh.cmd "apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 #{RETHINKDB_GPG_KEY}", sudo: true
+            sh.cmd 'add-apt-repository "deb http://download.rethinkdb.com/apt $(lsb_release -cs) main"', sudo: true
+            sh.cmd "apt-get update -qq", assert: false, sudo: true
+            package_version = rethinkdb_version + `lsb_release -cs`
```
BanzaiMan: This is incorrect. <tt>`lsb_releases -cs`</tt> will have to be evaluated at the script's runtime, not the compile time on `travis-build`.

Do something like:

```ruby
sh.cmd "package_version=#{rethinkdb_version}$(lsb_release -cs)"
sh.cmd "apt-get install -y -o Dpkg::Options::='--force-confnew' rethinkdb=$package_version", sudo: true, echo: true, timing: true
```

lib/travis/build/addons/rethinkdb.rb
```diff
@@ -0,0 +1,39 @@
...
+            sh.cmd "apt-get install -y -o Dpkg::Options::='--force-confnew' rethinkdb=#{package_version}", sudo: true, echo: true, timing: true
+            sh.echo "Installing RethinkDB default instance configuration"
+            sh.cmd "cp /etc/rethinkdb/default.conf.sample /etc/rethinkdb/instances.d/default.conf", sudo: true
+            sh.echo "Starting RethinkDB v#{rethinkdb_version}", ansi: :yellow
+            sh.cmd "service rethinkdb start", sudo: true, assert: false, echo: true, timing: true
+            sh.export 'TRAVIS_RETHINKDB_VERSION', rethinkdb_version, echo: false
+            sh.export 'TRAVIS_RETHINKDB_PACKAGE_VERSION', package_version, echo: false
```
BanzaiMan: This will need to be fixed as well, based on the comment above.

BanzaiMan:
Not sure where the taintedness is coming into play, but with my fixes, that is not a problem when I run the specs locally.

BanzaiMan:
A possible fix for these issues: https://gist.github.com/BanzaiMan/20cc01553a1948236138

danielmewes:
Thanks for the feedback and proposed fixes @BanzaiMan . I'll update the PR shortly.

Do you have any suggestions or example code for how to check for the operating system?

danielmewes:
I applied your fixes @BanzaiMan and the tests are now passing. However I couldn't get the OS check to work. I tried checking `ENV['TRAVIS_OS_NAME']`, but it was always empty during testing. Any ideas?

BanzaiMan:
@danielmewes How are you testing it?

danielmewes:
@BanzaiMan I was using the Vagrant file and running `bundle exec rspec spec` as described in the README https://github.com/travis-ci/travis-build/blob/master/README.md .

lib/travis/build/addons/rethinkdb.rb
```diff
@@ -0,0 +1,42 @@
...
+  module Build
+    class Addons
+      class Rethinkdb < Base
+        SUPER_USER_SAFE = true
+
+        RETHINKDB_GPG_KEY = '0x3A8F2399'
+
+        def after_prepare
+          sh.fold 'rethinkdb' do
+            #if ENV['TRAVIS_OS_NAME'] != 'linux'
```
BanzaiMan: If you want to warn OS X builds, you can do something like this:

```ruby
sh.fold 'rethinkdb' do
  sh.if "ENV['TRAVIS_OS_NAME'] != 'linux'" do
    sh.echo "The RethinkDB addon only works on Linux. $TRAVIS_OS_NAME is #{ENV['TRAVIS_OS_NAME']}.", ansi: :red
  end
  sh.else do
    sh.echo "Installing RethinkDB version #{rethinkdb_version}", ansi: :yellow
    ⋮
  end
end
```

BanzaiMan:
@danielmewes Do you need help?

danielmewes:
@BanzaiMan Thanks for your comments. Picking this up again now...

danielmewes:
@BanzaiMan Updated as suggested. This seems to be working now.

hueniverse:
What's the issue here? What needs to happen to get this merged? Who is this blocked with?

lib/travis/build/addons/rethinkdb.rb
```diff
@@ -0,0 +1,44 @@
...
+  module Build
+    class Addons
+      class Rethinkdb < Base
+        SUPER_USER_SAFE = true
+
+        RETHINKDB_GPG_KEY = '0x3A8F2399'
+
+        def after_prepare
+          sh.fold 'rethinkdb' do
+            sh.if "ENV['TRAVIS_OS_NAME'] != 'linux'" do
```
BanzaiMan: Keep in mind that the build script has to be compiled by the `travis-build` app, then sent to the VM that actually runs it.

`ENV['TRAVIS_OS_NAME']` is the value defined on the machine that compiles it, not the one defined on the VM that runs the build.

Also, `bash` will not allow this condition in the `if`.

lib/travis/build/addons/rethinkdb.rb
```diff
@@ -0,0 +1,44 @@
...
+              sh.echo "The RethinkDB addon only works on Linux. $TRAVIS_OS_NAME is #{ENV['TRAVIS_OS_NAME']}.", ansi: :red
```
BanzaiMan: The `TaintedOutput` exceptions are coming from here: https://travis-ci.org/travis-ci/travis-build/builds/124611026#L4673

`ENV['TRAVIS_OS_NAME']` is to blame.

Furthermore, at the point when this code is executed, `$TRAVIS_OS_NAME` is not yet defined, so you'll get an empty string there.

lib/travis/build/addons/rethinkdb.rb
BanzaiMan: I suggest:

```ruby
sh.if "$(uname) != 'Linux'" do
⋮
```

lib/travis/build/addons/rethinkdb.rb
BanzaiMan: I think `"The RethinkDB addon only works on Linux."` is sufficient for the message.

BanzaiMan: RethinkDB appears to work on the Mac, so I am not sure why you want to limit the scope.

hueniverse:
@danielmewes @BanzaiMan Who are we waiting on at this point? I am not sure if more changes are needed from @danielmewes or if @BanzaiMan needs to help fix something. Either way, this was supposed to be a quick solution back in March. It's now June...

Thanks!

danielmewes:
Thanks for the explanation @BanzaiMan .
I've updated the OS check.

RethinkDB works on Mac, but the setup procedure is entirely different and I don't want to deal with that now.

BanzaiMan:
@danielmewes Thanks for the update. I ran a quick test to install 2.3, but it is failing: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/493918#L143-L166

```
Installing RethinkDB version 2.3
rethinkdb: unrecognized service
Executing: gpg --ignore-time-conflict --no-options --no-default-keyring --secret-keyring /tmp/tmp.rvZe4OH5c8 --trustdb-name /etc/apt/trustdb.gpg --keyring /etc/apt/trusted.gpg --primary-keyring /etc/apt/trusted.gpg --keyring /etc/apt/trusted.gpg.d//apt.postgresql.org.gpg --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0x3A8F2399
gpg: requesting key 3A8F2399 from hkp server keyserver.ubuntu.com
gpg: key 3A8F2399: public key "RethinkDB Packaging <packaging@rethinkdb.com>" imported
gpg: Total number processed: 1
gpg:               imported: 1  (RSA: 1)
W: Failed to fetch http://download.rethinkdb.com/apt/dists/precise/Release  Unable to find expected entry 'main/source/Sources' in Release file (Wrong sources.list entry or malformed file)
E: Some index files failed to download. They have been ignored, or old ones used instead.
$ sudo apt-get install -y -o Dpkg::Options::='--force-confnew' rethinkdb=$package_version
Reading package lists... Done
Building dependency tree       
Reading state information... Done
E: Unable to locate package rethinkdb
Installing RethinkDB default instance configuration
cp: cannot stat `/etc/rethinkdb/default.conf.sample': No such file or directory
Starting RethinkDB v2.3
$ sudo service rethinkdb start
rethinkdb: unrecognized service
$ rethinkdb --version
/home/travis/build.sh: line 45: rethinkdb: command not found
```

Could you review these messages?

Also, do you want the build to fail immediately if installation fails for one reason or another?

danielmewes:
Ah, I think this is the same issue as https://github.com/travis-ci/travis-ci/issues/4772 . I'll change the `add-apt-repository` line...

danielmewes:
Pushed a new version. This should avoid the issue with `add-apt-repository`.

hueniverse:
@BanzaiMan Ping!

BanzaiMan:
@hueniverse @danielmewes Sorry for the delay. I have deployed this to org staging.

hueniverse:
@BanzaiMan thanks! What's the next step? ETA on production availability?

BanzaiMan:
Test, please.

BanzaiMan:
Is it intended that this addon works only with `sudo: required`?

danielmewes:
@BanzaiMan Yes, the way this is currently written requires `sudo`. Doing this without `sudo` would unfortunately be more complicated and fragile (we would have to extract the binary from the .deb package manually).

lib/travis/build/addons/rethinkdb.rb
```diff
@@ -0,0 +1,44 @@
+require 'shellwords'
+require 'travis/build/addons/base'
+
+module Travis
+  module Build
+    class Addons
+      class Rethinkdb < Base
+        SUPER_USER_SAFE = true
```
BanzaiMan: This should be `false` if it doesn't work on `sudo`-less environment.

lib/travis/build/addons/rethinkdb.rb
```diff
@@ -0,0 +1,44 @@
...
+        def after_prepare
+          sh.fold 'rethinkdb' do
+            sh.if "$(uname) != 'Linux'" do
+              sh.echo "The RethinkDB addon only works on Linux.", ansi: :red
+            end
+            sh.else do
+              sh.echo "Installing RethinkDB version #{rethinkdb_version}", ansi: :yellow
+              sh.cmd "service rethinkdb stop", sudo: true
+              sh.cmd "apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 #{RETHINKDB_GPG_KEY}", sudo: true
+              sh.cmd 'echo -e "\ndeb http://download.rethinkdb.com/apt $(lsb_release -cs) main" >> /etc/apt/sources.list', sudo: true
```
BanzaiMan: This doesn't work, because `sudo` cannot redirect output this way.

```ruby
sh.cmd 'echo -e "\ndeb http://download.rethinkdb.com/apt $(lsb_release -cs) main" | sudo tee -a /etc/apt/sources.list >/dev/null'
```

lib/travis/build/addons/rethinkdb.rb
```diff
@@ -0,0 +1,44 @@
...
+              sh.cmd "apt-get update -qq", assert: false, sudo: true
+              sh.cmd "package_version=#{rethinkdb_version}$(lsb_release -cs)"
```
BanzaiMan: Is this a correct version specification? We may end up with something like `2.3trusty`, which doesn't seem to exist. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/505561#L166

hueniverse:
@BanzaiMan does this mean this will not work with the new infrastructure?

@danielmewes requiring sudo makes this a lot less useful because it blocks it out of the new infrastructure. Goal is to make using rethink as easy as all the other built-in dbs available in Travis.

danielmewes:
@BanzaiMan Thanks for the comments. I'll look into those things.

@hueniverse Both the MariaDB (which I based this on) and PostgreSQL addons also require sudo on the destination system. Do you have an example for a DB addon that works without it that I could look at?

lib/travis/build/addons/rethinkdb.rb
danielmewes: Right, it doesn't currently correspond to a version in the APT repository. I should probably change it to get the version from `apt-cache` or a similar command...

lib/travis/build/addons/rethinkdb.rb
danielmewes: How does the `sh.cmd` command work if `sudo` is `true`? Shouldn't it execute the whole command in a `sudo` subshell?

BanzaiMan:
If the only requirement is `sudo`, you can do stuff in the `configure` phase instead.

BanzaiMan:
Scratch that. `after_prepare` also has `sudo` (and you are using it).

lib/travis/build/addons/rethinkdb.rb
BanzaiMan: `sudo: true` option just prepends the command with `sudo `. This is not good for your use case.

hueniverse:
@BanzaiMan So we are good as is with the sudo requirements?

BanzaiMan:
@hueniverse Yes, I think so. The version specification still needs to be fixed.

BanzaiMan:
Also a PR to docs, please.

hueniverse:
@BanzaiMan Can you point me where those docs are and a similar section to based this on?

BanzaiMan:
@hueniverse It should go into https://github.com/travis-ci/docs-travis-ci-com/blob/793a3d4a91f39987fc4392b642d60434e7f1f295/user/database-setup.md. (Which is rendered at https://docs.travis-ci.com/user/database-setup/). You can see similar sections (e.g., MariaDB) on that page.

danielmewes:
@BanzaiMan I fixed the `SUPER_USER_SAFE` line, the line for piping into `/etc/apt/sources.list`, and re-implemented how the RethinkDB package version is determined. It is now possible to set rethinkdb_version to any prefix of the target version, e.g. "2.3".
The script will then install the highest version with that prefix, e.g. "2.3.4~0precise".

danielmewes:
Documentation is coming.

danielmewes:
Documentation is in https://github.com/travis-ci/docs-travis-ci-com/pull/673

hueniverse:
@danielmewes you're my hero!

BanzaiMan:
OK, this has been deployed to staging one more time.

BanzaiMan:
Test: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/506951#L166

lib/travis/build/addons/rethinkdb.rb
```diff
@@ -0,0 +1,44 @@
+require 'shellwords'
+require 'travis/build/addons/base'
+
+module Travis
+  module Build
+    class Addons
+      class Rethinkdb < Base
+        SUPER_USER_SAFE = false
```
BanzaiMan: Since everything that requires `sudo` happens before `sudo` is disabled, you can enable this on containers:

```ruby
SUPER_USER_SAFE = true
```

danielmewes:
Thanks for testing and your comments @BanzaiMan . I turned `SUPER_USER_SAFE` back on.

BanzaiMan:
Deployed to production now.

danielmewes:
Thanks again for your help in getting this into shape @BanzaiMan !

hueniverse:
Thanks @BanzaiMan @danielmewes! Can't wait to try it out.

PR #677, opened 2016-03-10
Fold Julia installation log

simonbyrne:
This should make it easier to read the logs, especially on OS X where hdiutil spits out a bunch of noise.


simonbyrne:
cc @tkelman @ninjin @staticfloat

tkelman:
lgtm, didn't see this. weren't we going to have you take over as co maintainer in place of either Pontus or Elliot?

simonbyrne:
I'd be happy to: is there anything I need to do?

tkelman:
Add your name to the announce and docs, I think.

PR #697, opened 2016-04-12
expose event_type as environment variable 

Nef10:
This is helpful to detect a cron job (see travis-ci/travis-api#225)

Unfortunately I do not know how this could be tested easily, maybe we could look into this next Friday?


broady:
Weird. I don't see this env variable coming through at all.

https://travis-ci.org/broady/travisplay

`TRAVIS_EVENT_TYPE=`

BanzaiMan:
I don't think it is included in the build payload yet.

Nef10:
does work now and is deployed, see: https://travis-ci.org/HPI-BP2015H/test/builds/123265374

PR #700, opened 2016-04-14
allow easy using of other smalltalkCI versions

Nef10:
This adds easy to use support for testing against edge versions of smalltalkCI. It uses the same format as the [travis dpl](https://github.com/travis-ci/dpl/blob/master/TESTING.md):
```yaml
smalltalk_edge:
  source: HPI-BP2015H/smalltalkCI
  branch: dev
```
/cc @fniephaus 
fix hpi-swa/smalltalkCI#124

### DIFF
--- a/lib/travis/build/script/smalltalk.rb
+++ b/lib/travis/build/script/smalltalk.rb
@@ -60,7 +60,7 @@ module Travis
           sh.cmd "pushd $HOME > /dev/null", echo: false
           sh.fold 'download_smalltalkci' do
             sh.echo 'Downloading and extracting smalltalkCI', ansi: :yellow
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/smalltalk.rb
```diff
@@ -76,6 +76,22 @@ def script
 
         private
 
+          def smalltalk_ci_repo
+            if config.key?(:smalltalk_edge) and config[:smalltalk_edge].key?(:source)
+              return config[:smalltalk_edge][:source]
```
fniephaus: I wonder if this one-liner for `smalltalk_ci_repo ` is better (not sure if this is something people prefer in Ruby):
```ruby
return config.fetch(:smalltalk_edge, {}).fetch(:source, "hpi-swa/smalltalkCI")
```

Nef10:
@BanzaiMan could you have a look?

PR #701, opened 2016-04-15
Adapt sbt-extras for older versions of sbt

BanzaiMan:
Since 'addSbt warn' and 'addSbt info' are not defined in older
versions of sbt, we remove them.

See https://github.com/paulp/sbt-extras/issues/99


dwijnand:
Note that as of https://github.com/paulp/sbt-extras/pull/160 the problematic `addSbt` has been removed.

PR #709, opened 2016-04-27
Run `rvm use` at least once before running Ruby with RVM

BanzaiMan:
Fixes https://github.com/travis-ci/travis-ci/issues/5929

It is not entirely clear why, but this appears to remove the
unsightly warnings about `$PATH` on the Trusty image running on
non-Ruby jobs.


BanzaiMan:
Before: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/492895
After: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/492925

PR #712, opened 2016-04-28
Modify nvm install behavior when 'nvm install' fails

BanzaiMan:
Fixes https://github.com/travis-ci/travis-ci/issues/5333

With this change:

1. Install the latest version known to nvm if nodejs.org is reachable
2. If nodejs.org is not reachable or 'nvm install' fails for some reason,
   use the locally available and applicable version
3. If there is no version available locally, fail the build immediately


BanzaiMan:
Test: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/492969

Note that this modified `/etc/hosts` to simulate the situation where nodejs.org is unreachable. This was done for demonstration purposes only. This PR does not do such a thing.

PR #718, opened 2016-05-02
Use stable version of composer by default

Soullivaneuh:


BanzaiMan:
Could you explain the motivation for this PR?

Soullivaneuh:
@BanzaiMan Very simple: Composer has now stable versions to provides.

So using stable versions of composer will produce more reliable builds.

Soullivaneuh:
Build is working. Is something missing to have this PR merged?

GrahamCampbell:
@BanzaiMan I don't think this PR actually works. The version of composer travis has installed by default doesn't accept the stable flag since it was only added in a later version.

GrahamCampbell:
The actual fix for now is simply to actually run `composer self-update` twice.

Soullivaneuh:
> @BanzaiMan I don't think this PR actually works. The version of composer travis has installed by default doesn't accept the stable flag since it was only added in a later version.

Well, why not simply install a more recent version of composer in this case?

GrahamCampbell:
> Well, why not simply install a more recent version of composer in this case?

Then we wouldn't need to update it though. ;P

Given we're stuck on php versions from 2014 (other than php 7), it would be good to get a new version of php 5.5 and 5.6 on here and all.

Soullivaneuh:
> Then we wouldn't need to update it though. ;P

Well I just talk for this time, in order to have the `--stable` option working.

> Given we're stuck on php versions from 2014 or something (other than php 7), it would be good to get a new version of php 5.5 and 5.6 on here and all.

:+1:

Soullivaneuh:
@BanzaiMan any new?

Concrete example with a PHP 5.3 job:

By default, the composer is updated to an unstable version: https://travis-ci.org/sonata-project/exporter/jobs/138338244#L139

So we have to re-update it for a stable, making the previous command useless: https://travis-ci.org/sonata-project/exporter/jobs/138338244#L202

Adding the `--stable` option will permit to get rid of this nasty hack.

cc @Seldaek

GrahamCampbell:
@Soullivaneuh This is only happening because the version on travis is too old to know about stability even existing.

Soullivaneuh:
> @Soullivaneuh This is only happening because the version on travis is too old to know about stability even existing.

You mean the `--stable` option? In this case just update the original composer on those build first.

If I can do it myself on a PR, I'll be glad to do it @BanzaiMan.

BanzaiMan:
@Soullivaneuh We'd have to wait until the next build environment update.

Soullivaneuh:
Any new here? Thanks.

BanzaiMan:
Hello. Sorry about the delay. I was under the impression that there was no real consensus here to act on.

Regardless, I think this PR is superseded by https://github.com/travis-ci/travis-build/pull/923 at this time.

Seldaek:
It is indeed. Both issues could be closed if the base image would only include a composer build that's >1.0.0. Fixing this doesn't sound extremely hard but I don't know how regularly you rebuild images.

Soullivaneuh:
This is indeed fixed thanks to this: https://github.com/travis-ci/travis-build/blob/e5a45cbf49e0d9e27398e76e5f25dd7706feb6aa/lib/travis/build/script/php.rb#L182

PR #720, opened 2016-05-02
Install virutalenv on demand if it is not available

BanzaiMan:
This allows the Python builds to make progress on all images, including the Mac.


BanzaiMan:
`language: python` + `os: osx` jobs need to be routed to OS X in order to properly test this.

PR #724, opened 2016-05-04
Abort builds if Sauce Connect blocks for > 2 min.

DylanLacey:
Builds opening a Sauce Connect tunnel currently block until Sauce
Connect has produced a readyfile.  In some cases there is no readyfile
producted; This blocks the build until the global no-output timer (10
minutes as of this commit) terminates it.

Readyfiles are produced once a Sauce Connect tunnel is fully up.  This
can fail to happen for a variety of reasons, including the user having
too many tunnels open at once.

This commit introduces a 2 minute wait time (as advised by @ldx, Sauce
Labs employee and total Sauce Connect smart guy) for the readyfile,
after which the entire build is terminated.

The `travis_jigger` function is used to post a notification to the
console and then ultimately kill the process via `travis_terminate 2`
(similarly to `travis_assert`).  An earlier revision used `travis_wait`
but this produced confusing output.

This commit closes https://github.com/travis-ci/travis-ci/issues/3091.

### DIFF
--- a/lib/travis/build/addons/sauce_connect/templates/sauce_connect.sh
+++ b/lib/travis/build/addons/sauce_connect/templates/sauce_connect.sh
@@ -2,6 +2,15 @@
 
 export _SC_PID=unset
 
+function wait_for_sauce_connect_readyfile() {
+  readyfile=$1
+  echo "Waiting for Sauce Connect readyfile"
+
+  while [ ! -f ${readyfile} ]; do
+    sleep .5
+  done
+}
+
 function travis_start_sauce_connect() {
   if [ -z "${SAUCE_USERNAME}" ] || [ -z "${SAUCE_ACCESS_KEY}" ]; then
       echo "This script can't run without your Sauce credentials"
@@ -13,6 +22,8 @@ function travis_start_sauce_connect() {
 
   local sc_tmp sc_platform sc_distro sc_distro_fmt sc_distro_shasum \
     sc_readyfile sc_logfile sc_dir sc_tunnel_id_arg sc_actual_shasum
+
+  local timeout=2
 
   sc_tmp="$(mktemp -d -t sc.XXXX)"
   echo "Using temp dir $sc_tmp"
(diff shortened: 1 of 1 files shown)


BanzaiMan:
[This simple test](https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/493922) shows that it works (the timeout of 2 minutes is respected, since the build has incorrect credentials). Ideally, there is a better way to detect that the credentials are invalid and the tunnel is never going to come up.

bootstraponline:
> Ideally, there is a better way to detect that the credentials are invalid

Make [a REST API call](https://wiki.saucelabs.com/display/DOCS/The+Sauce+Labs+REST+API) to Sauce?

DanielHeath:
Almost none of the variable interpolations in this file are quoted, so if there are space characters anywhere you'll be in trouble. 

Compare:
`while [ ! -f ${readyfile} ]; do` 
`while [ ! -f "${readyfile}" ]; do`

If `$readyfile` is set to `missingfile || echo 'hacked you'`, the former will print 'hacked you' to the console while the latter will not.

DylanLacey:
Making a REST API call is a good approach for credentials, but the problem I was trying to solve was the generalised problem of Sauce Connect start up problems.  This can happen for a variety of reasons, some detectable in advance (assuming nothing changes between checks and actual start up, which takes several seconds) and some not.

Currently, an issue starting Sauce Connect will cause a build to hang until Travis's hung build timeout is reached, at which point the entire build is terminated.  This change should account for _most_ problems, known or otherwise, and end builds in 2 minutes, 5 times faster then the default.

@DanielHeath good call; I'll have a whack at correcting that and send an updated PR.

DylanLacey:
Globbing updated;  I left it out for arguments that can't have spaces.

BanzaiMan:
OK. Let's test this again tomorrow (July 15).

tristanbes:
what's the status of this PR please ?

meatballhat:
I'm sorry for the long delay!  This PR is no longer applicable.  Please holler if you feel it should be reopened, assuming the problems haven't been addressed elsewhere.

PR #727, opened 2016-05-06
Special case tip.

paddyforan:
When using the tip version of Go, we want to always consider it as after
1.5 (and, more importantly, 1.2). Previously (before
travis-ci/travis-build#658) this was fine, because 'tip' > '1.2'
evaluates true in Ruby. After the patch, we were using true comparisons
instead of string comparisons, and suddenly it was defaulting to
'0.0.1', which means 'tip' > '1.2' effectively starting evaluating
false. Which is not what we want. So now all our comparisons should be
special-casing tip.

### DIFF
--- a/lib/travis/build/script/go.rb
+++ b/lib/travis/build/script/go.rb
@@ -110,10 +110,10 @@ module Travis
           
           # see https://golang.org/doc/go1.6#go_command
           def uses_15_vendoring
-            if (go_version == 'go1' || comparable_go_version < Gem::Version.new('1.5'))
+            if (go_version == 'go1' || (go_version != 'tip' && comparable_go_version < Gem::Version.new('1.5')))
               return 'false'
             end
-            (comparable_go_version < Gem::Version.new('1.6')) ? '$GO15VENDOREXPERIMENT == 1' : '$GO15VENDOREXPERIMENT != 0'
+            (comparable_go_version < Gem::Version.new('1.6') && go_version != 'tip') ? '$GO15VENDOREXPERIMENT == 1' : '$GO15VENDOREXPERIMENT != 0'
           end
 
           def gobuild_args
@@ -136,6 +136,7 @@ module Travis
             when '1.0' then '1.0.3'
             when '1.2' then '1.2.2'
             when 'go1' then v
+            when 'tip' then v
             when /^go/ then v.sub(/^go/, '')
             else v
             end
@@ -149,7 +150,7 @@ module Travis
           end
 
           def go_get_cmd
-            (go_version == 'go1' || comparable_go_version < Gem::Version.new('1.2')) ? 'go get' : 'go get -t'
+            (go_version == 'go1' || (go_version != 'tip' && comparable_go_version < Gem::Version.new('1.2'))) ? 'go get' : 'go get -t'
           end
 
           def ensure_gvm_wiped


lib/travis/build/script/go.rb
```diff
@@ -110,10 +110,10 @@ def uses_make
           
           # see https://golang.org/doc/go1.6#go_command
           def uses_15_vendoring
```
BanzaiMan: I overlooked this in the previous PR, so I apologize for saying this late.

This method should be renamed to `uses_15_vendoring?` to make the intention clearer.

lib/travis/build/script/go.rb
```diff
@@ -149,7 +150,7 @@ def comparable_go_version
           end
 
           def go_get_cmd
-            (go_version == 'go1' || comparable_go_version < Gem::Version.new('1.2')) ? 'go get' : 'go get -t'
+            (go_version == 'go1' || (go_version != 'tip' && comparable_go_version < Gem::Version.new('1.2'))) ? 'go get' : 'go get -t'
```
BanzaiMan: The logic is getting a bit too complicated for a ternary operator. Please rework this into a more simple `if` statement.

lib/travis/build/script/go.rb
```diff
@@ -110,10 +110,10 @@ def uses_make
           
           # see https://golang.org/doc/go1.6#go_command
           def uses_15_vendoring
-            if (go_version == 'go1' || comparable_go_version < Gem::Version.new('1.5'))
+            if (go_version == 'go1' || (go_version != 'tip' && comparable_go_version < Gem::Version.new('1.5')))
```
BanzaiMan: Instead of `go_version != 'tip'`, how about having a method with more descriptive name, say `tip?` or `version_is_tip?`?

BanzaiMan: Also, `go_version != 'tip' && comparable_go_version < Gem::Version.new('1.5')` appears a few times, and I think it should also be extracted into a method.

lib/travis/build/script/go.rb
BanzaiMan: If you extract `go_version != 'tip' && comparable_go_version < Gem::Version.new('1.5')` into a method, this is not too bad.

BanzaiMan:
This needs specs.

lib/travis/build/script/go.rb
paddyforan: Oh, thanks! I'm, uh (spoilers), not really a Rubyist, so I tend to miss that kind of convention. I'll address that in this pass.

lib/travis/build/script/go.rb
paddyforan: After a bit of pondering on this, I think I've come to the conclusion that using helpers to group versions is both hard to understand ("Why are they asking about specific versions here?") and is just going to make things worse over time, as more versions get special treatment.

I think the best way to handle this is to provide, like I did for vendoring, a special helper that answers the question at hand. In this case, a `can_get_test_dependencies?` helper that returns `true` if there's a `-t` flag to use with `go get` to retrieve dependencies for tests. Then this becomes:

```ruby
(can_get_test_dependencies?) ? 'go get -t' : 'go get'
```

Which I think is much more readable and clearer in what it's actually asking. What do you think?

lib/travis/build/script/go.rb
paddyforan: Personally, `go_version == 'tip'` and `go_version.is_tip?` are equally clear to me. And `go_version == 'go1'` appears just as many times, but has no helper.

Three things about this:

1. You're going to be maintaining this long after I'm gone, so I'm happy to use `is_tip?` or whatever, if you'd prefer.
2. If you want me to use `is_tip?`, should I also refactor at the same time to use `is_go1?` instead of `go_version == 'go1'`, for consistency?
3. Why are we keeping `'go1'` around, anyways? `'tip'` I understand, as it has no semver version associated with it. But is there any harm in just setting `go1` to be `1.0.3` or whatever in `comparable_go_version` and removing all the logic around it?

paddyforan:
> This needs specs.

Is there any guide or doc for writing specs for travis-build, or should I try to crib off of other specs? I'm sorry, I'm not a Rubyist by nature, so I'm venturing into unusual territory here.

lib/travis/build/script/go.rb
BanzaiMan: On a second thought, this is not quite save us much, since the two place `#comparable_go_version` appears are not quite equal; one compares with `1.2`, the other with `1.5`.

BanzaiMan:
I've supplemented this PR with my own improvements.

PR #739, opened 2016-05-27
R - Dump fail logs first and use separate folds for each type

jimhester:
Generally the fail logs are the most useful, previous behavior dumped
them last, which caused them to be truncated if there was a log of build
output.

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -366,19 +366,23 @@ module Travis
         end
 
         def dump_logs
-          sh.fold "Check logs" do
-            export_rcheck_dir
-            sh.echo 'R CMD check logs', ansi: :yellow
-            ['out', 'log', 'fail'].each do |ext|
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/r.rb
```diff
@@ -366,19 +366,23 @@ def export_rcheck_dir
...
-              'do '\
-                'echo ">>> Filename: ${name} <<<";'\
-                'cat ${name};'\
-                'done'
-              sh.cmd cmd
-            end
+          export_rcheck_dir
+          dump_log("fail")
+          dump_log("log")
+          dump_log("log")
```
BanzaiMan: Is running `dump_log("log")` twice intentional?

jimhester: No, thanks for the catch, I will fix it.

jimhester:
This is now updated, sorry for the delay!

PR #740, opened 2016-06-01
Use Sauce Connect 4.3.16

ldx:


henryprecheur:
Number increased correctly & sha1 correctly copied.

:ship:

PR #743, opened 2016-06-07
Use my fork instead of yihui's

jimhester:
This will fix https://github.com/travis-ci/travis-ci/issues/6139 once it is deployed. All R builds are broken until then so it is high priority.

cc @BanzaiMan


bestdan:
Thanks @jimhester !

PR #744, opened 2016-06-09
Optionally turn off appliances with configuration

BanzaiMan:
Under some circumstances, it is desirable to skip running appliances;
e.g., for Travis Enterprise, running update_glibc appliances
can be considered unnecessary, because the builds run inside a
firewall.


BanzaiMan:
After some dicsussion, we decided to make this configuration opt-in, so that Enterprise customers won't have to do anything to skip this addon.

PR #749, opened 2016-06-21
Install Rebar as well as Hex in Elixir builds

zackehh:
Use case: https://travis-ci.org/zackehh/siphash-elixir/jobs/139247149

Right now if an Elixir dependency requires Rebar, a build will simply hang waiting for input. I think it would be better to have the build script install this the same way it does Hex, rather than having users need to know to add it to their install setups.

### DIFF
--- a/lib/travis/build/script/elixir.rb
+++ b/lib/travis/build/script/elixir.rb
@@ -44,6 +44,7 @@ export MIX_ARCHIVES=#{KIEX_MIX_HOME}elixir-#{elixir_version}' > #{KIEX_ELIXIR_HO
 
         def install
           sh.cmd 'mix local.hex --force', fold: "install.hex"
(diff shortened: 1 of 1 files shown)


BanzaiMan:
According to https://github.com/travis-ci/travis-ci/issues/6130, this is required only for Elixir 1.3.0 and up. We should avoid installing rebar unless it is necessary. See https://github.com/travis-ci/travis-build/commit/956931942c0ba227a449cceb48e1c259f958f68f for a possible (untested) solution.

BanzaiMan:
(I can create a PR off that commit and deploy it, but I think this PR can be shaped into something good.)

zackehh:
@BanzaiMan ah I didn't notice the issue filed for it already :) your PR looks like the best solution I think!

BanzaiMan:
@zackehh :-) I'm closing this in favor of #750, then.

PR #754, opened 2016-06-24
Switch Crystal repos to HTTPS

jhass:
Also fetch the repo key from a known location over HTTPS, to workaround https://github.com/travis-ci/travis-ci/issues/5461 hopefully (we'll have to see whether it causes other issues in the wild or not).

/cc @asterite, @waj and @will


will:
I haven't ran it to double check, but reading through this looks good

miketheman:
I had a failure from this change - and didn't know why:
https://travis-ci.org/miketheman/statsd.cr/builds/140672242

The target repo, being cloudfront, uses SNI, and [this issue](https://github.com/travis-ci/travis-ci/issues/5059) reports that the Precise version of wget is too old to support this.

I ended up updating my builder to run on a newer Trusty infra, like so: https://github.com/miketheman/statsd.cr/commit/a8814d1a06ba5229f64522dd2cd50dc58eeec8b9

This might be mitigated by using `curl` instead of `wget`.

jhass:
See #764

PR #755, opened 2016-06-26
Fix sexp matcher bug

BanzaiMan:
This PR does the following to the specs concerning deployment conditions

1. Fix `match_sexp` matcher. The intention of this matcher is to return success if the result returned by `sexp_find` is not empty, but we were only testing if it was a truthy value. Since `[]` is, this matcher always succeeded.
1. Fix incorrect specs. The issue above resulted in many specs that were giving incorrect results; specs had fallen out of sync with the code, but they were not manifest in previous builds.
1. Tweak the `sexp_find` helper. The change makes its recursive nature more readily apparent.

Huge props to @joshcheek, who paired with me and worked out some confusing parts of the code.

### DIFF
--- a/lib/travis/build/addons/deploy/script.rb
+++ b/lib/travis/build/addons/deploy/script.rb
@@ -101,8 +101,11 @@ module Travis
 
             def branch_condition
               return if on[:all_branches] || on[:tags]
-              branches  = Array(on[:branch] || default_branches)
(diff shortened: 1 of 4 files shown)


svenfuchs:
@BanzaiMan @joshcheek amazing, this thing never worked?

Your PR looks good to me. 

LGTM!

❤️

PR #757, opened 2016-06-26
Add support for maven wrapper

britter:
The [maven wrapper project](https://github.com/takari/maven-wrapper) adds a tiny shell script wrapper to maven builds, much like gradlew does for gradle builds. This way users can choose which maven version should be used by specifying it in a special properties files. Maven wrapper will then download the specified maven version to the m2 repository and use it for building the project. The mvnw wrapper script has the same interface as mvn itself.

This PR adds mvnw detection to JVM builds.

### DIFF
--- a/lib/travis/build/script/shared/jvm.rb
+++ b/lib/travis/build/script/shared/jvm.rb
@@ -27,6 +27,9 @@ module Travis
           sh.elif '-f build.gradle' do
             sh.cmd 'gradle assemble', retry: true, fold: 'install'
           end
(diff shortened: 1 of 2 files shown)


BanzaiMan:
Looks good. Could you also document this in https://github.com/travis-ci/docs-travis-ci-com? (https://github.com/travis-ci/docs-travis-ci-com/blob/gh-pages/user/languages/java.md)

britter:
@BanzaiMan done, see travis-ci/docs-travis-ci-com#661

PR #759, opened 2016-06-27
Fix Crystal builds

Sija:
We need to use double-quoted string since it contains expansion.


Sija:
Thanks! When is it goin' to be live @ travis-ci.org?

BanzaiMan:
It should be live now.

PR #764, opened 2016-06-28
Crystal: avoid using wget

jhass:
The version on Travis has TLS support that was nice in the 90s

/cc @will @asterite @waj

### DIFF
--- a/lib/travis/build/script/crystal.rb
+++ b/lib/travis/build/script/crystal.rb
@@ -12,7 +12,7 @@ module Travis
             version = select_version
             return unless version
 
-            sh.cmd %Q(wget '#{version[:key][:url]}' -O "$HOME/crystal_repository_key.asc")
+            sh.cmd %Q(curl -sSL '#{version[:key][:url]}' > "$HOME/crystal_repository_key.asc")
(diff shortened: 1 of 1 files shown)


asterite:
I think @waj reverted requirement of https for some things, so this might not be needed anymore.

@waj ?

will:
Even if the requirement is removed, and if all else is equal, it's probably better to use tls.

waj:
I reverted the redirection from HTTP to HTTPS, but still the latest should be used whenever possible.

The builds are failing after c31b27125a06d75dbd21bbaa25f304aae1f72f2e because the repository is hosted in CloudFront and it doesn't support SSLv3 (https://aws.amazon.com/security/security-bulletins/CVE-2014-3566-advisory/)

@jhass are you sure this will fix the issue with the builds? or it depends on the installed OpenSSL version?

jhass:
I didn't get a chroot to test, but my experience is that curl works in these cases while wget doesn't.

asterite:
Is there something we can do to push this forward? All Crystal projects using travis are basically failing because of this :-(

asterite:
@BanzaiMan thank you so much for this! ❤️

PR #765, opened 2016-06-28
add option to remove preinstalled homebrew

jeroenooms:
In R we need to explicitly test if a package can build on a vanilla OSX, without homebrew. Sometimes having `brew` installed has side effects on the installation, even if no particular brew packages are installed.

This PR adds the option to remove the preinstalled homebrew installation using the official uninstall script provided by the homebrew FAQ.

@jimhester @hadley @craigcitro

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -31,6 +31,7 @@ module Travis
           bioc: 'https://bioconductor.org/biocLite.R',
           bioc_required: false,
           bioc_use_devel: false,
+          disable_homebrew: false,
           r: 'release'
         }
 
@@ -157,6 +158,9 @@ module Travis
 
               setup_bioc if needs_bioc?
               setup_pandoc if config[:pandoc]
+
+              # Removes preinstalled homebrew
+              disable_homebrew if config[:disable_homebrew]
             end
           end
         end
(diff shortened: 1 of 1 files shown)


jeroenooms:
Corresponding documentation PR: https://github.com/travis-ci/docs-travis-ci-com/pull/646

jimhester:
It is too bad homebrew does not bundle the uninstall script when you install homebrew, but that is a homebrew problem, not ours.

otherwise LGTM

lib/travis/build/script/r.rb
```diff
@@ -483,6 +487,14 @@ def setup_pandoc
           end
         end
 
+        # Uninstalls the preinstalled homebrew
+        # See FAQ: https://github.com/Homebrew/brew/blob/master/share/doc/homebrew/FAQ.md
+        def disable_homebrew
```
BanzaiMan: You'd probably need to check if you're running on OS X.

jeroenooms:
@BanzaiMan thanks! I added a check that we're running OS X.

lib/travis/build/script/r.rb
jeroenooms: OK done below.

jeroenooms:
Ping? Looks good to merge?

PR #766, opened 2016-06-28
Add DisableInitramfs appliance

BanzaiMan:
This appliance disables updating initramfs.
Some packages hooks to initramfs updating process, which is useless in
our case, since we will not be using the updated initramfs.

This saves a few minutes of build time in some cases.
See https://github.com/travis-ci/travis-ci/issues/6219


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/505870#L815

PR #768, opened 2016-06-28
npm_registry key in yaml that is then written to NPM_CONFIG_REGISTRY env var

igorwwwwwwwwwwwwwwwwwwww:


igorwwwwwwwwwwwwwwwwwwww:
Ping @BanzaiMan, could you take a look? Do any other places need to be changed for this to work? [travis-yaml](https://github.com/travis-ci/travis-yaml)? Also, would you like to have tests for this patch -- and if so can you give me a hint for how to go about it?

igorwwwwwwwwwwwwwwwwwwww:
oh, and is there a good way to do echo escaping?

BanzaiMan:
Where do we mention `npm_registry` key? I'd never heard of this option.

BanzaiMan:
In general, appliances are for things that we configure and enforce. Addons are for user-defined configurations.

joshk:
I think the use of appliance is correct then.

BanzaiMan:
@joshk I don't think so. This will make whatever users put in `npm_registry` key value available to PRs, even when it is encrypted.

BanzaiMan:
Oh, that's not true. Disregard that.

igorwwwwwwwwwwwwwwwwwwww:
To give you some more context, the behaviour of this will likely change over time. I wanted to start with a yml key, then move to user/org, then % of traffic, then globally.

And also, once we have a more solid setup, it will become a boolean flag as opposed to passing the actual repository URL -- the repository URL will be hard-coded or set as an env var.

igorwwwwwwwwwwwwwwwwwwww:
Updated with comments as per @joshk. Once this is merged we can put a feature toggle into scheduler along the lines of `cache_settings`: https://github.com/travis-ci/travis-scheduler/blob/master/lib/travis/scheduler/payloads/worker.rb#L46

lib/travis/build/appliances/npm_registry.rb
```diff
@@ -0,0 +1,17 @@
+require 'travis/build/appliances/base'
+
+module Travis
+  module Build
+    module Appliances
+      class NpmRegistry < Base
+        def apply
+          if data[:npm_registry]
```
BanzaiMan: This logic should be moved to `apply?`.

BanzaiMan:
Needs specs.

igorwwwwwwwwwwwwwwwwwwww:
@BanzaiMan updated with `apply?` and specs!

PR #773, opened 2016-06-30
bundle update deps

igorwwwwwwwwwwwwwwwwwwww:


Gemfile
```diff
@@ -7,12 +7,12 @@ gem 'addressable', '~> 2.3'
 gem 'coder'
 gem 'jemalloc'
 gem 'metriks', '0.9.9.6'
-gem 'metriks-librato_metrics', github: 'eric/metriks-librato_metrics'
+gem 'metriks-librato_metrics', git: 'https://github.com/eric/metriks-librato_metrics'
```
GrahamCampbell: Why does this need to change?

igorwwwwwwwwwwwwwwwwwwww: The reason I made this change is that https provides transport security, whereas the git protocol does not.

GrahamCampbell: The git protocol does use transport security?

GrahamCampbell: https://git-scm.com/book/ch4-1.html#The-Git-Protocol

igorwwwwwwwwwwwwwwwwwwww: No, the git protocol is unauthenticated. It does not use SSH, and it does not use TLS. Here is a wireshark capture showing the plaintext information on the wire, from running `git clone git://github.com/travis-ci/travis-support.git`:

screen shot 2016-06-30 at 18 18 02

GrahamCampbell: What about `git clone git@github.com:travis-ci/travis-support`?

igorwwwwwwwwwwwwwwwwwwww: That's a clone over the SSH protocol -- which would also work as an alternative to `https`. However, that requires SSH to be setup and associated with a GitHub account.

PR #777, opened 2016-07-06
Add feature to temporarily pin hostnames to IP addresses

henrikhodne:
This is mainly useful as temporary workarounds if a certain mirror is
broken, or something similar.


henrikhodne:
Tested in staging, seems to work fine on all infras.

PR #778, opened 2016-07-06
Define BUNDLE_GEMFILE in all cases

BanzaiMan:
Even when the file does not exist.

This allows `before_script` to generate the file, and `install` to
use the generated file.


BanzaiMan:
https://github.com/search?p=5&q=%22BUNDLE_GEMFILE%22+filename%3A.travis.yml&type=Code&utf8=%E2%9C%93 returns too many hits to tweak this logic at this moment.

PR #779, opened 2016-07-07
PyPy on demand

BanzaiMan:
This PR allows installation of PyPy (both 2 and 3).

With this, users can specify PyPy versions in this way:

```yaml
language: python

python:
  - pypy2-5.3.1
  - pypy3.3-5.2.0-alpha1 # with optional trailing "-[a-z0-9]*"
```

### DIFF
--- a/lib/travis/build/script/python.rb
+++ b/lib/travis/build/script/python.rb
@@ -13,7 +13,7 @@ module Travis
         PYENV_PATH_FILE      = '/etc/profile.d/pyenv.sh'
         TEMP_PYENV_PATH_FILE = '/tmp/pyenv.sh'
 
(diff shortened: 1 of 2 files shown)


alex:
I assume teh reason not to implement this using pyenv is that it attempts to build from source which is super slow?

BanzaiMan:
@alex Frankly, I did not think about that. :-D

We could adopt our [cpython-builder](https://github.com/travis-ci/cpython-builder) to build PyPy archives; maybe it is a good idea to do so, since we need to install extra packages (such as numpy) anyway.

BanzaiMan:
Wheee. `pypy3.3-5.2.0-alpha1` fails to install `setuptools`. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/506157#L130

BanzaiMan:
Closing this in favor of leveraging cpython-builder.

PR #780, opened 2016-07-08
[csharp] Add preliminary .NET Core/CLI support

akoeplinger:
~~Only supports Linux/trusty for now. OSX coming in a future revision.~~ Should work on OSX now too, but hardcodes a .NET Core version.


akoeplinger:
@BanzaiMan A staging deploy of this would be great 😄

BanzaiMan:

PR #784, opened 2016-07-14
Capture gimme output and use that to activate

BanzaiMan:
When there is nothing in STDOUT from $(gimme X), we would execute
`eval ''`, which returns success, and the build continues.

Instead, we capture the output to an environment variable, so that
the failure in $() is transparent to the build script.

Resolves https://github.com/travis-ci/travis-ci/issues/6299

### DIFF
--- a/lib/travis/build/script/go.rb
+++ b/lib/travis/build/script/go.rb
@@ -41,7 +41,7 @@ module Travis
         end
 
         def setup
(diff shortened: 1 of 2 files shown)


BanzaiMan:
Tested in staging: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/506675

PR #785, opened 2016-07-15
Echo build and job id as part of system info at beginning of build

igorwwwwwwwwwwwwwwwwwwww:


joshk:
Hey @igorwwwwwwwwwwwwwwwwwwww, could you give some context on why you think this would be good to add?

igorwwwwwwwwwwwwwwwwwwww:
Totes. For single job builds it's very hard to find the build id, which is needed to start a debug build via the API. Instead of having to go DOM diving (what support currently suggests) having it in the logs would be quite helpful.

joshk:
Ok, sounds fair.

Just to note, this may be removed in the future once we add a proper `debug job` button, and some tweaks to the UI which may include this information.

igorwwwwwwwwwwwwwwwwwwww:
@BanzaiMan could you take a look? :)

PR #786, opened 2016-07-15
Add support for non-AWS S3-compatible caching backend

BanzaiMan:
This PR allows users to specify S3-compatible (with V4 signature) caching backend besides AWS S3.

The scheduler's configuration would look like this:

```yaml
cache_settings:
  builds.docker:
    type: s3
    s3:
      access_key_id: KEY_ID
      secret_access_key: SECRET_ACCESS_KEY
      hostname: MINIO_SERVER_HOSTNAME
      bucket: BUCKET_NAME
      scheme: http
      bucket_name_in_path: true
      aws_signature_version: '4'
    fetch_timeout: 1200
    push_timeout: 7200
```

I have tested this with [Minio](https://minio.io).

Note that `bucket_name_in_path: true` is now the default, so for our services, `bucket_name_in_path: false` should be defined.

### DIFF
--- a/lib/travis/build/script/shared/directory_cache/base.rb
+++ b/lib/travis/build/script/shared/directory_cache/base.rb
@@ -41,9 +41,13 @@ module Travis
 
           KeyPair = Struct.new(:id, :secret)
 
-          Location = Struct.new(:scheme, :region, :bucket, :path, :host) do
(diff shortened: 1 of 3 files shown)


BanzaiMan:
Since the S3 buckets can be [accessed via path](http://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html#access-bucket-intro), we should look into standardizing the access method to it, so that the code can be cleaned up.

PR #800, opened 2016-07-25
Remove correct directory when edge.source is set by env var

BanzaiMan:
Resolves https://github.com/travis-ci/dpl/issues/481


BanzaiMan:
Tested on staging: https://staging.travis-ci.org/BanzaiMan/dpl/builds/507013

PR #801, opened 2016-07-25
OS X R devel

jimhester:
Should fix https://github.com/travis-ci/travis-ci/issues/6268


BanzaiMan:
@jimhester Looks like we need the specs updated.

jimhester:
Sorry, should be ready now!

PR #804, opened 2016-07-27
Use https to clone repo when source URL starts with `https://` (as opposed to `git@`)

BanzaiMan:
Ordinarily, we prefer GIT+SSH (e.g., `git@github.com:OWNER/REPO.git`)
to HTTPS (e.g., `https://github.com/OWNER/REPO.git`) to clone repo.

In some cases, such as Travis Enterprise, the SSH port may be closed
as a matter of security policy.

If the `source_url` on the payload starts with `https` (which is true for .org) and the payload contains GitHub OAuth token, we write credentials to `~/.netrc` and rewrite the clone URL to HTTPS.


BanzaiMan:
See https://github.com/travis-ci/travis-scheduler/pull/53.

We will look at the git source URL scheme to decide whether to write to `~/.netrc`.

BanzaiMan:
https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/539009 shows that on .org, we do not use `~/.netrc` to clone.

lib/travis/build/git/clone.rb
```diff
@@ -72,6 +75,18 @@ def config
             data.config
           end
 
+          def write_netrc
+            sh.newline
+            sh.echo "Using $HOME/.netrc to clone repository.", ansi: :yellow
+            sh.newline
+            sh.raw "echo -e \"machine github.com\n  login #{data.token}\\n\" > $HOME/.netrc"
```
bnferguson: Does the `machine github.com` need to be customized for the GitHub install? For instance should this read 

```
machine ghe.volcanicislandfortress.com
```

for my GitHub Enterprise install?

BanzaiMan: A good catch! I'll look into this later this week.

PR #809, opened 2016-07-29
Move JWT addon to run before before_install

halkeye:
https://github.com/travis-ci/travis-ci/issues/6383

switch to before_install so if people want to use the credentials in installs steps (or before_install steps) it is available

### DIFF
--- a/lib/travis/build/addons/jwt.rb
+++ b/lib/travis/build/addons/jwt.rb
@@ -7,7 +7,7 @@ module Travis
       class Jwt < Base
         SUPER_USER_SAFE = true
 
-        def export
+        def before_before_install
           tokens = {}
           Array(config).each do |secret|
             pull_request = self.data.pull_request ? self.data.pull_request : ""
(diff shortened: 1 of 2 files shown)


BanzaiMan:
This certainly works, but you could also use the `export` stage, where all other environment variables are exported. (And it runs before any user-customizable phase such as `before_install`.)

halkeye:
oh sweet
def export?
committed

BanzaiMan:
Hmm. I might be mistaken. I'll take a deeper look next week. Sorry for the confusion here.

halkeye:
looks like scripts (script.rb) has export

 $ ag 'def export' lib/travis/ -c | awk -F/ '{print $4 "/"}' | sort | uniq
-c
      2 addons/
      1 appliances/
      1 builder.rb:1/
     22 script/
      1 script.rb:1/


The few in addons that have it are calling it themselves.

I've reverted that change (prefer to leave things in a working state).

BanzaiMan:
@655445a has been deployed to staging.

BanzaiMan:
Oh. There is no reason to test in staging, actually. I'll just merge this.

halkeye:
Awesome.

Lemme know when its live so I can run my tests again

BanzaiMan:
@halkeye Deployed to production.

PR #810, opened 2016-08-01
Fix https://github.com/travis-ci/travis-ci/issues/6394

paleo13:
This pull request fixes travis-ci/travis-ci#6394

Specifically, the latest and developmental versions are scraped from web-pages and build logs, and then used to determine which version of R should be installed for checking R packages.


BanzaiMan:
Superseded by #811.

PR #813, opened 2016-08-04
Extract errexit logic into methods

BanzaiMan:
This PR continues work of #812.

Ensure that caching and deployment occurs with `errexit` option off.

Some builds `source` bash files that set `errexit` (e.g., `set -e`). When our caching and deployment commands fail, this option leads to immediate job termination, which can be very confusing.

An example build is https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/507340.

On https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/507340#L166, caching command fails, and on https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/507340#L196 the deployment (which is just calling `false`) fails. In both of those cases, the job continues to completion.

### DIFF
--- a/lib/travis/build/addons/deploy/script.rb
+++ b/lib/travis/build/addons/deploy/script.rb
@@ -57,7 +57,6 @@ module Travis
           private
             def check_conditions_and_run
               sh.if(conditions) do
-                sh.raw 'set +e'
                 run
               end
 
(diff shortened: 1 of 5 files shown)


BanzaiMan:
Also in production: https://travis-ci.org/BanzaiMan/travis_production_test/builds/149808795

PR #814, opened 2016-08-06
Pass the correct archive filename to rm after Python install

edmorley:
After #781, the Python install step outputs:

> 2.7.11 is not installed; attempting download
> rm: cannot remove `2.7.11.tar.bz2': No such file or directory

...since the correct filename is actually `python-2.7.11.tar.bz2`.

This change factors out the filename to fix the `rm` instance, reduce duplication and to avoid different filenames inadvertently being used with each command in the future.

Fixes travis-ci/travis-ci#6416.


BanzaiMan:
Thank you!

PR #829, opened 2016-08-26
nix: Simplify setup

matthewbauer:
This just executes the file at https://nixos.org/nix/install. This URL
should be trustable. In addition, the build script provided should
cleanup the temporary directory. However, it makes the travis config
less robust because it will have to always use the latest Nix version.

I'm going with this approach mostly because it should be easier to
maintain in the future.

Fixes travis-ci/travis-ci#6520.

### DIFF
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -7,9 +7,7 @@ module Travis
   module Build
     class Script
       class Nix < Script
-        DEFAULTS = {
-          nix_version: '1.11.2'
-        }
+        DEFAULTS = {}
 
         def export
           super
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Sorry about the delay.

BanzaiMan:
Does this mean `nix` will always use the latest version defined in the install script? Is that wise?

matthewbauer:
@BanzaiMan Well it should be the latest stable version. I think it's okay for now. At least until there are some breaking changes to Nix, then I'll try to add some multi-versioning support.

rasendubi:
Does anything block this PR?

domenkozar:
I wonder (sorry for a bit offtopic), could we install nix and generate an image? This way a few seconds would be spared for each build.

rasendubi:
@BanzaiMan ping

PR #833, opened 2016-08-31
Backport 832 for 2.0: Introduce TRAVIS_BUILD_APT_WHITELIST_SKIP env var

bnferguson:
This backports #832.

Conflicting files:
- lib/travis/build/addons/apt.rb
- spec/build/addons/apt_spec.rb

cc @BanzaiMan

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -79,63 +79,27 @@ module Travis
 
             whitelisted = []
             disallowed = []
-<<<<<<< HEAD
 
             config_sources.each do |source_alias|
               source = source_whitelist[source_alias]
-              whitelisted << source.clone if source && source['sourceline']
-              disallowed << source_alias if source.nil?
-||||||| parent of eb5e562... Merge pull request #832 from travis-ci/ha-apt-bypass-whitelist
-            disallowed_while_sudo = []
 
-            config_sources.each do |src|
-              source = source_whitelist[src]
-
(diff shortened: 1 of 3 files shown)


BanzaiMan:
Hmmm. Those are some unhappy conflicts. I'll take a look.

bnferguson:
@BanzaiMan - I reworked the PR (but not the specs yet) to try to capture the intent of the original. Thoughts?

bnferguson:
@BanzaiMan - OK, I think this is ready for review now. I took a different approach. Bit by bit I realized I was rewriting more and more of what was already in master and this addon is pretty nicely isolated. So pulled in the master changes in this file and the spec and things seem to be checking out. What ends up coming in with it is the `paranoid` bits, which I'm not familiar with, otherwise it seems to be only the things we need. 

Mind taking another pass at it review wise? The big thing that I'm wondering now is will I need to pass in anything besides the `TRAVIS_BUILD_APT_WHITELIST_SKIP` env var in the Enterprise config. 

And will there be any issue with the lack of paranoid or `disable_sudo?` coming in on the data payload? (they should both default to nil either way right?)

bnferguson:
@BanzaiMan - I've tested this and it seems to work, if you have a moment today, do you mind giving it a quick once over review wise since the change is much larger now? I've confirmed how it behaves for my first concern up there, but still curious about: 

> And will there be any issue with the lack of paranoid or `disable_sudo?` coming in on the data payload? (they should both default to nil either way right?)

Thank youuuuuu! ⚡

PR #837, opened 2016-09-01
Update dart.rb to sleep after xvfb start

a14n:
give xvfb some time to start

### DIFF
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -98,6 +98,8 @@ module Travis
             if with_content_shell
               sh.export 'DISPLAY', ':99.0'
               sh.cmd 'sh -e /etc/init.d/xvfb start'
+              # give xvfb some time to start
+              sh.cmd 't=0; until (xdpyinfo -display :99 &> /dev/null || test $t -gt 10); do sleep 1; let t=$t+1; done'
               sh.cmd 'pub run test -p vm -p content-shell -p firefox'
             else
               sh.cmd 'pub run test'


a14n:
@kevmoo

a14n:
This follow the snippet on https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI and could reduce issues with content shell.

BanzaiMan:
In general, I think it is better to sleep until certain conditions are met.

Here, we want to actually check that the display is available.

We'd want something like:

```shell
t=0; sh -e /etc/init.d/xvfb start; until (xdpyinfo -display :99 &> /dev/null || test $t > 10); do sleep 1; let t=$t+1; done
```

(We can still merge this, if a constant sleep is sufficient for this use case.)

kevmoo:
@BanzaiMan the bash is strong with you - WOW

a14n:
Thanks for the tip. It could be worth to update https://docs.travis-ci.com/user/gui-and-headless-browsers/#Using-xvfb-to-Run-Tests-That-Require-a-GUI with that.

BTW there seems to be a little typo : `test $t > 10` vs. `test $t -gt 10`

```sh
t=0; sh -e /etc/init.d/xvfb start; until (xdpyinfo -display :99 &> /dev/null || test $t -gt 10); do sleep 1; let t=$t+1; done
```

a14n:
PTAL

BanzaiMan:
Ah, yes. `-gt` is more appropriate for integer comparison. (`>` worked for my simple case, but I suspect it's not great in general.)

kevmoo:
LGTM!

BanzaiMan:
sorry for the delay.

a14n:
thanks :)

PR #840, opened 2016-09-20
Update Sauce Connect to 4.4.0

andyli:
Sauce Connect 4.4.0 should be compatible with 4.3.x according to its release note at https://wiki.saucelabs.com/display/DOCS/2016/09/14/Sauce+Connect+Proxy+Version+4.4.0+Released


BanzaiMan:
I think the checksums needs updating.

andyli:
Indeed! Just updated.

PR #841, opened 2016-09-20
Apply CURL_USER_AGENT to Julia builds as well

staticfloat:
This should add a recognizable user agent to `curl` requests made against our AWS servers, so we can filter CI out from other download types.

This is borrowed from D's implementation [here](https://github.com/travis-ci/travis-build/blob/88d9e0592aa9a96be867e9e0aa3aed757523c160/lib/travis/build/script/d.rb#L27-L28).

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -36,13 +36,14 @@ module Travis
 
           sh.fold 'Julia-install' do
             sh.echo 'Installing Julia', ansi: :yellow
+            sh.cmd 'CURL_USER_AGENT="Travis-CI $(curl --version | head -n 1)"'
             case config[:os]
             when 'linux'
               sh.cmd 'mkdir -p ~/julia'
-              sh.cmd %Q{curl -s -L --retry 7 '#{julia_url}' } \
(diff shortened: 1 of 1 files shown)


staticfloat:
Note that I have basically no way of testing this, so I'm flying blind here.

tkelman:
worth testing on the travis staging instance just to make sure it doesn't break anything across os'es?

tkelman:
@BanzaiMan ping? how much trouble is it to test on staging?

BanzaiMan:
Hello. This has been deployed to staging.

tkelman:
Thanks! What's the best way to opt in to staging on some test repo? It's been a while since the last time I used it.

BanzaiMan:
@tkelman Go to https://staging.travis-ci.org/profile/ and enable the repo you want to test, then push to the repo. When you are done, go back to https://travis-ci.org/profile/ and re-enable the repo there so that the next push will build on production. (Or, you can set up a dedicated repo for staging.)

tkelman:
Thanks for the reminder! I think we're good here, nothing seems to have exploded: https://staging.travis-ci.org/JuliaCI/JSON.jl/builds/513523

PR #843, opened 2016-09-21
Update release version to Julia v0.5

jgreener64:
With the release of Julia v0.5, I think this is how the Travis build should deal with `release`.

Pinging @tkelman, @ninjin, @staticfloat, and @simonbyrne as suggested by my Travis output.

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -100,7 +100,7 @@ module Travis
             case config[:julia].to_s
             when 'release'
               # CHANGEME on new minor releases (once or twice a year)
-              url = "s3.amazonaws.com/julialang/bin/#{osarch}/0.4/julia-0.4-latest-#{ext}"
+              url = "s3.amazonaws.com/julialang/bin/#{osarch}/0.5/julia-0.5-latest-#{ext}"
             when 'nightly'
               url = "s3.amazonaws.com/julianightlies/bin/#{osarch}/julia-latest-#{nightlyext}"
             when /^(\d+\.\d+)\.\d+$/


tkelman:
An annoying consequence of this is that everyone who still supports 0.4 in their `REQUIRE` file but was using `release` instead of `0.4` on Travis won't be testing against 0.4 any more, and might not notice if they accidentally break their package there. I'm planning on writing a script to submit many auto-generated PR's to the packages where this applies.

ninjin:
Looks good.

@tkelman: Is there any way that we can avoid the same issue for 0.6?  My head is a bit hazy due to a cold, but the only way I can see us pulling it off is if Travis supports some sort of "X and above" syntax.  Which, at least last time I checked, they don't.  Since each entry in the build matrix will have to be specified explicitly.

tkelman:
I think the way to fix this going forward is change `PkgDev.generate` to not use `release` by default. Use whatever julia version goes into `REQUIRE` by default, and nightly.

ninjin:
Sounds sensible.  That way each new packaged will be "pinned" to whatever version it was created for, unless the maintainer explicitly wants to test against newer releases.

tkelman:
I'd like to wait a bit more before changing this, try to fix most of the misuses before we make the switch.

BanzaiMan:
Ping me when this is ready.

juliohm:
Is there any planned date for when this PR will be merged?

tkelman:
No, hundreds of packages are still using release to test against 0.4. Now they're silently not testing against a version of julia they still should be supporting. Anyone who wants to test against 0.5 can easily do so by listing a version number. We should stop using and remove release as a key here.

juliohm:
@tkelman but so if people want to support a specific version, shouldn't they list it explicitly? I find the release keyword very useful as I don't have plans to be testing against older versions of Julia. I only plan to support the current release and what is coming up in nightly. Having to modify the travis file for every release is something that I am trying to avoid. Supporting old releases is a pain specially at this stage of Julia where things are changing radically.

tkelman:
The way the package manager works is you have to change the REQUIRE file to drop support for old versions of julia. Having the REQUIRE file not match the travis matrix is wrong, and using "release" ensures that they will stop matching.

juliohm:
I see the release keyword as a convenient way of specifying the current
release, how would I specify it otherwise?

-Júlio

juliohm:
Anyways, I'm just sharing my personal preferences, I agree with your point.

Thanks,
-Júlio

tkelman:
If you're paying attention and want to specify the latest version, you can look it up and specify it by version number. The issue is when you haven't touched the package for a while, and someone submits a pull request to change some code or add a feature. Chances are they won't change your .travis.yml or REQUIRE file unless they're checking things really carefully, in which case their proposed change will not be tested on the versions you intended to support when you last modified the package.

juliohm:
Makes sense, I changed my mind after your comments. Thank you for
explaining it.

-Júlio

PR #854, opened 2016-10-09
Implement SourceClear's addon.

hendychua:
This addon makes it easier to use SourceClear with Travis-CI to scan projects for vulnerable libraries used.

### DIFF
--- a/lib/travis/build/addons/srcclr.rb
+++ b/lib/travis/build/addons/srcclr.rb
@@ -9,16 +9,14 @@ module Travis
         SUPER_USER_SAFE = true
 
         def before_finish
-          srcclr_config = @config.to_s.strip.shellescape
+          srcclr_config = config.strip.shellescape
 
           sh.if('$TRAVIS_TEST_RESULT = 0') do
             sh.fold 'after_success' do
               if srcclr_config.casecmp('true') == 0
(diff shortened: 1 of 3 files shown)


lib/travis/build/addons/srcclr.rb
```diff
@@ -0,0 +1,29 @@
...
+require 'travis/build/addons/base'
+
+module Travis
+  module Build
+    class Addons
+      class Srcclr < Base
+        SUPER_USER_SAFE = true
+
+        def before_finish
+          srcclr_config = @config.to_s.strip.shellescape
```
BanzaiMan: Use `config` method to access the addon's configuration.

What is the rationale for `#to_s`? If the user passes `true` or `false` from the YAML file, that value should be passed to the addon as is. It is possible that the user passes something like `TrUe`, but that is a string with those characters, so the user is *expecting* that to be valued as such.

hendychua: The expected way to use the addon in `.travis.yml` is `srcclr: 'true'`. `#to_s` was used in case of a usage error where user did not pass in any values i.e. `srcclr: `. In such a case, the `config` is of `Hash` type instead of a `String` type, and calling `#strip` will throw an error.

I have removed `#to_s` as the user is expected to pass in the correct values now.

BanzaiMan:
Let us continue discussion on #856, which has my modifications.

PR #855, opened 2016-10-12
node_js: Allow skipping `npm install` with npm_skip_install

beheh:
Currently `npm install` is always run if a package.json is present.

This PR introduces a `npm_skip_install` to flag to disable this behaviour. This is useful, since a lot of alternative package managers are popping up, most recently [Yarn](https://yarnpkg.com/) ([GitHub](https://github.com/yarnpkg/yarn)), which work as full drop-in replacements for npm. To use them you currently would have to `rm -rf node_modules/` which also increases run time by quite some bit.

I therefore propose this flag. The default behaviour is the same - when the flag is omitted, we continue to run `npm install` as usual. I've also slightly cleaned up a related spec and added an additional one.

Issue: travis-ci/travis-ci#6713.

### DIFF
--- a/lib/travis/build/script/node_js.rb
+++ b/lib/travis/build/script/node_js.rb
@@ -40,9 +40,11 @@ module Travis
         end
 
         def install
-          sh.if '-f package.json' do
-            sh.cmd "npm install #{config[:npm_args]}", retry: true, fold: 'install'
-          end
+            unless npm_skip_install?
+              sh.if '-f package.json' do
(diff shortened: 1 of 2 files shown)


simi:
I think you can provide your own install command and this default one will be skipped.

BanzaiMan:
Thanks for opening the PR.

As @simi has pointed out, you can override the "install" section at will, depending on the requirements of the repository.

beheh:
Thanks for your help!

PR #856, opened 2016-10-12
Add support for SourceClear addon

BanzaiMan:
Supersedes https://github.com/travis-ci/travis-build/pull/854


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/513941

BanzaiMan:
@hendychua Is it desirable that the failure in SourceClear agent mark the entire build a failure (as seen in the test above)?

hendychua:
@BanzaiMan yes that should be the case because there might be a misconfiguration on the user's part and they should be alerted.

PR #859, opened 2016-10-13
Allow beta versions in Firefox on-the-fly installation

BanzaiMan:
* Version numbers such as `50.0b6` are now accepted
* When an invalid version is specified, the build script compiles correctly to be executed.


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/513997

cotsog:
Fixes https://github.com/travis-pro/team-sapphire/issues/420.

PR #862, opened 2016-10-19
Upgrade_dependencies argument 

muschellij2:
Added an argument of `upgrade_deps` to have the option to not upgrade dependencies if they  are already installed for R.  @jimhester @hadley @craigcitro


muschellij2:
Can someone comment as to why this is blocked?

jimhester:
The `travis_wait` changes need to be removed from this PR, they are not appropriate for the default build script.

I personally feel like adding a `update_deps` option is unwise, CRAN always expects all dependencies to be upgraded.

If users do not want to upgrade their dependencies to the latest versions defining a custom `install:` block in their `.yaml` seems more appropriate.

PR #869, opened 2016-10-25
Export TRAVIS_SUDO

meatballhat:
The rationale here is that users with mixed matrices have to determine sudo availability themselves, and doing something like `if ! sudo echo testing &>/dev/null; then export CONTAINERED=1; fi` in `.travis.yml` will trigger the queue selection matcher to route the job to fully-virtualized.  The most frequent solutions to this of which I'm aware are to add an env var to the `sudo: false` matrix entry, or doing such a sudo availability check in a separate script, which then also requires that script to have a side effect such as writing to a file or `eval`'ing the script's output.


rkh:
What will the value be set to on Mac?

meatballhat:
@rkh I would expect Mac to always be something other than `false`.  Whether that's `true`, or `required` or some other thing, I'm not sure, and I'm guessing there are no tests asserting as much.  I guess I wasn't thinking about anything other than `TRAVIS_SUDO=false` :grin:.

BanzaiMan:
What is the rationale for adding this env var?

meatballhat:
@BanzaiMan is the description unclear? :heart:

BanzaiMan:
Does this work for older repos the default of which is `sudo: required` when the `sudo` key is absent?

meatballhat:
@BanzaiMan I think I need more information.  What do you consider "working"?

BanzaiMan:
@meatballhat To me, working means the value is reliable in our builds under all circumstances. The only indication whether or not `sudo` is available is that `/usr/sbin/sudo` has the SUID bit set, i.e.,

```
test -u /usr/sbin/sudo # or `which sudo`
```

Relying on the configuration payload seems wrong to me.

meatballhat:
@BanzaiMan What about basing it on `paranoid`?

BanzaiMan:
@meatballhat Ah. Yes. That might work.

BanzaiMan:
Test 1 (old repo, defaulting to GCE): https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/515285#L166
Test 2 (old repo, with explicit `sudo: false`): https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/515286#L113

BanzaiMan:
@meatballhat We'll need docs.

meatballhat:
Adding docs to default env var list OK?

BanzaiMan:
@meatballhat Absolutely!

backspace:
Hey, did this ever get added to the documentation? It seems like not, so I made [this PR](https://github.com/travis-ci/docs-travis-ci-com/pull/814) to document `TRAVIS_SUDO`.

PR #871, opened 2016-10-27
Detect internal ruby at runtime

meatballhat:
rather than guessing a version based on the job payload.

Early versions of this patch implemented selection based on `rvm list` with tools like `sed` and `awk`, which would have meant the addition of a base image dependency for the use of `casher` and `dpl`.  The latest implementation is done with pure `bash` bits, which adds a bit of extra weight to the header, totaling `1445` bytes before rendering:

``` diff
--- a/wc-c
+++ b/wc-c
@@ -1 +1 @@
-    4559 lib/travis/build/templates/header.sh
+    6004 lib/travis/build/templates/header.sh
```

At first I thought that this wouldn't make a huge difference due to gzipping, but then I confirmed that the HTTP API does not have gzipping in place.  The addition of gzip/deflate support is part of #872.

With gzipping, the difference in header weight before rendering is `472` bytes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ env:
   global:
   - RUBYOPT='-W0'
   - COVERAGE=1
+  matrix:
+  - INTEGRATION_SPECS=0
+  - INTEGRATION_SPECS=1
 
 script:
   - bundle exec rspec spec
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,3 +33,5 @@ deploy:
   region: us-east-1
   on:
     branch: master
+    conditions:
+    - $INTEGRATION_SPECS == 1
--- a/lib/travis/build/script.rb
+++ b/lib/travis/build/script.rb
@@ -45,6 +45,10 @@ module Travis
   module Build
     class Script
       TEMPLATES_PATH = File.expand_path('../templates', __FILE__)
+      INTERNAL_RUBY_REGEX = ENV.fetch(
+        'TRAVIS_INTERNAL_RUBY_REGEX',
+        'ruby-(2\.[0-2]\.[0-9]|1\.9\.3)'
+      ).untaint
       DEFAULTS = {}
 
       class << self
@@ -138,7 +142,15 @@ module Travis
         end
 
         def header
-          sh.raw template('header.sh', build_dir: BUILD_DIR), pos: 0
+          sh.raw(
+            template(
+              'header.sh',
+              build_dir: BUILD_DIR,
+              internal_ruby_regex: INTERNAL_RUBY_REGEX,
+              root: '/',
+              home: HOME_DIR
+            ), pos: 0
+          )
         end
 
         def configure
--- a/lib/travis/build/script.rb
+++ b/lib/travis/build/script.rb
@@ -47,7 +47,7 @@ module Travis
       TEMPLATES_PATH = File.expand_path('../templates', __FILE__)
       INTERNAL_RUBY_REGEX = ENV.fetch(
         'TRAVIS_INTERNAL_RUBY_REGEX',
-        'ruby-(2\.[0-2]\.[0-9]|1\.9\.3)'
+        '^ruby-(2\.[0-2]\.[0-9]|1\.9\.3)'
       ).untaint
       DEFAULTS = {}
 
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -90,11 +90,15 @@ travis_internal_ruby() {
     source ~/.rvm/scripts/rvm &>/dev/null
   fi
 
-  rvm list \
-    | awk -F- '/ruby-(2\.[0-2]\.[0-9]|1\.9\.3)/ { print $2 }' \
-    | awk '{ print $1 }' \
-    | sort -n -r \
-    | head -1
+  local internal_ruby
+  internal_ruby="$(
+    rvm list \
+      | awk -F- '/ruby-(2\.[0-2]\.[0-9]|1\.9\.3)/ { print $2 }' \
+      | awk '{ print $1 }' \
+      | sort -n -r \
+      | head -1
+  )"
+  echo "${internal_ruby:-default}"
 }
 
 travis_assert() {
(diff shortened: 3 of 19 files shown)


meatballhat:
Appears to work OK yey.  https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/builds/514934

Should any part of this be made configurable?  The regex for valid ruby versions is currently buried inside a bash func in the `header.sh`.

lib/travis/build/templates/header.sh
```diff
@@ -85,6 +85,22 @@ function travis_nanoseconds() {
...
 
+travis_internal_ruby() {
+  if ! type rvm &>/dev/null; then
+    source ~/.rvm/scripts/rvm &>/dev/null
+  fi
+
+  local internal_ruby
+  internal_ruby="$(
+    rvm list \
+      | awk -F- '/ruby-(2\.[0-2]\.[0-9]|1\.9\.3)/ { print $2 }' \
```
BanzaiMan: I think this regex will miss `2.1.10`.

meatballhat: The regex matches OK, but `sort -n` means that `2.1.5` is "greater than" `2.1.10`.

meatballhat: @BanzaiMan I'm sure a solution is possible to select a future 2.1.10 over 2.1.5, but I'm also certain the code will be more convoluted, and I don't expect we'll have to really care about such a scenario.  WDYT?

meatballhat:
@BanzaiMan the spec failures are due to output mismatch thanks to rvm's GEM_PATH warning.  Thoughts on silencing the `source <bash-profile>` calls?

lib/travis/build/templates/header.sh
```diff
@@ -71,20 +73,47 @@ travis_time_finish() {
...
 
   $cmd -u $format
 }
 
+travis_internal_ruby() {
+  if ! type rvm &>/dev/null; then
+    source <%= home %>/.rvm/scripts/rvm &>/dev/null
+  fi
+  local i selected_ruby rubies_array rubies_array_sorted rubies_array_len
+  rubies_array=( "$(
```
henrikhodne: Should this be without the `"` around `$()`?

```
$ foo_ary=( "$( echo 1; echo 2; echo 3)" )
$ echo "${foo_ary[@]}"
1
2
3
$ echo "${foo_ary[0]}"
1
2
3
$ echo "${foo_ary[1]}"

```

meatballhat: Found it.  Pushing fix.

lib/travis/build/templates/header.sh
```diff
@@ -71,20 +73,47 @@ travis_time_finish() {
...
+  local i selected_ruby rubies_array rubies_array_sorted rubies_array_len
+  rubies_array=( $(
+    rvm list strings \
+      | while read -r v; do
+          if [[ ! "${v}" =~ <%= internal_ruby_regex %> ]]; then
+            continue
+          fi
+          v="${v//ruby-/}"
+          v="${v%%-*}"
+          echo "${v//./}_${v}"
```
henrikhodne: Wouldn't removing . mean that 2.1.10 is sorted above 2.3.0, since 2110 > 230?

BanzaiMan: In [`node_js.rb`](https://github.com/travis-ci/travis-build/blob/321ab1d94611723c0cd17a2c71b9cc74971e8d7e/lib/travis/build/script/node_js.rb#L120), we define:

```bash
function vers() {\n  printf \"1%03d%03d%03d%03d\" $(echo \"$1\" | tr '.' ' ')\n}
```

which compares dot-delimited version numbers (up to 4 of them) into a single integer so that they can be compared.

Maybe this function should be moved to `header.sh`, so that it can be used elsewhere.

meatballhat: @henrikhodne yes.  I really want to be able to not care about hypothetical future rubies like `2.1.10`.  Really really.

meatballhat: @BanzaiMan extraction and reuse SGTM :+1:

henrikhodne:
How does the sorting thing handle non-MRI rubies, like JRuby? Is that even something we need to worry about?

meatballhat:
@henrikhodne IIUC those rubies won't even be considered since RVM does not prefix them with `ruby-`, unless of course we decide to alter the regex

lib/travis/build/script.rb
```diff
@@ -45,6 +45,10 @@ module Travis
   module Build
     class Script
       TEMPLATES_PATH = File.expand_path('../templates', __FILE__)
+      INTERNAL_RUBY_REGEX = ENV.fetch(
+        'TRAVIS_INTERNAL_RUBY_REGEX',
+        'ruby-(2\.[0-2]\.[0-9]|1\.9\.3)'
```
BanzaiMan: @henrikhodne We are lazily skipping `jruby-*` here (assuming that JRuby all of a sudden start using these version numbers).

henrikhodne: This regex should maybe be `^ruby-...`, although there are no known other versions that I think of that contain the ruby-2.? or ruby-1.9.3 substring.

.travis.yml
```diff
@@ -10,6 +10,9 @@ env:
   global:
   - RUBYOPT='-W0'
   - COVERAGE=1
+  matrix:
+  - INTEGRATION_SPECS=0
+  - INTEGRATION_SPECS=1
```
BanzaiMan: Deployment needs fixing.

meatballhat: done!

.travis.yml
```diff
@@ -30,3 +33,5 @@ deploy:
   region: us-east-1
   on:
     branch: master
+    conditions:
```
BanzaiMan: I believe it is singular.

.travis.yml
meatballhat: For clarification: "Deployment needs fixing." =~ "A condition needs to be added to the `deploy:` section to ensure it only runs for one build matrix entry."

.travis.yml
meatballhat: It is `Array(hash[:conditions])` coerced

BanzaiMan: Ah. right.

PR #872, opened 2016-10-29
Various touchups to HTTP API

meatballhat:
including:
- serving static files with default (protected) sinatra static middleware
- disabling auth under more strict conditions
- minor alterations to response of `POST /script`
  - sets content type of `application/x-sh`
  - sets status code of 201
- including uptime and version headers in response to `GET /` and `GET /uptime`


rkh:
I don't think 201 is correct here. This does not create a resource that can be referenced by a URI set in the Location header, can it?

meatballhat:
@rkh That's very true with the current implementation.  I'm on the fence wrt purity, so I'm happy to switch it back to 200.

lib/travis/api/build/app.rb
```diff
@@ -54,27 +74,30 @@ class App < Sinatra::Base
...
 
-        get %r{/files/([\w\.]+)} do |file|
-          file_path = File.expand_path("../../files/#{file}", __FILE__)
-          if File.exist? file_path
-            content_type :txt
-            File.read(file_path)
-          else
-            status 404
-          end
+        get('/') { uptime }
```
igorwwwwwwwwwwwwwwwwwwww: ❤️

lib/travis/api/build/app.rb
```diff
@@ -37,6 +27,36 @@ class App < Sinatra::Base
...
+          return if auth_disabled?
+
+          unless env.key?('HTTP_AUTHORIZATION')
+            halt 401, 'missing Authorization header'
+          end
+
+          type, token = env['HTTP_AUTHORIZATION'].to_s.split(' ', 2)
+
+          ENV['API_TOKEN'].split(',').each do |valid_token|
+            if type == 'token' && token == valid_token
```
igorwwwwwwwwwwwwwwwwwwww: should use constant time compare

lib/travis/build.rb
```diff
@@ -10,6 +10,11 @@ module Build
     BUILD_DIR = File.join(HOME_DIR, 'build')
 
     class << self
+      def version
+        @version ||= `git rev-parse HEAD 2>/dev/null || \\
+                        echo "${SOURCE_VERSION:-fafafaf}"`.strip
```
igorwwwwwwwwwwwwwwwwwwww: how about "unknown" instead of "fafafaf"?

lib/travis/api/build/app.rb
```diff
@@ -37,6 +27,36 @@ class App < Sinatra::Base
           if ENV.key?('LIBRATO_EMAIL') && ENV.key?('LIBRATO_TOKEN') && ENV.key?('LIBRATO_SOURCE')
             use Metriks
           end
+
+          use Rack::Deflater
```
igorwwwwwwwwwwwwwwwwwwww: 👍

lib/travis/api/build/app.rb
```diff
@@ -54,27 +74,30 @@ class App < Sinatra::Base
...
 
-          content_type :txt
-          Travis::Build.script(payload).compile
-        end
+          compiled = Travis::Build.script(payload).compile
 
-        get '/uptime' do
-          status 204
+          content_type 'application/x-sh'
+          status 201
```
henrikhodne: Is 201 the best status code for this? No resource is being "created", travis-build mostly just "computes" one. For 201s I would expect I could then hit some endpoint to retrieve the created resource.

lib/travis/build.rb
BanzaiMan: The intention might to show a hex string here. We have `ffffff` in another place (in `system_info`).

meatballhat: `fafafaf` (and `fffffff`), to me, mean "unknown according to us", as opposed to being unknown according to something we don't control like a 3rd party lib or Heroku itself.  Yes, it's a bit special.  We are the primary consumers of this, so I consider the use of `fafafaf` to be a combination of useful and silly that is not harmful.  Please allow the `fafafaf`.

lib/travis/api/build/app.rb
meatballhat: Also noted by @rkh at the top level.  I agree that 200 is more appropriate.

lib/travis/build.rb
Lyoness: `fafafaf` as an internal meaning only to travis <- i collect these special tribal knowledge tidbits of importance. worth adding someplace official-ish?

lib/travis/api/build/app.rb
rkh: `Rack::Utils.secure_compare(a, b)`

lib/travis/build.rb
igorwwwwwwwwwwwwwwwwwwww: I personally find it very unclear and confusing. But if you really want to keep it, I won't veto.

lib/travis/api/build/app.rb
```diff
@@ -37,6 +27,35 @@ class App < Sinatra::Base
...
+
+          unless env.key?('HTTP_AUTHORIZATION')
+            halt 401, 'missing Authorization header'
+          end
+
+          type, token = env['HTTP_AUTHORIZATION'].to_s.split(' ', 2)
+
+          ENV['API_TOKEN'].split(',').each do |valid_token|
+            return if Rack::Utils.secure_compare(type, 'token') &&
+                      Rack::Utils.secure_compare(token, valid_token)
```
henrikhodne: From the docs for `secure_compare`:

> NOTE: the values compared should be of fixed length, such as strings that have already been processed by HMAC. This should not be used on variable length plaintext strings because it could leak length info via timing attacks.

I'm not sure if leaking length info is a bad thing, though? And if it's something we can avoid?

meatballhat: This is an improvement, at least? 😕

lib/travis/api/build/app.rb
```diff
@@ -54,27 +73,30 @@ class App < Sinatra::Base
...
-          file_path = File.expand_path("../../files/#{file}", __FILE__)
-          if File.exist? file_path
-            content_type :txt
-            File.read(file_path)
-          else
-            status 404
-          end
+        get('/') { uptime }
+        get('/uptime') { uptime }
+        get('/boom') { raise StandardError, ':bomb:' }
```
BanzaiMan: 💣

lib/travis/api/build/app.rb
meatballhat: I'm confused about the utility of `Rack::Utils.secure_compare` if it does not do constant-time comparison 😐

meatballhat: OH.   This means that the API in its current form will return 403 for everybody.  https://github.com/rack/rack/blob/7dd7646a932419f4d308bf2fbc91f4e77edfd4cb/lib/rack/utils.rb#L396-L404

henrikhodne: It is "constant-time", but it will return fast if the length of the two strings are unequal, so you could figure out how long the actual token is by sending requests that have longer and longer tokens and check when there's a peak in response time due to it suddenly checking the content and not returning early. I don't leaking this information is necessarily bad, though, I think it's mostly noted in the docs for the method in case you don't want to leak that information.

PR #875, opened 2016-10-31
Ensure that securely compared strings in API are hashed first

meatballhat:


henrikhodne:
The code looks correct, but this is leaving the realm of what I know much about security-wise, so I'll have to do more research before I can say if this is more or less secure (for example, I don't know if the process of hashing leaks information? Maybe it's better to hash them all once and just hash the request token on each request?)

meatballhat:
@henrikhodne this is def out of the realm of what I know about security-wise :grin:.  I like your idea regarding hashing the "static" strings once at app load time.

igorwwwwwwwwwwwwwwwwwwww:
Left a comment on the original PR here: https://github.com/travis-ci/travis-build/pull/872, quoting here:

> not be an issue in almost any scenario, since keys/tokens tend to be fixed length, and the security comes from the key space.
>
> as in: congrats, you know it's a 128 bit key, good luck finding the value in 2^128 possibilities.

In other words, I don't think this PR is necessary.

meatballhat:
@igorwwwwwwwwwwwwwwwwwwww happily closing! 😸

PR #876, opened 2016-11-02
Fetch version from dyno metadata var

meatballhat:
instead of `SOURCE_VERSION`, which is provided by a third party buildpack

### DIFF
--- a/lib/travis/build.rb
+++ b/lib/travis/build.rb
@@ -12,7 +12,7 @@ module Travis
     class << self
       def version
         @version ||= `git rev-parse HEAD 2>/dev/null || \\
-                        echo "${SOURCE_VERSION:-unknown}"`.strip
+                        echo "${HEROKU_SLUG_COMMIT:-unknown}"`.strip
       end
 
       def self.register(key)


meatballhat:
@BanzaiMan r?

BanzaiMan:
I was under the impression that `SOURCE_VERSION` was provided by Heroku. https://devcenter.heroku.com/changelog-items/630

`HEROKU_SLUG_COMMIT`, on the other hand, is provided by the "labs" feature, and may be null under certain circumstances: https://devcenter.heroku.com/articles/dyno-metadata I am not sure if this is a concern for us, though.

BanzaiMan:
See https://twitter.com/hiro_asari/status/755478275791654912

meatballhat:
In my tests, I found `SOURCE_VERSION` was available at slug compile time, but not dyno runtime, unless using a third party buildpack to effectively record the variable for use at dyno runtime.  Switching to the labs var is preferable in my mind because it's being provided directly by Heroku.

BanzaiMan:
We don't use any third-party buildpack, and `SOURCE_VERSION` is still available. I don't think we need to switch.

```sh-session
$ heroku buildpacks -a travis-build-production
=== travis-build-production Buildpack URL
heroku/ruby
```

meatballhat:
@BanzaiMan `SOURCE_VERSION` is available for you at dyno runtime?  If so, I can't explain what I found in my tests.

BanzaiMan:
@meatballhat Oh. Whoa. I think I had it backwards. In https://github.com/travis-ci/travis-build/pull/796, I clearly indicate that we should prefer `HEROKU_SLUG_COMMIT` to `SOURCE_VERSION`.

PR #878, opened 2016-11-04
Fix revdep_script

BanzaiMan:
When execution gets here, we are hit by `NoMethodError`.
See, for example, https://travis-ci.org/davidgohel/rvg/jobs/172562948.


BanzaiMan:
@jimhester Could you check this is correct? Thanks.

jimhester:
LGTM

PR #883, opened 2016-11-10
Ensure SHELL, TERM, and USER are defined

meatballhat:
A recent rollout of refreshed infrastructure for running container-based builds revealed the assumption that `SHELL`, `TERM`, and `USER` env vars were being defined by ... something ... somewhere... and that they would always be around.  As it turns out, when executing a job via the docker remoting API as opposed to SSH, these vars are not defined.  This addition ensures that they are defined, with values that *should* work for the use cases of which we're currently aware.

### DIFF
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -12,6 +12,16 @@ ANSI_GREEN="\033[32;1m"
 ANSI_RESET="\033[0m"
 ANSI_CLEAR="\033[0K"
 
+if [ $TERM = dumb ]; then
+  unset TERM
+fi
+: "${SHELL:=/bin/bash}"
+: "${TERM:=xterm}"
+: "${USER:=travis}"
+export SHELL
+export TERM
+export USER
+
 TRAVIS_TEST_RESULT=
 TRAVIS_CMD=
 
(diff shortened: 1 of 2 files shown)


lib/travis/build/templates/header.sh
```diff
@@ -12,6 +12,13 @@ ANSI_GREEN="\033[32;1m"
 ANSI_RESET="\033[0m"
 ANSI_CLEAR="\033[0K"
 
+: "${SHELL:=/bin/bash}"
+: "${TERM:=xterm-256color}"
```
henrikhodne: Does our web interface's ansiparser support 256 colors? If not, I think `xterm` might be better?

BanzaiMan: It is set to `xterm` on GCE. https://travis-ci.org/BanzaiMan/travis_production_test/builds/174902179#L458

meatballhat: I don't think it supports 256.  I think I misinterpreted something said in chat.  Checking...

meatballhat: Hh weird, sorry I refreshed and now I see Hiro's comment.

meatballhat:
Verified via https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/builds/515748

PR #885, opened 2016-11-11
[R] Add -Wall -pedantic to the site Makevars

jimhester:
CRAN sets these when checking packages, so it is useful to have similar
checks on travis

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -49,6 +49,7 @@ module Travis
           sh.export 'TRAVIS_R_VERSION_STRING', config[:r].to_s, echo: false
           sh.export 'R_LIBS_USER', '~/R/Library', echo: false
           sh.export 'R_LIBS_SITE', '/usr/local/lib/R/site-library:/usr/lib/R/site-library', echo: false
+          sh.export 'R_MAKEVARS_SITE', '~/.Makevars.site', echo: false
           sh.export '_R_CHECK_CRAN_INCOMING_', 'false', echo: false
           sh.export 'NOT_CRAN', 'true', echo: false
           sh.export 'R_PROFILE', "~/.Rprofile.site", echo: false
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/r.rb
```diff
@@ -164,6 +165,10 @@ def configure
               options_repos = "options(repos = c(#{repos_str}))"
               sh.cmd %Q{echo '#{options_repos}' > ~/.Rprofile.site}
 
+              # Set -Wall -pedantic in the site Makevars to catch more issues,
+              # as CRAN runs with -pedantic
+              sh.cmd %Q{echo 'PKG_CXXFLAGS=-Wall -pedantic\nPKG_CFLAGS=-Wall -pedantic' > ~/.Makevars.site}
```
krlmlr: I think this has to be `PKG_CXXFLAGS := $(PKG_CXXFLAGS)` etc., otherwise `PKG_CXXFLAGS` given by the package's Makevars aren't honored.

krlmlr: Also, might be worthwhile to add `LTO=-O0` and `LDFLAGS=-fuse-ld=gold` to speed up compilation and linking, respectively.

jimhester: From [R Installation and Administration](https://cran.r-project.org/doc/manuals/r-release/R-admin.html#Customizing-package-compilation)
> There is also provision for a site-wide Makevars.site file under R_HOME/etc (in a sub-architecture-specific directory if appropriate). This is read immediately after Makeconf, and the path to an alternative file can be specified by environment variable R_MAKEVARS_SITE.

So the site wide Makevars is processed before the user specified Makevars.

Those other options sound like useful suggestions, but I don't think they are appropriate to enable for all packages, setting too many options will just make it harder for users to replicate travis results on their machines.

krlmlr: Shouldn't we then use another environment variable? I'm not sure if user's Makevars are expected to honor existing `PKG_???` values. (Mine don't.)

jimhester: The benefit of using a site wide Makevars is that the user is free to override or inherit from it if they want to.

It is up to the user if they want to honor existing `PKG_` values.

krlmlr: Do we want the user to override our `-Wall -pedantic` if CRAN doesn't seem to allow this?

BanzaiMan:
Sorry about the delay. Is this ready for testing?

jimhester:
Yes I believe so, if we could deploy it to the testing env I can run some tests the next couple of days.

BanzaiMan:
@jimhester Cool. I plan to deploy this tomorrow (2016-12-13) at 14:00 UTC.

BanzaiMan:
deployed to .org staging

BanzaiMan:
This is not tested, as far as I know. Is it still needed? If so, please let us know.

krlmlr:
I'd like to see this implemented. What do we need to test?

PR #887, opened 2016-11-13
Update deps and bump ruby to 2.3.1

meatballhat:


.travis.yml
```diff
@@ -18,6 +18,9 @@ script:
   - bundle exec rspec spec
   - for f in examples/*.sh.txt; do echo "checking bash syntax $f"; bash -n $f; done
 
+after_success:
+- codeclimate-test-reporter
```
BanzaiMan: https://travis-ci.org/travis-ci/travis-build/jobs/175423342#L912 it's not there. Do we need it?

meatballhat: Ack, this should probably be `bundle exec`'d.  This was previously being done in-process, but that way is no longer supported.

meatballhat:
:+1:

PR #889, opened 2016-11-14
Reload jdk_switcher_rc after removing symlink

BanzaiMan:
So that `$ORACLEJDK8_JAVA_HOME` is set correctly.

This resulted in worse problem than before, where we ended up with:

    update-java-alternatives: directory does not exist: /usr/lib/jvm/java-8-oracle-amd64


BanzaiMan:
`$HOME/.jdk_switcher_rc` is read before the symlink is removed, so `jdk_switcher` would still try to load Java 8u65 instead (with a link that no longer exists). This PR fixes the issue.

PR #890, opened 2016-11-15
apt: Remove "BETA" apt addon quite mature now

grooverdan:
Output fix only - Its a lot more mature than BETA now.

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -97,7 +97,7 @@ module Travis
           end
 
           def add_apt_sources
-            sh.echo "Adding APT Sources (BETA)", ansi: :yellow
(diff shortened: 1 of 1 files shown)


BanzaiMan:
I disagree.

BanzaiMan:
I apologize for being terse earlier.

My general feeling is that we are still feeling out what and how the addon should work. The case in point is the recent addition of Trusty support.

The general configuration probably would not change (to avoid breakage), but the underpinning could.

BanzaiMan:
We'll discuss more internally and get back to you! Thanks for reminding us about the "BETA" designation of this feature.

BanzaiMan:
We plan to remove this BETA flag shortly.

This PR is superseded by https://github.com/travis-ci/travis-build/pull/928.

PR #891, opened 2016-11-16
Remove apt source for apt.dockerproject.org

BanzaiMan:
There is an issue with this APT source preventing use of `apt`.

See
https://github.com/travis-ci/travis-ci/issues/6881 for more details.

### DIFF
--- a/lib/travis/build/appliances/rm_dockerproject_apt_source.rb
+++ b/lib/travis/build/appliances/rm_dockerproject_apt_source.rb
@@ -1,0 +1,25 @@
+require 'travis/build/appliances/base'
+
+module Travis
+  module Build
+    module Appliances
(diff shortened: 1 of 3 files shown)


lib/travis/build/appliances/rm_dockerproject_apt_source.rb
```diff
@@ -0,0 +1,27 @@
+require 'travis/build/appliances/base'
+
+module Travis
+  module Build
+    module Appliances
+      class RmDockerprojectAptSource < Base
+        MSG = <<-EOMSG
+Due to infrastructure issues with apt.dockerproject.org we're disabling the apt repository for now.
```
justinjdickow: "we've disabled" :P

PR #894, opened 2016-11-17
Fix ANSI escape sequence for green

BanzaiMan:
It was actually yellow before.


renee-travisci:
Tests aren't passing :(

PR #895, opened 2016-11-17
Add yarn support

BanzaiMan:
This PR adds yarn support to Travis CI.

See https://github.com/travis-ci/travis-ci/issues/6720

# How Yarn support works

## Yarn usage detection

If both `packag.json` and `yarn.lock` file exist in the repository root, we activate yarn support.

If `yarn.lock` exists but the defined Node.js version does not meet the minimum requirement for Yarn (versions less than 4), we fall back to the default `npm install`.

## Making the `yarn` command available

During the default `install` phase, if the `yarn` command is not available, it is installed with the following command, as described in https://yarnpkg.com/en/docs/install#alternatives-tab:

```
curl -o- -L https://yarnpkg.com/install.sh | bash
```

This is much faster on OS X/macOS than the documented `brew update && brew install yarn`, and as fast as (or slightly faster than) installing it via `apt-get install` on Ubuntu.

If repositories are already installing `yarn` in some way during the `before_install` or the overridden `install`, the installation will not take place (because `yarn` is available in the former case, and our installation steps will not take place in the latter case).

## Caching

This PR also defines a new cache directive:

```yaml
cache: yarn
```

or, equivalently, with other cache directives:

```yaml
cache:
  yarn: true
  bundler: true
  directories:
    - node_modules
```


Daniel15:
Awesome, thank you @BanzaiMan! 😄

lib/travis/build/script/node_js.rb
```diff
@@ -169,6 +194,12 @@ def setup_npm_cache
           def iojs_3_plus?
             (config[:node_js] || '').to_s.split('.')[0].to_i >= 3
           end
+
+          def install_yarn
+            sh.cmd    "curl -o- -L https://yarnpkg.com/install.sh | bash", echo: true
```
Daniel15: I wonder if it's worth adding an option to specify which version of Yarn to use? We have a GitHub issue to add a `--version` flag to the install script (see https://github.com/yarnpkg/yarn/issues/1856) but you could achieve it today by downloading the right tarball (eg. https://yarnpkg.com/downloads/0.16.1/yarn-v0.16.1.tar.gz) and extracting it yourself. That's essentially what the install script is doing 😄

BanzaiMan:
I should add that the cache directive will cache `$HOME/.yarn-cache`, and that we modify `$PATH` to include `$HOME/.yarn/bin`.

lib/travis/build/script/node_js.rb
BanzaiMan: How common is such a need? If possible, I would like us to Do The Reasonable Thing™ automatically. If a different version of yarn is desired, I think the users can do that at their own will. (We do not support different versions of Bundler or Pip, for example.)

BanzaiMan:
To think, maybe we should run `npm install` if Node.js version does not meet `yarn`'s requirement.

BanzaiMan:
Documentation PR is https://github.com/travis-ci/docs-travis-ci-com/pull/811.

lib/travis/build/script/node_js.rb
BanzaiMan: If a different version of `yarn` is desired, the user can install it on `before_install`.

jaumard:
The problem by using `yarn.lock` is that project who don't put this file under versioning will not be able to use yarn at all :( 

Use case : a node framework that want to test his installation with npm and yarn will not be able to do it, with Trails framework we want to test installations and generators with both. 

A `packageManager: npm/yarn` field on `travis.yml` would be also great. 

Anyway, thanks for your great job !! :D

BanzaiMan:
@jaumard Could you elaborate on what you are asking? If there is no `yarn.lock` in the repository root and you still want to use yarn, you can always override the `install` phase to use yarn. You'd still need to install yarn until we roll out images that has it pre-installed, but I tend to think this is an unusual use case. Am I mistaken?

jaumard:
@BanzaiMan of course we can always override install phase but I think it will be better/easier/natural to just have a field to choose witch package manager we want like we choose the node versions or os. 

> this is an unusual use case
Yes and no if all library/frameworks/generators modules want to test yarn and npm install it will be a lot of cases ^^

BanzaiMan:
We will consider `package_manager` key for future enhancement.

We will deploy this Monday or Tuesday next week.

BanzaiMan:
By the way, you can see the test here: https://staging.travis-ci.org/BanzaiMan/example-yarn-package/builds/525479

mattgrande:
I just had a build complete (node 6.9.1, yarn.lock file present) and I noticed that it still used `npm install`.

As I'm typing this, I'm noticing that this was just merged in three hours ago... Can I assume this is still being rolled out, and my builds will eventually use yarn, as it becomes available?

SpaceK33z:
> We will deploy this Monday next week.

see above

mattgrande:
Missed that; thanks.

lib/travis/build/script/node_js.rb
```diff
@@ -41,7 +41,20 @@ def announce
 
         def install
           sh.if '-f package.json' do
-            sh.cmd "npm install #{config[:npm_args]}", retry: true, fold: 'install'
+            sh.if "-f yarn.lock" do
+              if version.to_i < 4
```
BanzaiMan: This matches `stable`, `node`, and most notably `v4`, which will cause problems.

We should check the Node.js version at run time instead.

jdevalk:
Is deploy still planned for today?

joshk:
@jdevalk that is the plan :)

Hiro lives on the East Coast, so will be online soon to start the roll out.

lib/travis/build/script/node_js.rb
SimenB: Does it work if I specify an LTS code name? (`argon` or `boron` currently)

jdevalk:
Awesome, I hope @BanzaiMan will update us when it's live :D my builds are waiting ;)

ai:
Agree, let's merge current patch and then will improve it in other PRs. I soooo waiting `yarn` support without long `before_install` scripts 😊

lib/travis/build/script/node_js.rb
```diff
@@ -58,6 +71,19 @@ def cache_slug
           super << '--node-' << version
         end
 
+        def setup_cache
+          if data.cache?(:yarn)
+            sh.fold 'cache.yarn' do
+              sh.echo ''
+              directory_cache.add '$HOME/.yarn-cache'
```
zupo: This has moved to `$HOME/.cache/yarn` and so currently `cache: yarn` does not really work, the cache is empty.

joshk: This was updated 2 years ago: https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/node_js.rb#L95

zupo: Ugh, :facepalm: Sorry for the noise! 

(That said, I still do have problems with empty yarn cache between build stages, but haven't nailed it down to a self-contained reproducible example, yet)

joshk: ah, that is a different issue, caches are individual per job. There is not built in facility to share artifacts between jobs and stages right now. We are looking into addressing this.

PR #896, opened 2016-11-21
Determine Node.js version at run time

BanzaiMan:
nvm supports additional shorthands for Node.js versions, such as
"stable", as well as leading "v" (e.g., "v6"), which evaluates to "0"
when passed to `#to_i`. This is bad news for determining whether
`yarn`'s requirment is met.

Instead, we use `node --version` output at run time to decide.

Note that `node --version` does have a leading "v", so we need to drop
that before passing on to the `vers2int` function.


BanzaiMan:
Here's a test: https://staging.travis-ci.org/BanzaiMan/example-yarn-package/jobs/525556

PR #900, opened 2016-11-28
Accommodate firefox binaries by fixing $PATH

BanzaiMan:
The firefox archive contains more than just `firefox`, so we need to put
the directory into which we expanded the archive to `$PATH`, so that all
associated binaries are found correctly.

This resolves https://github.com/travis-ci/travis-ci/issues/4649.


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/526115#L159

PR #901, opened 2016-11-30
Handle heterogenous MySQL execution environments

meatballhat:
with artisanally crafted PHP/HHVM goodies.

### DIFF
--- a/lib/travis/build/appliances/services.rb
+++ b/lib/travis/build/appliances/services.rb
@@ -24,6 +24,15 @@ module Travis
               end
             end
 
+            if services.delete('mysql')
+              sh.if "$(grep -E '^socket = /run/mysql-5.6' #{HOME_DIR}/.my.cnf)" do
+                sh.cmd "sudo service mysql-5.6 start", assert: false, echo: true, timing: true
+              end
+              sh.else do
+                sh.cmd "sudo service mysql start", assert: false, echo: true, timing: true
+              end
+            end
+
             services.each do |name|
               sh.cmd "sudo service #{name.shellescape} start", assert: false, echo: true, timing: true
             end
--- a/lib/travis/build/templates/php.sh
+++ b/lib/travis/build/templates/php.sh
@@ -1,0 +1,28 @@
+<%# munge php.ini configs with correct mysql-5.6 socket path when applicable %>
+travis_configure_php_mysql56() {
+  if ! grep -qE '^socket = /run/mysql-5.6' <%= home %>/.my.cnf; then
+    return
+  fi
+
+  local mysql_unix_port
+  mysql_unix_port="$(grep ^socket <%= home %>/.my.cnf)"
+  mysql_unix_port="${mysql_unix_port##*=}"
+  mysql_unix_port="${mysql_unix_port// /}"
+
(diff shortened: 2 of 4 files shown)


meatballhat:
Appears to work https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/builds/527158

lib/travis/build/appliances/services.rb
```diff
@@ -24,6 +24,15 @@ def apply
               end
             end
 
+            if services.delete('mysql')
```
BanzaiMan: Can we announce the start of `mysql` service? (or does it happen elsewhere?)

lib/travis/build/appliances/services.rb
```diff
@@ -24,6 +24,15 @@ def apply
...
+              sh.if "$(grep -E '^socket = /run/mysql-5.6' #{HOME_DIR}/.my.cnf)" do
+                sh.cmd "sudo service mysql-5.6 start", assert: false, echo: true, timing: true
```
joepvd: The service name change should be announced and documented for people who change the configuration. According to GitHub search, there are +/- 250 public repos who have `sudo service mysql` in the build config. Can the service be renamed/aliased?

lib/travis/build/templates/php.sh
```diff
@@ -0,0 +1,28 @@
...
+        sed -i "/^[a-z_]*mysql.*socket/s:\$: ${MYSQL_UNIX_PORT}:" "${ini_file}"
+      done
+  fi
+
+  if [[ -f '<%= root %>/etc/hhvm/php.ini' ]]; then
+    echo "
+pdo_mysql.default_socket = ${MYSQL_UNIX_PORT}
+mysqli.default_socket = ${MYSQL_UNIX_PORT}
+hhvm.mysql.socket = ${MYSQL_UNIX_PORT}
+" | sudo tee -a '<%= root %>/etc/hhvm/php.ini' >/dev/null
```
joepvd: hhvm might need a restart after this config change.

joepvd:
Notify when done:

https://secure.helpscout.net/conversation/274962408/38692/
https://github.com/travis-ci/travis-ci/issues/6855
https://github.com/travis-ci/travis-ci/issues/6842

lib/travis/build/appliances/services.rb
meatballhat: I suspect a service alias is the easiest solution.  I don't want to take on the burden of maintaining our own MySQL cookbook (again), and the community cookbook we're using adds a suffix to `mysql-` by default.

lib/travis/build/templates/php.sh
meatballhat: SGTM

lib/travis/build/appliances/services.rb
meatballhat: The situation is worse than I thought: https://travis-ci.org/meatballhat/cloaked-octo-tyrion/builds/180440706

I'm going to work on cleaning up the various mysql service bits, including aliasing `mysql-5.6` => `mysql`, although I'm pretty sure that upstart is going to have opinions about this.

meatballhat:
This should no longer be necessary thanks to https://github.com/travis-ci/travis-cookbooks/pull/784

PR #903, opened 2016-12-06
Remove `./node_modules/.bin` from $PATH

BanzaiMan:
Some npm modules define commands with widely used bash functions
(and probably others).
Putting this component in `$PATH` causes subtle issues, such as
https://github.com/travis-ci/travis-ci/issues/5092.
We are removing it, so that this is no longer the case.

This will most likely break some builds, so we add a warning.

### DIFF
--- a/lib/travis/build/script/node_js.rb
+++ b/lib/travis/build/script/node_js.rb
@@ -18,6 +18,7 @@ module Travis
         def setup
           super
           convert_legacy_nodejs_config
+          remove_node_modules_bin_from_path
           update_nvm
           nvm_install
           npm_disable_prefix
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Tested in staging.

Before: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/528065#L168
After: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/528067#L160-L178

nathanhammond:
Note that, to my knowledge, this change breaks every Ember addon out there because of our default blueprint: https://github.com/ember-cli/ember-cli/blob/74562a4/blueprints/addon/files/.travis.yml#L40

backspace:
This will break things for almost all Ember projects, as `ember test` is the default way to run tests and that assumes the Node-installed `ember` command is available.

nathanhammond:
This doesn't appear to have landed yet we currently have reported failures from users in our Slack channel. Also, hi, @backspace! <3

joshk:
I'm not sure I support this change. This will cause major issues for many people using Node.

Turbo87:
@BanzaiMan instead of removing the path completely, would it maybe be possible to move it to the end of the list as a fallback? that way the native executables are preferred, but the Ember addon builds won't fail.

BanzaiMan:
The offending bit is https://github.com/travis-ci/travis-cookbooks/pull/786, which renders this PR obsolete. I'll work on a fix for the issue shortly.

mc10:
Since this PR is now obsolete, it can be closed now right?

PR #904, opened 2016-12-06
Embed GHC-finding func for haskell jobs

meatballhat:
so that there is no dependency on the underlying VM having such a tool
available.  Also, modified the func a bit to accept only a default version and
otherwise use the versions available on the system for matching.


meatballhat:
Related to https://github.com/travis-ci/packer-templates/issues/366

PR #905, opened 2016-12-06
Add ./node_modules/.bin to $PATH

BanzaiMan:
https://github.com/travis-ci/travis-cookbooks/pull/786 removed it, but
this broke many Ember-based repositories, so we will reinstate it.

This is meant to be a hot fix, which would be reverted after a new set
of build images are rolled out.

### DIFF
--- a/lib/travis/build/script/node_js.rb
+++ b/lib/travis/build/script/node_js.rb
@@ -17,6 +17,7 @@ module Travis
 
         def setup
           super
+          prepend_path './node_modules/.bin'
           convert_legacy_nodejs_config
           update_nvm
           nvm_install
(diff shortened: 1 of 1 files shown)


backspace:
Someone suggested [appending instead of prepending](https://github.com/travis-ci/travis-build/pull/903#issuecomment-265223249), would that be better to address the reason for this?

BanzaiMan:
@backspace I don't think that'll work, because `which` is a Bash built-in function, and it's not on `$PATH`. That issue must be addressed separately.

Turbo87:
@BanzaiMan right, that makes sense, thanks!

PR #909, opened 2016-12-07
Switch from raw ENV to travis-config

meatballhat:
- [x] test in staging

### DIFF
--- a/lib/travis/build.rb
+++ b/lib/travis/build.rb
@@ -5,9 +5,9 @@ module Travis
     HOME_DIR  = '$HOME'
     BUILD_DIR = File.join(HOME_DIR, 'build')
 
-		def config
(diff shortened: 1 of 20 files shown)


lib/travis/build.rb
```diff
@@ -44,3 +46,9 @@ def by_lang(lang)
     end
   end
 end
+
+require 'core_ext/hash/deep_symbolize_keys'
```
BanzaiMan: Is there a reason for this juxtaposition?

meatballhat: If left at the top, there's an unresolvable dependency cycle.  Try it!

meatballhat:
This is behaving well in staging, fwiw :+1:

PR #912, opened 2016-12-11
Fix .NET Core 1.1 Support

oschwald:
I believe this will fix https://github.com/travis-ci/travis-ci/issues/6920. The issue is that .NET Core 1.1 has a version of `1.0.0-preview2-1-003177`. Presumably, future 1.1.x releases with the `project.json` build system will have the same version string format.

### DIFF
--- a/lib/travis/build/script/csharp.rb
+++ b/lib/travis/build/script/csharp.rb
@@ -13,7 +13,7 @@ module Travis
         }
 
         MONO_VERSION_REGEXP = /^(\d{1})\.(\d{1,2})\.\d{1,2}$/
-        DOTNET_VERSION_REGEXP = /^\d{1}\.\d{1,2}\.\d{1,2}(-preview\d{1}-\d{6})?$/
+        DOTNET_VERSION_REGEXP = /^\d{1}\.\d{1,2}\.\d{1,2}(?:-preview\d+(\.\d+)?(?:-\d)?-\d{6})?$/
 
         def configure
           super


oschwald:
Pinging @joshua-anderson @akoeplinger @nterry as requested by the docs.

Joshua-Anderson:
Change looks good! Looking at all the available versions, there also seems to be a `1.0.0-preview2.1-003155`, do we want to support that as well?

oschwald:
@Joshua-Anderson, I force pushed a change for that version string format. I also made the group non-capturing, as I didn't see that being used anywhere.

BanzaiMan:
hello! I'll deploy this to https://staging.travis-ci.org tomorrow (2016-12-13) at 14:00 UTC.

syedhassaanahmed:
@oschwald Do we need to handle `1.0.0-preview2-1-003177` here as well? https://github.com/travis-ci/travis-build/blob/master/spec/build/script/csharp_spec.rb

oschwald:
@syedhassaanahmed, it might be nice to add a test there for the various version numbers in the wild.

oschwald:
@BanzaiMan, thanks! A [1.0.0-preview2-1-003177 (.NET Core 1.1) build](https://staging.travis-ci.org/maxmind/MaxMind-DB-Reader-dotnet/jobs/529213) seemed to work well for me.

BanzaiMan:
deployed to production

rovale:
Is it also possible to add support for [1.0.0-preview5-004232](https://github.com/dotnet/cli#installers-and-binaries)? This version deals with the csproj project format.

oschwald:
@rovale, that version string should already work with the regex. If you are having problems using it, please create a new issue.

rovale:
Yeah, thanks, the regex works, that doesn't seem to be the problem. I created a new issue: #7026

PR #914, opened 2016-12-13
Update sauce connect to 4.4.2

halkeye:


BanzaiMan:
Sorry for overlooking this. We merged #921 instead.

Thanks for taking the time to work on this PR. We really appreciate it.

PR #917, opened 2016-12-21
Set `cat` as default $PAGER

BanzaiMan:
Some commands may default to PAGER which requires user interaction.
This could hang the build.


BanzaiMan:
I do not believe this value is set in any of the build images, so any user who relies on this value should be setting it themselves, and the user-defined value will override this. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/529993#L122

PR #923, opened 2017-01-10
Update composer to 1.0.0 once

BanzaiMan:
Composer 1.0.0 beta 2 introduced the concept of channels in
`composer self-update`:
https://github.com/composer/composer/blob/master/CHANGELOG.md#100-beta2---2016-03-27
Before this feature, it always updated to the latest commit, which
could be buggy.

By stepping at 1.0.0, we ensure that `composer self-update` updates
to the stable release, avoiding a buggy commit.

### DIFF
--- a/lib/travis/build/script/php.rb
+++ b/lib/travis/build/script/php.rb
@@ -160,6 +160,7 @@ hhvm.libxml.ext_entity_whitelist=file,http,https
         end
 
         def composer_self_update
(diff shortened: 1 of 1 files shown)


BanzaiMan:
1. This adds a few seconds to PHP builds, but I reckon it is an acceptable cost.
1. When the build images are updated and have a version of Composer newer than 1.0.0, it is once downgraded to 1.0.0, before it is upgraded again to the latest stable version. When all the images are updated, then, this commit should be reverted.

PR #925, opened 2017-01-12
Recover SC startup error gracefully

BanzaiMan:
While waiting for Sauce Connect to start, sleep only when the `sc`
command is running.
When `sc` stops running, get out of the sleep loop immediately, so
that we do not waste time waiting for the readyfile that will never
come.

At the same time, we retry opening Sauce Connect tunnel up to 3 times
just in case the failure is temporary.

See https://github.com/travis-ci/travis-ci/issues/3091 for more
context.


BanzaiMan:
A test: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/531885

BanzaiMan:
This has been deployed to production.

PR #926, opened 2017-01-12
Allow for hiding system info dump

meatballhat:
This change allows for users to opt out of the `system_info` fold, and associated addition to logs, by including:

``` yaml
system_info: false
```

The default behavior remains consistent with current behavior, with the contents of the system info file `cat`'d to stdout.


BanzaiMan:
👍  for making this optional. Is it desirable to have a blurb about the addon in the default `Build system information` fold?

meatballhat:
@BanzaiMan I could go either way on including info about the addon in the default fold.  Sounds reasonable to me 👍

BanzaiMan:
@meatballhat Cool. Let's do that. Since it _was_ there by default for a while, users may wonder where it went.

meatballhat:
Current default appearance of expanded fold:

lib/travis/build/appliances/show_system_info.rb
```diff
@@ -34,13 +35,26 @@ def show_travis_build_version
...
-              sh.cmd "cat #{info_file}"
+              sh.echo "Info file: #{info_file}"
             end
+            sh.if "-f #{info_json_file}" do
+              sh.echo "Info JSON file: #{info_json_file}"
+            end
+          end
+
+          def advertise_system_info_addon
+            sh.echo "Extended system info available via:", ansi: :yellow
```
BanzaiMan: How about `…available via the "system_info" addon`?

meatballhat: I'm cool with that, def.

meatballhat:
Current expanded fold:

ctrlaltjustine:
Hello engineers! From a user interface and user experience point of view, I'd like to add my 2¢ if I may?

First, before we take away information users have been used to seeing for year, I would be in favor of running this past users to see if they'd be interested in this feature.

Then if there is interest from users, (if I'm understand this user-flow correctly) I believe a better user experience would effectively be reversing what this PR would require users to do. By that I mean, I think users should opt-in to hide this information from their log — instead of hiding it by default and requiring users to get it back.

I understand the thinking behind this was largely about cost savings to storage in S3. If users in fact feel this information is unnecessary, with the user-flow I mentioned about, they'll hopefully adjust their travis.yml to hide the information, and hopefully we'll still see some of those savings.

emdantrim:
If we stored the information somewhere in the docs or some other location, and just linked to it in the build, we could solve this problem while still only keeping its contents a click away. This would allow us to still reduce 400 lines to ...a couple. Would that help?

meatballhat:
@emdantrim I'm guessing the problem with linking is that the log display doesn't currently support linkification, or embedding of links in any way.  IIUC support for this has been discussed elsewhere, but I don't know how close to reality it may be.

joshk:
Thank you for making these changes.

To understand this feature best, and the motivation behind it, what prompted work on this? eg. did any customers or users ask for this?

What is the success criteria for this feature? eg. are we looking for a percentage or rough number of users to opt into this?

emdantrim:
@joshk We store 400 lines of data about images that is an exact readout of a file that exists on a build image, which never changes. We're paying for storing the same thing over and over, for every build -- there is no deduplication on S3 (or at least, not that is visible to us), so we end up paying to store the same lines millions of times. Most people probably don't even glance through this information, so to me it seems kind of hard to justify the cost of printing out all the information as a convenience when we literally pay actual money to store that information later on.

Linking to it elsewhere would give us a the same ability, but we would only have to save the information once per image.

joshk:
Thanks for the information @emdantrim 

I talked to @roidrage about the storage costs and the total we would save would be around $100 per month based on 100 million jobs. 

S3 costs are really trivial compared to our VM usage, so I think we can put the cost reasoning to one side.

Instead it is better to look at the user benefit. Since this information is hidden by default, and removing the information doesn't speed up the job noticeably, plus users would be opting in to remove the information from the log entirely, I don't think adding this feature should be merged in.

There are ways we can measure how many people read the system information section. I had it working with heap.com, but the cost of this service would have been about $1k a month, so I'm looking at other solutions.

meatballhat:
@joshk Please note that the default behavior was inverted based on yours and @ctrlaltjustine's feedback.  Are you still against merging?

joshk:
Hi @meatballhat 

I am sorry. My first comment wasn't very clear as my concern at the end of it was partially hidden under other statements.

I wasn't trying to say that by switching the logic I would be supportive of the feature, but I can see how it could be read that way.

It isn't clear, to me at least, if any users have asked for this feature, or what the core motivtian of this change is. And as the cost savings argument is not weighing heavily on the decision, I don't think this feature brings value to Travis.

I'm sorry, but I do not think we should merge this in.

PR #930, opened 2017-01-17
Add "1.8.x" as an alias for Go 1.8rc1

tianon:
As always, happy to adjust, rebase, amend, force push, etc as desired. :+1:

I didn't update any of the other `1.x` aliases since this is still a pre-release. :smile:


tianon:
I'd appreciate any hints as to what I screwed up here -- I added to both places that `1.7.x` is referenced, but it appears that my alias isn't getting applied appropriately (since the test checking to make sure it is fails). :disappointed:

meatballhat:
@tianon This mapping needs to be added to the config default, too ❤️   Take a look in `./lib/travis/build/config.rb`.

tianon:
@meatballhat yeah -- I thought I had, but it looks like GitHub simply delayed my force push :sweat_smile: :crossed_fingers:

PR #934, opened 2017-01-20
Add Elm language support

rtfeldman:
This adds [Elm](http://elm-lang.org/) support to Travis.

Happy to do as much revision to this as necessary! Also I'd be happy to push branches of some of my Elm projects to have them use `language: elm` in their `.travis.yml` files for the staging tests.


BanzaiMan:
Hi, there! Thank you for your interest in adding Elm to Travis CI.

If you have not done so, please read https://docs.travis-ci.com/user/languages/community-supported-languages/ for some additional details. (Some of the details are a bit out of date; in particular, the bits on `travis-core` in https://docs.travis-ci.com/user/languages/community-supported-languages/#Code since that library is no longer relevant to community-supported languages. See https://github.com/travis-ci/docs-travis-ci-com/pull/819/.)

rtfeldman:
Yep! That page was my starting point. 😁

rtfeldman:
@BanzaiMan Just wanted to check - is there anything you need from me on this? Or just waiting on review?

BanzaiMan:
Please gather a few other volunteers to provide support. Thanks.

lukewestby:
@banzaiman @rtfeldman I can help

avh4:
I can help as well.  I'll also add this to the elm-community projects list so we can document who's actively part of the support team.

rtfeldman:
@BanzaiMan done! 😄

stoeffel:
@rtfeldman I can help!

rtfeldman:
@BanzaiMan friendly ping! 😃

rtfeldman:
@BanzaiMan anything we still need to do to unblock review?

rtfeldman:
@BanzaiMan friendly ping! 🙂

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,113 @@
...
+          def elm_version
+            config[:elm_version].to_s
+          end
+
+          def elm_test_version
+            config[:elm_test_version].to_s
+          end
+
+          def install_elm
+            sh.cmd 'npm install -g elm-use@1.1.0', retry: true
```
BanzaiMan: Does `1.1.0` need to be hardcoded here?

BanzaiMan:
Am I correct to understand that Elm requires a Node.js/JavaScript image?

rtfeldman:
> Am I correct to understand that Elm requires a Node.js/JavaScript image?

@BanzaiMan Elm doesn't, but `elm-test` does, and it's reasonable to assume people will want that for CI. 😄

lib/travis/build/script/elm.rb
rtfeldman: Doesn't need to be hardcoded, but I wouldn't want it to auto-upgrade to a breaking version.

This might be better?

```ruby
sh.cmd 'npm install -g elm-use@">=1.1.0 <0.2.0"', retry: true
```

I could also extract an `ELM_USE_NPM_VERSION_RANGE = ">=1.1.0 <0.2.0"` constant.

BanzaiMan: Do you mean `<0.2.0` or `<2.0.0`?

`>=1.1.0 <0.2.0` feels like it should resolve to an empty set.

BanzaiMan:
Are there other requirements, such as required Node.js version and NPM version?

lib/travis/build/script/elm.rb
rtfeldman: sorry, yes - `2.0.0`

so:

```ruby
sh.cmd 'npm install -g elm-use@">=1.1.0 <2.0.0"', retry: true
```

rtfeldman:
@BanzaiMan ideally Node 6 and npm 3 (any minor versions within those major versions should be fine).

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,113 @@
+module Travis
+  module Build
+    class Script
+      class Elm < Script
```
BanzaiMan: Does it make sense to inherit from `NodeJs` class? This would allow some additional nifty features that may prove useful.

rtfeldman: @BanzaiMan I like it! Added.

.nvmrc
```diff
@@ -0,0 +1 @@
+6.10.0
```
BanzaiMan: What is the rationale for adding this?

rtfeldman: The `NodeJs` class uses this to determine Node version:

```ruby
          def use_nvm_default
            sh.if '-f .nvmrc' do
              sh.echo "Using nodejs version from .nvmrc", ansi: :yellow
              install_version '$(< .nvmrc)'
            end
            sh.else do
              install_version DEFAULT_VERSION
            end
          end
```

`DEFAULT_VERSION` is 0.10, which is too low a version for what we need, so unless I'm misunderstanding, adding `.nvmrc` should override that. 🙂

BanzaiMan: The file is used by the build script `travis-build` generates. Unless you need this file when `travis-build` runs and compiles the bash script, this is not necessary. (Notice that `.nvmrc` does not currently exist for Node.js builds, either.)

rtfeldman: Ahh got it...then how would I specify the Node version?

BanzaiMan: Not sure what you mean by it. `travis-build` does not. Users do, via the content of `.travis.yml`. This will be passed to these processes as `config[:node_js]`

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,113 @@
+module Travis
+  module Build
+    class Script
+      class Elm < Travis::Build::Script::NodeJs
```
BanzaiMan:
```ruby
class Elm < NodeJS
```

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,113 @@
...
+
+          sh.if '! -d sysconfcpus/bin' do
+            install_sysconfcpus
+          end
+        end
+
+        def announce
+          super
+          sh.echo 'elm version:'
+          sh.cmd 'elm --version'
```
BanzaiMan: Pass `echo: true` options, and remove the line above.

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,113 @@
...
+
+          sh.echo 'elm-test version:'
+          sh.cmd 'elm-test --version'
```
BanzaiMan: Same here.

.nvmrc
rtfeldman: By default we're doing several `npm install` commands, which are installing Node packages that require a later version than Node 0.10.

This means that if I don't provide a higher default Node version, by default this build will fail. It will only work if users manually override the Node version to be high enough.

I think it would make more sense to use a high enough Node version by default. Do you think overriding `DEFAULT_VERSION` would be a better way to do that?

lib/travis/build/script.rb
```diff
@@ -18,6 +18,7 @@
 require 'travis/build/script/csharp'
 require 'travis/build/script/d'
 require 'travis/build/script/dart'
+require 'travis/build/script/elm'
```
BanzaiMan: This should appear after `node_js` is `require`d.

spec/build/script/elm_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  describe 'script' do
+    it 'runs "elm-test"' do
+      should include_sexp [:cmd, 'elm-test', echo: true, timing: true]
+    end
+  end
+
+  describe '#cache_slug' do
+    subject { described_class.new(data).cache_slug }
+    it { is_expected.to eq("cache-#{CACHE_SLUG_EXTRAS}--elm-latest") }
```
BanzaiMan: Since `Elm` is now a subclass of `NodeJs`, the cache slug will have `--node` before `--elm`.

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,112 @@
+require 'node_js'
```
BanzaiMan: This is unnecessary.

.nvmrc
BanzaiMan: I think it is reasonable to provide a default version of Node.js for the `elm` builds that is different from `node_js` builds.

Do take care to provide useful information if the user ends up specifying something that doesn't work.

BanzaiMan:
This has been deployed to staging. https://staging.travis-ci.org/

BanzaiMan:
Does it make sense to allow specifying `elm` version (a la `node_js` key for Node.js builds)?

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,110 @@
...
+            config[:elm_version].to_s
+          end
+
+          def elm_test_version
+            config[:elm_test_version].to_s
+          end
+
+          def install_elm
+            npm_install '-g elm-use@">=1.1.0 <2.0.0"'
+            sh.cmd 'elm-use #{elm_version}', retry: true
```
BanzaiMan: This should be: `"elm-use #{elm_version}"`.

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,110 @@
...
+
+            convert_binary_to_sysconfcpus 'elm'
+            convert_binary_to_sysconfcpus 'elm-make'
+            convert_binary_to_sysconfcpus 'elm-package'
+            convert_binary_to_sysconfcpus 'elm-format'
+          end
+
+          def install_elm_test
+            npm_install '-g elm-test@#{elm_test_version}'
```
BanzaiMan: `"-g elm-test@#{elm_test_version}"`

lib/travis/build/script/elm.rb
BanzaiMan: Does `elm-use v1.1.0` (for example), supposed to exit with zero? It does not appear to: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/545661#L798

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,110 @@
...
+            # to work around https://github.com/travis-ci/travis-ci/issues/6656
+            sh.mv "$(npm config get prefix)/bin/#{binary_name}", "$(npm config get prefix)/bin/#{binary_name}-old"
+
+            # Use printf instead of echo here.
+            # see https://github.com/rtfeldman/node-elm-compiler/pull/50
+            sh.cmd ('printf "#\041/bin/bash\n\necho \"Running ' + binary_name + ' with sysconfcpus -n 2\"\n\n$TRAVIS_BUILD_DIR/sysconfcpus/bin/sysconfcpus -n 2 ' + binary_name + '-old \"\$@\"" > $(npm config get prefix)/bin/' + binary_name)
+            sh.chmod '+x', "$(npm config get prefix)/bin/#{binary_name}"
+          end
+
+          def install_sysconfcpus
```
BanzaiMan: These commands generate a lot of output, so I suggest putting it in a fold.

rtfeldman:
> Does it make sense to allow specifying `elm` version (a la `node_js` key for Node.js builds)?

I used `config[:elm_version]` and `config[:elm_test_version]` to attempt that...are other steps necessary?

```ruby
          def elm_version
            config[:elm_version].to_s
          end

          def elm_test_version
            config[:elm_test_version].to_s
          end
```

rtfeldman:
by the way, thanks for all the help @BanzaiMan!

BanzaiMan:
Upon further testing, providing a different default Node.js version from `NodeJs`'s appears to be a little more involved. (Because `nvm install` is invoked in `super`, where `Travis::Build::Script::Elm::DEFAULT_VERSION` has no effect.) This means that users must specify `node_js` in `.travis.yml`.

BanzaiMan:
What I meant was that, does it make sense for users to have in their `.travis.yml`:

```yaml
language: elm
elm:
  - 1.1.0
  - 1.2.0 # or whatever future and past versions that they want to test
```

allowing (with extra work on my part) users to have a build matrix based on the `elm` versions.

I am not sure how tightly `elm-test` versions are coupled with `elm` versions. It may not make sense to have, for example, `elm_test_version` in the configuration.

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,125 @@
...
+            config[:elm_version].to_s
+          end
+
+          def elm_test_version
+            config[:elm_test_version].to_s
+          end
+
+          def install_elm
+            npm_install '-g elm-use@">=1.1.0 <2.0.0"'
+            sh.cmd "elm-use #{elm_version}", retry: true
```
BanzaiMan: This command is still exiting with nonzero. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/545677#L309

If it is expected to be (which seems rather silly, and I would argue that it is bad behavior, but that is another issue entirely), pass `assert: false` and drop `retry: true`.

rtfeldman: `elm-use` 1.1.1 returns `0`; I just upgraded the version range to that. Should be fixed now!

rtfeldman:
> I am not sure how tightly elm-test versions are coupled with elm versions. It may not make sense to have, for example, `elm_test_version` in the configuration.

They are decoupled for now, but in a future release of the Elm platform, `elm-test` will be included in it.

I like this idea; a build matrix would be useful. What if I hardcoded mappings between Elm version and `elm-test` version for now, with the understanding that once they get rolled together that won't be necessary for future releases?

rtfeldman:
@BanzaiMan what did you think of [this idea](https://github.com/travis-ci/travis-build/pull/934#issuecomment-284083694)?

> What if I hardcoded mappings between Elm version and elm-test version for now, with the understanding that once they get rolled together that won't be necessary for future releases?

BanzaiMan:
> What if I hardcoded mappings between Elm version and elm-test version for now, with the understanding that once they get rolled together that won't be necessary for future releases?

I think this is fine.

rtfeldman:
@BanzaiMan I believe https://github.com/travis-ci/travis-build/pull/934/commits/b633a4332f80ec13b01cf7dd35e25c4196cc0477 should address https://github.com/travis-ci/travis-build/pull/934#issuecomment-283972005 in a way where users don't have to manually specify the Node version...unless I'm missing something!

rtfeldman:
Any suggestions on how I should expose https://github.com/travis-ci/travis-build/pull/934#issuecomment-287053155 ? I know what the mappings are, but I'm not sure where to put them in the code such that they'll be recognized by `matrix`.

rtfeldman:
@BanzaiMan Friendly ping re: https://github.com/travis-ci/travis-build/pull/934#issuecomment-288936075 - any ideas?

Also, I view `matrix` as a nice-to-have, and would be happy to revisit that later after getting the current version into a usable state for people. 🙂

knewter:
@BanzaiMan I'd love for this to be merged. Anything holding it up aside from (as I well understand) the overwhelming weight of life itself? :)

BanzaiMan:
Is `elm use latest` supposed to succeed in an empty (for Elm, that is) directory? My test is failing. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/598943#L1098

BanzaiMan:
Hmmm. `elm use latest` should return 0. Not clear yet why it's failing here.

BanzaiMan:
Oh, I see. `elm-use latest` returns `1`. I don't know why it behaves differently, but it does.

```sh-session
travis@b66175e461f1:~$ elm-use latest
elm-use v1.1.1
travis@b66175e461f1:~$ echo $?
1
travis@b66175e461f1:~$ elm use latest
elm-use v1.1.1
travis@b66175e461f1:~$ echo $?
0
```

BanzaiMan:
With a switch to `elm use …`, it goes a little further. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/598941

rtfeldman:
@BanzaiMan Ah yeah, I dropped support for `latest` in favor of explicit versions.

Should be fixed in https://github.com/travis-ci/travis-build/pull/934/commits/7d444528d82596c0f08d97ff6a2f91d5aa49c46a - sorry about that!

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,127 @@
...
+            sh.chmod '+x', "$(npm config get prefix)/bin/#{binary_name}"
+          end
+
+          def install_sysconfcpus
+            sh.fold 'sysconfcpus' do
+              # this is a prerequisite for the convert_binary_to_sysconfcpus method
+              # which provides an epic build time improvement - see https://github.com/elm-lang/elm-compiler/issues/1473#issuecomment-245704142
+              sh.cmd 'git clone https://github.com/obmarg/libsysconfcpus.git', retry: true
+              sh.cd 'libsysconfcpus'
+              sh.cmd './configure --prefix=$TRAVIS_BUILD_DIR/sysconfcpus'
```
BanzaiMan: This runs in the `configure` stage, which precedes `export`, so `$TRAVIS_BUILD_DIR` is not available here. As a result, we'd have `--prefix=/sysconfcpus`, which would cause the subsequent `make install` to fail.

https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/600846#L737

Moving this to `#setup` (https://github.com/travis-ci/travis-build/commit/b6cc481951e3a807d260550926bcccb983424e2c) gets you what you want. See, for example: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/600859#L1034

BanzaiMan:
Also, fix the specs, please.

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,127 @@
...
+
+        def setup
+          super
+
+          install_elm
+          install_elm_test
+        end
+
+        def script
+          sh.cmd 'elm-format --validate .'
```
BanzaiMan: Does this command belong in `script`?

screendriver:
Any news on that @rtfeldman ? The latest update was in July.

lib/travis/build/script/elm.rb
rtfeldman: Hm, I think it makes sense there if I understand the purpose of `script` correctly!

`elm-format --validate .` should fail the build if it returns a nonzero exit code (indicating someone has made a contribution that doesn't pass `elm-format` validation) and it shouldn't bother running tests until after that has been verified.

rtfeldman:
@BanzaiMan Sorry I took so long circling back to the spec failures on this, but fixed now!

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,127 @@
...
+
+          def elm_version
+            config[:elm_version].to_s
+          end
+
+          def elm_test_version
+            config[:elm_test_version].to_s
+          end
+
+          def install_elm
```
BanzaiMan: This portion turns out to be rather chatty (https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/711533#L1026-L1144). I suggest putting it in a fold.

lib/travis/build/script/elm.rb
BanzaiMan: This comment is still valid. See https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/711533#L696.

lib/travis/build/script/elm.rb
BanzaiMan: At the moment, every command in `script` is executed (see https://github.com/travis-ci/travis-ci/issues/1066). So, the intention is to stop the build upon failure of this command, I suggest moving it to `install`.

lib/travis/build/script/elm.rb
rtfeldman: Ah, thanks! Moved in 1b7980728851f5f193b002cca45d07804ffabdef

lib/travis/build/script/elm.rb
rtfeldman: Put behind a fold in b6697ff5bbe8de4ef3a044fe312cdb5e56a9a384

lib/travis/build/script/elm.rb
rtfeldman: Put behind a fold in b6697ff5bbe8de4ef3a044fe312cdb5e56a9a384

lib/travis/build/script/elm.rb
rtfeldman: So I'd rather leave it in `script` because end users might want to override it as part of their test scripts. I changed it to use `&&` in 9e913d1f01d43fa138ce25e988971d1f9261ad64 - based on that thread, I figure this should do the trick, yeah?

reitzig:
Just to express my moral support for this venture, it would be really cool to auto-deploy Elm apps to, say, Github pages. Keep rocking!

rtfeldman:
@BanzaiMan Merge conflicts resolved, and tests passing! 😺

BanzaiMan:
Thanks again! I'll take a look next week.

rtfeldman:
@BanzaiMan awesome! Is there anything else I need to do before merge? Not sure what happens next. 😄

BanzaiMan:
Do we have a list of community maintainers for Elm?

rtfeldman:
@banzaiman yep!

Earlier in this discussion, @stoeffel, @avh4, and @lukewestby volunteered. I know all of them personally. 😄

avh4:
Documenting the community maintainers for Elm here: https://github.com/elm-community/Manifesto/pull/87

rtfeldman:
@BanzaiMan merged `master` in, so merge conflict is now resolved!

Let me know if there are any other steps I need to take on this. Very excited to have it so close to the finish line! 😄

Kurren123:
Any update on this? Would be great to see it merged

carlad:
Hi @Kurren123 , apologies for the delay on this. The relevant developers who can help with this aren't around this week, but I will get someone to take a look at it on Monday. Thanks for you patience.

joshk:
Cross posting my comment on the `travis-yml` repo ... https://github.com/travis-ci/travis-yml/issues/63#issuecomment-408895903

Kurren123:
@avh4 @lukewestby @stoeffel  I'd be happy to add a banner in the same way as they've done for Dart, listing the community supporters you mentioned [here](https://github.com/elm-community/Manifesto/issues/63). As for any bug fixes I'm quite new to travis so it might take me a while.

carlad:
Thanks @Kurren123 
It looks like the errors were a hosts config issue with our staging environment. We've addressed these, so once you've done the banner additions it should be good to go.

carlad:
Hey @rtfeldman I'm still trying to get consensus on how to proceed with this. 
I'm not sure if you've merged this [PR](https://github.com/rtfeldman/travis-build/pull/1) into your branch yet?

rtfeldman:
Ohhh, I misread "After this your fork should be merged as per [this PR](https://github.com/travis-ci/travis-build/pull/934)" as "This [PR] should be merged **after** your PR." 😅

My mistake - it's been merged now!

joshk:
Sorry for my radio silence.

There is only one change we should make before merging this in. Instead of using `elm_version` and `elm_test_version` we should be using `elm` and `elm_test`, which is consistent with what we do with all other languages.

Sorry for only noticing this late in the piece.

BanzaiMan:
@rtfeldman (Or anyone else) let us know if you are ready to have a spin with this. Thanks!

rtfeldman:
@BanzaiMan Thanks for the ping! There's been a new release of Elm since the last round, and `elm-use` has become obsolete, so I made updates accordingly in addition to [the requested change](https://github.com/travis-ci/travis-build/pull/934#issuecomment-418031168). 😃

Let me know if there are any additional changes needed in the updated code. Otherwise I think we're good to go!

lib/travis/build.rb
```diff
@@ -49,6 +49,8 @@ def by_lang(lang)
           Script::ObjectiveC
         when 'bash', 'sh', 'shell', 'minimal' then
           Script::Generic
+        when 'elm' then
```
BanzaiMan: I don't think this `when` clause is necessary, as `elm` should translate to `Script::Elm` naturally.

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,183 @@
+module Travis
+  module Build
+    class Script
+      class Elm < NodeJs
+        # Default NodeJS version to install
+        DEFAULT_NODE_VERSION = '8.9.1'
```
BanzaiMan: Is this the preferred version? v10 has entered the LTS stage, so maybe that's preferable?

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,183 @@
...
+        end
+
+        def script
+          sh.cmd 'elm-format --validate . && elm-test'
+        end
+
+        def setup_cache
+          if data.cache?(:elm)
+            sh.fold 'cache.elm' do
+              # Cache the ~/.elm directory.
```
BanzaiMan: This comment seems a little off. Seems we can remove this entirely.

lib/travis/build/script/elm.rb
rtfeldman: Updated in 74f7059eab2e893a5cea513c06846070403feebe

lib/travis/build/script/elm.rb
rtfeldman: As in, remove the comment, or some of the code too?

I might be misunderstanding something - is the home directory cached by default?

Elm 0.19 creates a persistent cache that it uses a cross builds, which it puts in `~/.elm`. I want to make sure that gets cached across Travis builds...but maybe I don't need to do anything special to make that happen?

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,183 @@
...
+          def elm_test_version
+            config[:elm_test].to_s
+          end
+
+          def elm_format_version
+            config[:elm_format].to_s
+          end
+
+          def install_elm
+            npm_install "-g elm@elm-#{elm_version}"
```
BanzaiMan: Did Elm's numbering scheme change? https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/732882#L824-L828 shows:

```
$ npm install -g elm@elm-0.19.0
npm ERR! code ETARGET
npm ERR! notarget No matching version found for elm@elm-0.19.0
npm ERR! notarget In most cases you or one of your dependencies are requesting
npm ERR! notarget a package version that doesn't exist.
```

BanzaiMan: Dropping `elm-` works: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/732909

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,183 @@
...
+          end
+
+          def install_elm_format
+            npm_install "-g elm-format@elm-#{elm_format_version}"
+
+            convert_binary_to_sysconfcpus 'elm-format'
+          end
+
+          def install_elm_test
+              sh.if "$(vers2int $(echo `node --version` | tr -d 'v')) -lt $(vers2int #{ELM_TEST_REQUIRED_NODE_VERSION})" do
```
BanzaiMan: `vers2int` has been renamed to `travis_vers2int` (to explicit the origin of the function).

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,183 @@
...
+  module Build
+    class Script
+      class Elm < NodeJs
+        # Default NodeJS version to install
+        DEFAULT_NODE_VERSION = '10.13.0'
+
+        DEFAULTS = {
+          elm: '0.19.0',
+          elm_test: '0.19.0',
+          elm_format: '0.19.0'
```
BanzaiMan: The latest version is 0.8.1.

lib/travis/build/script/elm.rb
BanzaiMan: I just noticed that the use of `npm_install` messes up the folds, as `npm_install` itself sets up a fold, and nested folds don't work well. The method should be overridden to remove the fold.

Additionally, it checks whether or not the packages are locked, and invokes different commands. This may not be appropriate for Elm. https://github.com/travis-ci/travis-build/blob/149ff67390be18b128975cc298d4ae4be3d001b2/lib/travis/build/script/node_js.rb#L190-L195

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,183 @@
...
+          sh.echo 'and mention \`@avh4\`, \`@lukewestby\`, \`@stoeffel\` and \`@rtfeldman\`' \
+            ' in the issue', ansi: :green
+
+
+          sh.if '! -d sysconfcpus/bin' do
+            install_sysconfcpus
+          end
+
+          sh.fold 'install.elm' do
+            install_elm
```
BanzaiMan: I suggest adding a colored `echo` command above this, which acts as a label for the `install.elm` fold. (Currently, the fold looks indistinguishable from other lines.)

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,183 @@
...
+            # see https://github.com/rtfeldman/node-elm-compiler/pull/50
+            sh.cmd ('printf "#\041/bin/bash\n\necho \"Running ' + binary_name + ' with sysconfcpus -n 2\"\n\n$TRAVIS_BUILD_DIR/sysconfcpus/bin/sysconfcpus -n 2 ' + binary_name + '-old \"\$@\"" > $(npm config get prefix)/bin/' + binary_name)
+            sh.chmod '+x', "$(npm config get prefix)/bin/#{binary_name}"
+          end
+
+          def install_sysconfcpus
+            sh.fold 'sysconfcpus' do
+              # this is a prerequisite for the convert_binary_to_sysconfcpus method
+              # which provides an epic build time improvement - see https://github.com/elm-lang/elm-compiler/issues/1473#issuecomment-245704142
+              sh.cmd 'git clone https://github.com/obmarg/libsysconfcpus.git', retry: true
```
BanzaiMan: Ditto for this fold.

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,183 @@
...
+        end
+
+        def announce
+          super
+          sh.cmd 'elm --version', echo: true
+          sh.cmd 'elm-test --version', echo: true
+
+          # elm-format doesn't have --version,
+          # but the first line of `elm-format --help` prints the version
+          sh.cmd 'elm-format --help | head -n 1', echo: true
```
BanzaiMan: I also suggest adding a blank line after this one, so that it is visually distinct from the next command (which is typically user-defined `before_install`, or `install`).

lib/travis/build/script/elm.rb
BanzaiMan: See https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/732930 for an example build with the fold fixed.

lib/travis/build/script/elm.rb
rtfeldman: In f53ea8a36a6ee1ab463e482bcd1bdd5e1f182708 I've changed these to use the `elm0.19.0` tag by default, which points to the appropriate version for the 0.19 release of Elm. (For `elm-format`, that's 0.8.1., so `npm install -g elm-format@elm0.19.0` gets you `elm-format` 0.8.1, and that tag will get updated if there are bugfixes on `elm-format` that are still compatible with Elm 0.19.)

lib/travis/build/script/elm.rb
rtfeldman: Updated in 294455217d26c8a01fe8aa979fa5e5a8127ead9c

lib/travis/build/script/elm.rb
rtfeldman: Thanks! Since these all use `npm install -g`, I made a separate method in f53ea8a36a6ee1ab463e482bcd1bdd5e1f182708 which runs that and switched everything to do that.

Also fixed the incorrect npm version strings while I was at it. 👍

lib/travis/build/script/elm.rb
rtfeldman: Added in 17d4a121f4d18126b59a902eac364b8711395d63

lib/travis/build/script/elm.rb
rtfeldman: Added in e386ff219defbe58bbe931cc0b66b57e866d818c

lib/travis/build/script/elm.rb
rtfeldman: Added in e386ff219defbe58bbe931cc0b66b57e866d818c

BanzaiMan:
Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/734314

BanzaiMan:
Thank you, @rtfeldman! It looks very good.

We'll be moving to https://travis-ci.community/ for support issues. We'll tweak the display text accordingly before releasing.

The current plan is to announce Elm support on December 4, 2018.

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,191 @@
...
+            '/new?labels=community:elm', ansi: :green
+          sh.echo 'and mention \`@avh4\`, \`@lukewestby\`, \`@stoeffel\` and \`@rtfeldman\`' \
+            ' in the issue', ansi: :green
+
+
+          sh.if '! -d sysconfcpus/bin' do
+            install_sysconfcpus
+          end
+
+          sh.echo 'Installing elm, elm-test, and elm-format', ansi: :green
```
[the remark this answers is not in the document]
rtfeldman: Ah, thanks! Fixed in 96e46d27e9a75ddec3cda2ed6afc8d0bd12e5d73

lib/travis/build/script/elm.rb
```diff
@@ -0,0 +1,191 @@
...
+            sh.mv "$(npm config get prefix)/bin/#{binary_name}", "$(npm config get prefix)/bin/#{binary_name}-old"
+
+            # Use printf instead of echo here.
+            # see https://github.com/rtfeldman/node-elm-compiler/pull/50
+            sh.cmd ('printf "#\041/bin/bash\n\necho \"Running ' + binary_name + ' with sysconfcpus -n 2\"\n\n$TRAVIS_BUILD_DIR/sysconfcpus/bin/sysconfcpus -n 2 ' + binary_name + '-old \"\$@\"" > $(npm config get prefix)/bin/' + binary_name)
+            sh.chmod '+x', "$(npm config get prefix)/bin/#{binary_name}"
+          end
+
+          def install_sysconfcpus
+            sh.echo 'Installing sysconfcpus', ansi: :green
```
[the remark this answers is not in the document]
rtfeldman: Fixed in 96e46d27e9a75ddec3cda2ed6afc8d0bd12e5d73

BanzaiMan:
Needs docs PR.

rtfeldman:
> Needs docs PR.

Is this something I should do? 😄

BanzaiMan:
@rtfeldman I think I can do this, given the short time frame we've got. I should've pointed this out much earlier.

BanzaiMan:
@rtfeldman What is the intended interactions among `elm`, `elm_test`, and `elm_format`? Currently, these can be independently defined, which I think is OK, but if one changes *only* `elm`, then `elm_test` and `elm_format` fall out of sync with it. That is to say:

```yaml
language: elm
elm: 0.20.0 # a future release
```

will use Elm 0.20.0, but `elm_test` and `elm_format` remains at `elm0.19.0`, which is the default.

I am unfamiliar with the Elm ecosystem, but I imagine that this is not an ideal situation. From the user's perspective, its seems to me that `elm` value should propagate to the other two tools that depend on it (unless, of course, `elm_test` or `elm_format` is overridden.)

BanzaiMan:
On a related note: how useful is the matrix expansion on `elm_test` and `elm_format`? It makes sense for the repository to test on multiple Elm versions, but I would like to not have to expand on `elm_test` and `elm_format`, to prevent enormous build matrices.

BanzaiMan:
@rtfeldman 

We plan to announce this tomorrow (2018-11-27) with the changes I suggested earlier. Namely:

1. The `elm` version propagates to `elm_test` and `elm_format` tag, unless these are explicitly overridden in the configuration.
1. There will be no matrix expansion on `elm_test` and `elm_format`.

BanzaiMan:
@rtfeldman I can work on the changes I've suggested.

BanzaiMan:
See https://github.com/travis-ci/travis-build/commit/9fbe7e4dcacbf5bd82bbee206a0c207fba0f176c, including specs, which shows that:

1. The `elm` value drives `elm_test` and `elm_format` versions.
1. The `elm` value could be with or without the `elm` prefix (e.g., `0.19.0` or `elm0.19.0`) and `TRAVIS_ELM_VERSION` will retain it if the default is used, or the user supplies it; but `TRAVIS_ELM_TEST_VERSION` and `TRAVIS_ELM_FORMAT_VERSION` will always be prefixed with `elm`.

rtfeldman:
Looks good, thanks for all your help @BanzaiMan! Super excited to see this land! 😃

PR #939, opened 2017-01-23
Hide `export` command's STDERR

BanzaiMan:
When the `export` command fails because the statement is malformed (e.g., the value has leading white spaces), STDERR will spill some (or all) of the secrets.

We avoid such risk by sending STDERR of the commands with `secure` option (which is true when we are `export`ing env vars) to `/dev/null`.

### DIFF
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -39,6 +39,7 @@ travis_cmd() {
       --display) display=$2;  shift 2;;
       --retry)   retry=true;  shift ;;
       --timing)  timing=true; shift ;;
+      --secure)  secure=" 2>/dev/null"; shift ;;
       *) break ;;
     esac
   done
@@ -52,9 +53,12 @@ travis_cmd() {
   fi
 
   if [[ -n "$retry" ]]; then
-    travis_retry eval "$cmd"
+    travis_retry eval "$cmd $secure"
   else
-    eval "$cmd"
+    eval "$cmd $secure"
+    if [[ -n $secure && $? -ne 0 ]]; then
+      echo "The previous command failed"
+    fi
   fi
   result=$?
 
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -57,7 +57,9 @@ travis_cmd() {
   else
     eval "$cmd $secure"
     if [[ -n $secure && $? -ne 0 ]]; then
-      echo "The previous command failed"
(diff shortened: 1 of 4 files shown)


BanzaiMan:
This addresses some of https://github.com/travis-ci/travis-ci/issues/1775.

BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/534335#L429-L430

Currently the message is simply `The previous command failed`. Perhaps we can consider a little more descriptive message, say `The previous command failed, probably due to malformed secret. Please be sure to escape special characters.`, or similar.

lib/travis/build/templates/header.sh
```diff
@@ -52,9 +53,12 @@ travis_cmd() {
   fi
 
   if [[ -n "$retry" ]]; then
-    travis_retry eval "$cmd"
+    travis_retry eval "$cmd $secure"
   else
-    eval "$cmd"
+    eval "$cmd $secure"
+    if [[ -n $secure && $? -ne 0 ]]; then
+      echo "The previous command failed"
```
Lyoness: https://github.com/travis-ci/travis-build/pull/939#issuecomment-274646429

` echo "The previous command failed, probably due to malformed secret. Please be sure to escape special characters."` 

does it log spaces as whitespace (for leading whitespace)?

BanzaiMan: The white spaces are special in bash, since they delimit commands and arguments. The leading white spaces here are particularly harmful, since we end up executing something like:

```
export SECRET=    12345678-super-duper-secret
```

The leading numeral makes this an invalid token, which would be dumped to STDERR. We want to avoid that.

Perhaps we can show examples of "special characters": `…special characters such as ' ' and '$'.`.

BanzaiMan:
Also https://github.com/travis-ci/travis-ci/issues/4139.

joshk:
this is very cool!

regarding the message, how about:

`The previous command failed, possibly due to a malformed secure environment variable.`
`Please be sure to escape special characters.`

The special chars example is a good idea, or a link to docs about this.

lib/travis/build/templates/header.sh
Lyoness: yes.  Maybe also a url to docs. Even if not a link, another place for further help.

BanzaiMan:
Tested with the new message: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/534549#L429-L432

```
$ export SECRET=[secure]
The previous command failed, possibly due to a malformed secure environment variable.
Please be sure to escape special characters such as ' ' and '$'.
For more information, see http://www.tldp.org/LDP/abs/html/special-chars.html.
```

PR #940, opened 2017-01-24
Save export's result before printing message

BanzaiMan:
See https://github.com/travis-ci/travis-build/commit/12ac41dfc22346d2d5489540bad04f169fbf66d9#commitcomment-20594568


BanzaiMan:
Since #939 was reverted by #941, this is no longer valid.

PR #942, opened 2017-01-24
Hide `export` command's STDERR

BanzaiMan:
This PR redoes what #939 aimed to do.

In addition to reverting the revert of #939, we heed the advice given in https://github.com/travis-ci/travis-build/pull/939#commitcomment-20594568 and saves off the command's actual result before printing the warning.

### DIFF
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -26,7 +26,7 @@ TRAVIS_TEST_RESULT=
 TRAVIS_CMD=
 
 travis_cmd() {
-  local assert output display retry timing cmd result
+  local assert output display retry timing cmd result secure
 
   cmd=$1
   TRAVIS_CMD=$cmd
@@ -54,15 +54,16 @@ travis_cmd() {
 
   if [[ -n "$retry" ]]; then
     travis_retry eval "$cmd $secure"
+    result=$?
   else
     eval "$cmd $secure"
-    if [[ -n $secure && $? -ne 0 ]]; then
+    result=$?
+    if [[ -n $secure && $result -ne 0 ]]; then
       echo -e "${ANSI_RED}The previous command failed, possibly due to a malformed secure environment variable.${ANSI_CLEAR}
 ${ANSI_RED}Please be sure to escape special characters such as ' ' and '$'.${ANSI_CLEAR}
 ${ANSI_RED}For more information, see http://www.tldp.org/LDP/abs/html/special-chars.html.${ANSI_CLEAR}"
     fi
   fi
-  result=$?
 
   if [[ -n "$timing" ]]; then
     travis_time_finish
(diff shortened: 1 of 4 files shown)


BanzaiMan:
Test: PHP 7 (https://github.com/travis-ci/travis-ci/issues/7213) https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/534659

lib/travis/build/templates/header.sh
```diff
@@ -52,11 +53,17 @@ travis_cmd() {
...
+    travis_retry eval "$cmd $secure"
+    result=$?
   else
-    eval "$cmd"
+    eval "$cmd $secure"
+    result=$?
+    if [[ -n $secure && $result -ne 0 ]]; then
+      echo -e "${ANSI_RED}The previous command failed, possibly due to a malformed secure environment variable.${ANSI_CLEAR}
+${ANSI_RED}Please be sure to escape special characters such as ' ' and '$'.${ANSI_CLEAR}
+${ANSI_RED}For more information, see http://www.tldp.org/LDP/abs/html/special-chars.html.${ANSI_CLEAR}"
```
BanzaiMan: We might want to mention our own documentation as well (or instead) https://docs.travis-ci.com/user/encryption-keys

Maybe https://docs.travis-ci.com/user/encryption-keys should point to the Bash documentation above.

BanzaiMan: https://github.com/travis-ci/docs-travis-ci-com/pull/945

lib/travis/shell/generator/bash.rb
```diff
@@ -144,7 +144,7 @@ def handle_else(cmds)
         private
 
           def handle_secure_vars(key, value, options)
-            if options[:echo] && options.delete(:secure)
+            if options[:echo] && options[:secure]
```
bnferguson: The `delete` is moved here so you can pass along the `--secure` in https://github.com/travis-ci/travis-build/pull/942/files#diff-8c5fd04469601ade1797d06043753110R37 

Do we still need to delete it at another point? Or are we safe to leave the `.delete` off?

lib/travis/build/templates/header.sh
```diff
@@ -52,11 +53,17 @@ travis_cmd() {
   fi
 
   if [[ -n "$retry" ]]; then
-    travis_retry eval "$cmd"
+    travis_retry eval "$cmd $secure"
+    result=$?
```
bnferguson: Nice, fixed the previous code's issue with swallowing the result code. 👍

lib/travis/shell/generator/bash.rb
BanzaiMan: Since `options` is passed around in this class everywhere (instead of it being an instance variable), and `#handle_secure_vars` is called from `#handle_export`, which then just invokes `#handle_cmd` for the call site to call `export`, I think it is OK to leave the `:secure` key-value pair here.

BanzaiMan:
Builds in production showing that it is working:

https://travis-ci.org/BanzaiMan/travis_production_test/builds/197163937 https://travis-ci.org/BanzaiMan/travis_production_test/builds/197164172 https://travis-ci.org/BanzaiMan/travis_production_test/builds/197164491

BanzaiMan:
No reports of problems in production.

PR #946, opened 2017-01-27
Extend support for multiple R versions to trusty environment

jimhester:
Previously this was only available in precise.

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -81,44 +81,31 @@ module Travis
                 # times to work around flaky connection to Launchpad PPAs.
                 sh.cmd 'sudo apt-get update -qq', retry: true
 
-                # Install precompiled R if on precise
-                sh.if '$(lsb_release -cs) == "precise"' do
-                  # Install only the dependencies for an R development environment except for
-                  # libpcre3-dev or r-base-core because they will be included in
-                  # the R binary tarball.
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/r.rb
```diff
@@ -81,44 +81,33 @@ def configure
...
-
-                  r_filename = "R-#{r_version}.xz"
-                  r_url = "https://s3.amazonaws.com/rstudio-travis/R-#{r_version}.xz"
-                  sh.cmd "curl -Lo /tmp/#{r_filename} #{r_url}", retry: true
-                  sh.cmd "tar xJf /tmp/#{r_filename} -C ~"
-                  sh.export 'PATH', "$HOME/R-bin/bin:$PATH", echo: false
-                  sh.export 'LD_LIBRARY_PATH', "$HOME/R-bin/lib:$LD_LIBRARY_PATH", echo: false
-                  sh.rm "/tmp/#{r_filename}"
-                end
+                sh.export 'TRAVIS_OS_VERSION', "$(lsb_release -cs)"
```
BanzaiMan: Hmmm. I am not sure if we should define this variable. `$TRAIVS_*` should be more universally useful. As this is contained in Linux jobs, perhaps use of `$(lsb_release -cs)` is acceptable on line 99.

jimhester:
I removed use of `TRAVIS_OS_VERSION` (4401381), thank you for the review!

PR #948, opened 2017-02-01
Do not update packages

jimhester:
This inadvertently caused an error because the repo was not set when it was called.

I originally added it to prevent errors if recommended packages are out of date (https://github.com/travis-ci/travis-ci/issues/6850#issuecomment-266548181), but that is fairly rare so I think it is better to just remove it for now.

This is fairly urgent to merge and deploy as all R builds are currently broken.

Fixes https://github.com/travis-ci/travis-ci/issues/7252


krlmlr:
Unfortunately, this very often leads to outdated package repos in combination with Travis's cache feature. I'm not sure what's the best solution here.

PR #952, opened 2017-02-03
Introduce TRAVIS_BUILD_AUTH_DISABLED env var for Enterprise

bnferguson:
In Enterprise 2.0 we rely on the behavior where if there is no API token, then authorization is disabled. After a refactor in `travis-build` this non-explicit behavior no longer functioned, breaking the way that workers currently get their build scripts. 

This PR introduces an env var to toggle this behavior back on. ~~At the moment you still also need to not set an API_KEY. If you do set one, this var will be ignored.~~

~~I'm mixed on this, it's truer to the old way but I like to think that when you say "auth disabled" it means it. Getting this out there for discussion but I think I may rework this so that the tests can still pass and we don't have fiddly side effects.~~

[Edit: I did dis] 

Additionally, I think it's worth having a discussion on if having auth disabled is the best thing for Enterprise and if not what the migration for existing workers will be.

### DIFF
--- a/lib/travis/api/build/app.rb
+++ b/lib/travis/api/build/app.rb
@@ -20,8 +20,11 @@ module Travis
         end
 
         configure do
-          use Sentry unless Travis::Build.config.sentry_dsn.empty?
-          use Metriks if librato_disabled? 
+          use Sentry unless Travis::Build.config.sentry_dsn.nil?
+          use Metriks unless Travis::Build.config.librato.email.nil? ||
+                             Travis::Build.config.librato.token.nil? ||
+                             Travis::Build.config.librato.source.nil?
+
           use Rack::Deflater
         end
 
@@ -39,9 +42,7 @@ module Travis
           end
 
           def librato_disabled? 
-            (Travis::Build.config.librato.email.nil? || Travis::Build.config.librato.email.empty?) ||
-            (Travis::Build.config.librato.token.nil? || Travis::Build.config.librato.token.empty?) ||
-            (Travis::Build.config.librato.source.nil? || Travis::Build.config.librato.source.empty?)
+
           end
         end
 
--- a/lib/travis/api/build/app.rb
+++ b/lib/travis/api/build/app.rb
@@ -30,6 +30,7 @@ module Travis
 
         helpers do
           def auth_disabled?
+            Travis::Build.config.auth_disabled? ||
             api_tokens.empty? && (
               settings.development? || settings.test?
             )
--- a/lib/travis/api/build/app.rb
+++ b/lib/travis/api/build/app.rb
@@ -40,10 +40,6 @@ module Travis
             @api_tokens ||=
               Travis::Build.config.api_token.to_s.split(',').map(&:strip)
           end
-
-          def librato_disabled? 
-
-          end
         end
 
         before '/script' do
--- a/lib/travis/api/build/app.rb
+++ b/lib/travis/api/build/app.rb
@@ -20,10 +20,10 @@ module Travis
         end
 
         configure do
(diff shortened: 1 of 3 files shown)


bnferguson:
Made the changes I mentioned. Also added a test. This is ready for review.

lib/travis/api/build/app.rb
```diff
@@ -39,6 +40,10 @@ def api_tokens
             @api_tokens ||=
               Travis::Build.config.api_token.to_s.split(',').map(&:strip)
           end
+
+          def librato_disabled?
```
BanzaiMan: Is this required?

lib/travis/api/build/app.rb
```diff
@@ -20,16 +20,17 @@ class App < Sinatra::Base
...
 
         configure do
-          use Sentry unless Travis::Build.config.sentry_dsn.empty?
-          use Metriks unless Travis::Build.config.librato.email.empty? ||
-                             Travis::Build.config.librato.token.empty? ||
-                             Travis::Build.config.librato.source.empty?
+          use Sentry unless Travis::Build.config.sentry_dsn.nil?
+          use Metriks unless Travis::Build.config.librato.email.nil? ||
+                             Travis::Build.config.librato.token.nil? ||
+                             Travis::Build.config.librato.source.nil?
```
BanzaiMan: Are empty Strings acceptable here?

BanzaiMan:
In general I think this is a good idea to implement, based on the issue description.

lib/travis/api/build/app.rb
bnferguson: ack, it is not. Left over from trying to track down some config issues. So I broke out the above logic, but rolled it back but not completely.

lib/travis/api/build/app.rb
bnferguson: Glad you asked about this, because maybe you have a solution. Really neither nil or empty are acceptable. This line of code never gets used in .com/.org but in Enterprise these get passed as nils when not set. Which results in an error of `empty?` not being available on `nil`. The original code before getting converted to `travis-config` checked if `ENV['TRAVIS_LIBRATO_EMAIL']` was true. Which would cover empty strings but not nils. 

What I really want is `.blank?` but that doesn't seem to be available here (also I'm not sure we set librato source ever in Enterprise. What exactly is that one, that seems to be new).

BanzaiMan: If any of these things is `nil`, could it be that the corresponding key is missing in the config?

If so, I suspect this is a general problem for any key missing in the config. A couple of solutions come to mind.

1. Coerce config values to Strings before calling `#empty?`; e.g., `Travis::Build.config.sentry_dsn.to_s.empty?`
1. Modify `Travis::Build` so that `""` is returned when the key is missing in the config. I think this is a divergence from the general `Hashr` behavior, so this is less preferred.

bnferguson: > If any of these things is nil, could it be that the corresponding key is missing in the config?

Yup, that's generally what's happening here. They're not getting set because the customer doesn't have Librato setup. 



> Coerce config values to Strings before calling `#empty?`; e.g., `Travis::Build.config.sentry_dsn.to_s.empty?`

I think this is probably the most direct approach. Overall I'm not fond of the fact that we don't know if and when we have config and how it could be set to `""` or `nil` but that's a much bigger problem. Normalizing that could be a good maintenance project for `travis-config`, as it feels like there are possible bugs waiting around in projects due to `nil`/`empty` checking. 

Thanks for the review on this stuff!

bnferguson:
Reworked based on the feedback. Tested on my Enterprise server. Looks like it's working now. 👍

PR #953, opened 2017-02-07
Improve "travis compile"'s URL choice.

nex3:
Rather than parsing the current branch's remote, which will often be a
private URL that can't be accessed from a container, this generates the
URL based on the provided slug and the --protocol parameter, which
defaults to "git".

Closes travis-ci/travis-ci#7287


meatballhat:
@nex3 I'm sorry for the long delay! I'm going to close this out given the conflicts, but please holler if you are interested in picking the work up again 💖

PR #954, opened 2017-02-07
Choose a valid branch with "travis compile".

nex3:
This defaults to the current branch, and falls back to not specifying
--branch at all and using whatever the repository default it.

Closes travis-ci/travis-ci#7286

### DIFF
--- a/init.rb
+++ b/init.rb
@@ -5,7 +5,8 @@ module Travis
 
       on('--branch NAME', 'the branch to check out when cloning the repo') {|c, name| c.branch = name}
 
-      attr_accessor :slug, :source_url, :branch
+      attr_accessor :slug, :source_url
+      attr_writer :branch
 
       def setup
         error "run command is not available on #{RUBY_VERSION}" if RUBY_VERSION < '1.9.3'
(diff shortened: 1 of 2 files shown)


init.rb
```diff
@@ -65,10 +67,19 @@ def data
...
               :fetch_timeout => 60,
               :push_timeout => 60
+            },
+            :job => {
+              :branch => choose_branch
             }
           }
         end
 
+        def choose_branch
```
BanzaiMan: The name is a bit off. This should be simply `branch`.

init.rb
```diff
@@ -65,10 +67,19 @@ def data
...
+          return branch if branch
```
BanzaiMan: Consequently, `branch` should not be an `attr_accessor`, but a wrapper to choose a proper value based on `@branch`.

init.rb
nex3: Done.

init.rb
nex3: Done.

meatballhat:
@nex3 Sorry for yet another delay here! Same situation here given conflicts, so please holler if you're interested in picking this back up 💖

PR #961, opened 2017-02-18
Update Go version aliases

tianon:
Fixes https://github.com/travis-ci/gimme/issues/85


lib/travis/build/config.rb
```diff
@@ -53,7 +53,7 @@ def go_version_aliases_hash
         go_version_aliases: ENV.fetch(
           'TRAVIS_BUILD_GO_VERSION_ALIASES', (
             {
-              '1' => '1.7.4',
+              '1' => '1.8',
```
tianon: As a side note, I get why this is up here (sort order), but it'd be easier to not forget to update it if it were down next to `1.x` -- would it be acceptable to make that change the next time I make this sort of PR? :smile: :innocent: (don't want to hold this one up since folks are blocked on getting this one in, it's a pretty superficial change, and I probably won't have cycles again for a few days to update the PR with the change :sweat_smile:)

sagikazarmark:
Might worth checking that the `$TRAVIS_GO_VERSION` env var is actually expanded to 1.8.0, not 1.8.

Not sure how this works, just a note.

See: https://docs.travis-ci.com/user/languages/go#Environment-Variable

BanzaiMan:
Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/542634

shurcooL:
Thank you for fixing this!

sagikazarmark:
As I suspected (and specifically mentioned) the `$TRAVIS_GO_VERSION` env var is either not populated correctly or the [documentation](https://docs.travis-ci.com/user/languages/go#Environment-Variable) is not specific enough about what we should expect there.

https://travis-ci.org/deshboard/boilerplate-service/jobs/203277254#L369

shurcooL:
As I understand, by "real Go version" it means whatever `go version` prints. Which is "1.8" for current version. The regexp example is definitely misleading.

BanzaiMan:
We welcome your suggestions on the documentation at https://github.com/travis-ci/docs-travis-ci-com. Thanks!

PR #967, opened 2017-02-22
When `eval`ing secure commands, send STDERR to /dev/null

BanzaiMan:
In cases where `eval` is passed an invalid statement, it will
dump the entire string to STDERR, causing leak.

These messages should be suppressed if `$secure` is set

Resolves https://github.com/travis-ci/travis-ci/issues/7359.


BanzaiMan:
Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/543382 (it uses the same secret as one reported in https://github.com/travis-ci/travis-ci/issues/7359.

PR #968, opened 2017-02-22
Fixing the regex for csharp

blackdwarf:
Fixes the regex for csharp to allow `-rc*` builds to also be available. 

/cc @BanzaiMan @akoeplinger


Joshua-Anderson:
Tested the regex changes and LGTM. Thanks for foxing this up!

blackdwarf:
@Joshua-Anderson not a problem, glad to help. I cannot tell why the build is failing, though...seems like some Docker issue?

BanzaiMan:
Currently blocked by https://github.com/travis-ci/travis-ci/issues/7367. I've tried a few restarts, but it's still having problems. Will try again in a few hours.

PR #969, opened 2017-02-23
Revamp Dart support.

nex3:
This adds:

* Support for running different test commands in different builds.
* Support for running any browser in XVFB on Linux.
* Support for the Dart static analyzer.
* Support for the Dart code formatter.

### DIFF
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -147,10 +147,8 @@ MESSAGE
         end
 
         def install
-          sh.fold 'pub_get' do
-            sh.echo 'pub get'
-
-            sh.if '-f pubspec.yaml' do
+          sh.if '-f pubspec.yaml' do
+            sh.fold 'pub_get' do
               sh.cmd "pub get"
             end
           end
@@ -174,7 +172,7 @@ MESSAGE
             sh.fold "deprecated.unittest" do
               sh.deprecate <<MESSAGE
 DEPRECATED: The unittest package is deprecated. Please upgrade to the test
-package. See https://github.com/dart-lang/test.
+package. See https://github.com/dart-lang/test#readme.
 MESSAGE
             end
 
@@ -242,10 +240,10 @@ MESSAGE
             urlEnd = ''
             # support of "dev" or "stable"
             if ["stable", "dev"].include?(task[:dart])
-              urlEnd = "#{config[:dart]}/release/latest"
+              urlEnd = "#{task[:dart]}/release/latest"
             # support of "stable/release/1.15.0" or "be/raw/110749"
             elsif task[:dart].include?("/")
-              urlEnd = config[:dart]
+              urlEnd = task[:dart]
             # support of dev versions like "1.16.0-dev.2.0" or "1.16.0-dev.2.0"
             elsif task[:dart].include?("-dev")
               urlEnd = "dev/release/#{task[:dart]}"
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -86,6 +86,8 @@ MESSAGE
           sh.echo 'and mention `@nex3` and \`@a14n\`' \
             ' in the issue', ansi: :green
 
+          sh.export 'PUB_ENVIRONMENT', 'travis'
+
           sh.fold 'dart_install' do
             sh.echo 'Installing Dart', ansi: :yellow
             sh.cmd "curl #{archive_url}/sdk/dartsdk-#{os}-x64-release.zip > $HOME/dartsdk.zip"
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -4,13 +4,53 @@ module Travis
       class Dart < Script
         DEFAULTS = {
           dart: 'stable',
-          with_content_shell: 'false'
+          with_content_shell: false,
+          install_dartium: false,
+          xvfb: true
         }
+
+        attr_reader :task
+
+        def initialize(*args)
+          super
+
+          @task = config[:dart_tasks] || {}
+          @task = {task.to_sym => true} if task.is_a?(String)
+          @task[:install_dartium] = config[:install_dartium] unless task.include?(:install_dartium)
+          @task[:xvfb] = config[:xvfb] unless task.include?(:xvfb)
+          @task[:dart] ||= config[:dart]
+
+          # Run "pub run test" by default if no other tasks are specified.
+          @task[:test] ||= true if !@task[:dartanalyzer] && !@task[:dartfmt]
+        end
 
         def configure
           super
 
-          if with_content_shell
+          if config[:with_content_shell]
+            if config.include?(:dart_tasks)
+              sh.failure "with_content_shell can't be used with dart_tasks."
+              return
+            elsif config[:install_dartium]
+              sh.failure "with_content_shell can't be used with install_dartium."
+              return
+            elsif !config[:xvfb]
+              sh.failure "with_content_shell can't be used with xvfb."
+              return
+            end
+
+            sh.fold "deprecated.with_content_shell" do
+              sh.deprecate <<MESSAGE
+DEPRECATED: with_content_shell is deprecated. Instead use:
+
+    dart_tasks:
+    - test: --platform vm
+    - test: --platform firefox
+    - test: --platform dartium
+      install_dartium: true
+MESSAGE
+            end
+
             sh.fold 'content_shell_dependencies_install' do
               sh.echo 'Installing Content Shell dependencies', ansi: :yellow
 
@@ -31,7 +71,7 @@ module Travis
         def export
           super
 
-          sh.export 'TRAVIS_DART_VERSION', config[:dart], echo: false
+          sh.export 'TRAVIS_DART_VERSION', task[:dart], echo: false
         end
 
         def setup
@@ -43,16 +83,12 @@ module Travis
             ansi: :green
           sh.echo '  https://github.com/travis-ci/travis-ci/issues' \
             '/new?labels=community:dart', ansi: :green
-          sh.echo 'and mention \`@a14n\`, \`@devoncarew\` and \`@sethladd\`' \
+          sh.echo 'and mention `@nex3` and \`@a14n\`' \
             ' in the issue', ansi: :green
 
           sh.fold 'dart_install' do
             sh.echo 'Installing Dart', ansi: :yellow
-            os = case config[:os]
-            when 'linux' then 'linux-x64'
-            when 'osx' then 'macos-x64'
-            end
-            sh.cmd "curl #{archive_url}/sdk/dartsdk-#{os}-release.zip > $HOME/dartsdk.zip"
+            sh.cmd "curl #{archive_url}/sdk/dartsdk-#{os}-x64-release.zip > $HOME/dartsdk.zip"
             sh.cmd "unzip $HOME/dartsdk.zip -d $HOME > /dev/null"
             sh.cmd "rm $HOME/dartsdk.zip"
             sh.cmd 'export DART_SDK="$HOME/dart-sdk"'
@@ -60,7 +96,31 @@ module Travis
             sh.cmd 'export PATH="$HOME/.pub-cache/bin:$PATH"'
           end
 
-          if with_content_shell
+          if task[:install_dartium]
+            sh.fold 'dartium_install' do
+              sh.echo 'Installing Dartium', anis: :yellow
+
+              sh.cmd "mkdir $HOME/dartium"
+              sh.cmd "cd $HOME/dartium"
+              sh.cmd "curl #{archive_url}/dartium/dartium-#{os}-x64-release.zip > dartium.zip"
+              sh.cmd "unzip dartium.zip > /dev/null"
+              sh.cmd "rm dartium.zip"
+              sh.cmd 'dartium_dir="${PWD%/}/$(ls)"'
+
+              # The executable has to be named "dartium" in order for the test
+              # runner to find it.
+              if os == 'macos'
+                sh.cmd 'ln -s "$dartium_dir/Chromium.app/Contents/MacOS/Chromium" dartium'
+              else
+                sh.cmd 'ln -s "$dartium_dir/chrome" dartium'
+              end
+
+              sh.cmd 'export PATH="$PWD:$PATH"'
+              sh.cmd "cd -"
+            end
+          end
+
+          if config[:with_content_shell]
             if config[:os] != 'linux'
               sh.failure 'Content shell only supported on Linux'
             end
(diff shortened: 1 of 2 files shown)


nex3:
See also travis-ci/docs-travis-ci-com#980.

kevmoo:
@BanzaiMan We'd love to get this through ASAP. If you can help...

Thanks!

lib/travis/build/script/dart.rb
```diff
@@ -87,62 +147,114 @@ def announce
...
+          def os
+            config[:os] == 'osx' ? 'macos' : 'linux'
+          end
+
           def archive_url
             urlEnd = ''
             # support of "dev" or "stable"
-            if ["stable", "dev"].include?(config[:dart])
+            if ["stable", "dev"].include?(task[:dart])
               urlEnd = "#{config[:dart]}/release/latest"
```
a14n: `config[:dart]` -> `task[:dart]` ?

lib/travis/build/script/dart.rb
```diff
@@ -87,62 +147,114 @@ def announce
...
             # support of "stable/release/1.15.0" or "be/raw/110749"
-            elsif config[:dart].include?("/")
+            elsif task[:dart].include?("/")
               urlEnd = config[:dart]
```
a14n: `config[:dart]` -> `task[:dart]` ?

lib/travis/build/script/dart.rb
```diff
@@ -87,62 +147,114 @@ def announce
         end
 
         def install
-          sh.if '-f pubspec.yaml' do
-            sh.cmd "pub get"
+          sh.fold 'pub_get' do
+            sh.echo 'pub get'
+
+            sh.if '-f pubspec.yaml' do
```
nex3: Done.

lib/travis/build/script/dart.rb
```diff
@@ -87,62 +147,114 @@ def announce
...
+              pub_run_test
             end
           end
           # tests with test_runner for old tests written with unittest package
-          sh.elif '[[ -d packages/unittest ]] || grep -q ^unittest: .packages 2> /dev/null', raw: true do
+          sh.elif package_installed?('unittest'), raw: true do
+            sh.fold "deprecated.unittest" do
+              sh.deprecate <<MESSAGE
+DEPRECATED: The unittest package is deprecated. Please upgrade to the test
+package. See https://github.com/dart-lang/test.
```
nex3: I think the GitHub version is a lot more readable. I've updated the link to anchor to the README, though.

lib/travis/build/script/dart.rb
nex3: Done.

lib/travis/build/script/dart.rb
nex3: Done.

kevmoo:
@cotsog @BanzaiMan Any help with this? We'd love to roll it out ASAP.

Thanks!

BanzaiMan:
Sorry about the delay. I was traveling.

I plan to deploy this to staging around 15:00 UTC 2017-03-02.

kevmoo:
Wonderful! Thanks.

a14n:
@BanzaiMan : Could you confirm that another PR should be done on https://github.com/travis-ci/travis-core to enable expansion for `dart_tasks` key? See https://github.com/travis-ci/docs-travis-ci-com/pull/980/files#diff-fba593cd6adfc02e488944ce868fcdd5R41 for the documentation.

Something like the following diff on https://github.com/travis-ci/travis-core/blob/master/lib/travis/model/build/config.rb#L61

```
-      'dart'        => [:dart],
+      'dart'        => [:dart, :dart_tasks],

**Comment by BanzaiMan** [comment]
@a14n Matrix expansion is not handled differently (in an app without `travis-core`), so no PR would be necessary (I'd have to do this). This change has not be publicly documented.

How is `dart_tasks` expected to be used? The examples in the deprecation warning seems a bit terse to me, though it may make complete sense to seasoned Dart programmers.

**Comment by nex3** [comment]
@BanzaiMan There's full documentation in travis-ci/docs-travis-ci-com#980, but the broad idea is that each task defines a separate job that either runs tests, runs the static analyzer, or verifies that the code is formatted properly. So it should be a matrix key.

**Comment by BanzaiMan** [comment]
As `dart_tasks` (as currently named) is meant to be a matrix dimension, I think the correct name is `dart_task`, as it will contain a scalar value when it is passed to `travis-build`.

**Comment by BanzaiMan** [comment]
Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/545209

File: lib/travis/build/script/dart.rb [hunk 5]
```diff
@@ -43,24 +83,46 @@ def setup
             ansi: :green
           sh.echo '  https://github.com/travis-ci/travis-ci/issues' \
             '/new?labels=community:dart', ansi: :green
-          sh.echo 'and mention \`@a14n\`, \`@devoncarew\` and \`@sethladd\`' \
+          sh.echo 'and mention `@nex3` and \`@a14n\`' \
```
a14n: `\` are missing and leads to [the following error in the job](https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/545217#L119).

>/home/travis/build.sh: line 440: @nex3: command not found

lib/travis/build/script/dart.rb
```diff
@@ -88,61 +150,111 @@ def announce
...
               sh.cmd 't=0; until (xdpyinfo -display :99 &> /dev/null || test $t -gt 10); do sleep 1; let t=$t+1; done'
               sh.cmd 'pub run test -p vm -p content-shell -p firefox'
             else
-              sh.cmd 'pub run test'
+              pub_run_test
             end
           end
           # tests with test_runner for old tests written with unittest package
-          sh.elif '[[ -d packages/unittest ]] || grep -q ^unittest: .packages 2> /dev/null', raw: true do
+          sh.elif package_installed?('unittest'), raw: true do
```
a14n: There seems to be an issue here (on my build on staging https://staging.travis-ci.org/a14n/dart-rational/jobs/545227#L233 ) :

```
/home/travis/build.sh: line 459: syntax error near unexpected token `elif'
/home/travis/build.sh: line 459: `elif [[ -d packages/unittest ]] || grep -q ^unittest: .packages 2> /dev/null; then'
```

a14n:
From another job for `"test": "--platform chrome"` https://staging.travis-ci.org/a14n/dart-rational/jobs/545230#L235 Chrome is not available (see travis-ci/travis-ci#6683).

a14n:
@BanzaiMan from a UI point of view I'd like to see a column by expansion key (here `dart_tasks`) on the table of all jobs https://staging.travis-ci.org/a14n/dart-rational/builds/545226 . For now, we have to go into each job and click on _config_ to see the _dart_tasks_ value used for the job.

BanzaiMan:
@a14n That's a PR for `travis-web`.

BanzaiMan:
As for Chrome… Not sure if we have it on Precise. (If not, it is unlikely that it will be added.)

Is it possible to detect this and act accordingly?

lib/travis/build/script/dart.rb
```diff
@@ -88,61 +150,111 @@ def announce
...
           end
+
+          dartanalyzer
+          dartfmt
         end
 
         private
 
+          def pub_run_test
+            return unless (args = task[:test])
```
BanzaiMan: When this `return` happens, there will be no statement returned.

This causes the error mentioned in https://github.com/travis-ci/travis-build/pull/969/files/df3283cf981efb229e6fc3e7d695ab37acd5c046#r103981482:

```bash
if [[ -d packages/test ]] || grep -q ^test: .packages 2> /dev/null; then
elif [[ -d packages/unittest ]] || grep -q ^unittest: .packages 2> /dev/null; then
⋮
fi
```

So, something needs to be returned. For example, `sh.raw ":"`.

lib/travis/build/script/dart.rb
```diff
@@ -88,61 +150,111 @@ def announce
...
+          def package_installed?(package)
+            "[[ -d packages/#{package} ]] || grep -q ^#{package}: .packages 2> /dev/null"
+          end
+
+          def os
+            config[:os] == 'osx' ? 'macos' : 'linux'
+          end
+
           def archive_url
             urlEnd = ''
```
BanzaiMan: I didn't notice this until now, but can we move to snake case, please? (Did I miss any other?)

lib/travis/build/script/dart.rb
BanzaiMan: Assigning `args` inside `unless` seems a bit awkward. Please split these two things.

lib/travis/build/script/dart.rb
```diff
@@ -88,61 +150,111 @@ def announce
...
+            args = args.is_a?(String) ? " #{args}" : ""
+
+            # Mac OS doesn't need or support xvfb-run.
+            xvfb_run = 'xvfb-run -s "-screen 0 1024x768x24" '
+            xvfb_run = '' if task[:xvfb] == false || os == "macos"
+            sh.cmd "#{xvfb_run}pub run test#{args}"
+          end
+
+          def dartanalyzer
+            return unless (args = task[:dartanalyzer])
```
BanzaiMan: Same as https://github.com/travis-ci/travis-build/pull/969/files#r103991442

lib/travis/build/script/dart.rb
```diff
@@ -88,61 +150,111 @@ def announce
...
+            args = '.' unless args.is_a?(String)
+            sh.cmd "dartanalyzer #{args}"
+          end
+
+          def dartfmt
+            return unless (args = task[:dartfmt])
```
BanzaiMan: Same as https://github.com/travis-ci/travis-build/pull/969/files#r103991442

lib/travis/build/script/dart.rb
nex3: Done.

lib/travis/build/script/dart.rb
nex3: Done.

lib/travis/build/script/dart.rb
nex3: Done.

lib/travis/build/script/dart.rb
nex3: Done.

lib/travis/build/script/dart.rb
nex3: Done.

BanzaiMan:
@nex3 I see what you are saying about `dart_tasks`. However, the use of the singular `dart_task` is consistent with other matrix dimension keys (`rvm`, `python`, `env`, `otp_release`, `gemfile`, …).

nex3:
Done. The documentation will need to be updated as well.

BanzaiMan:
@nex3 A good point. Do you want to open a new PR, or can I just revert the revert and change `dart_tasks` to `dart_task`? In other words, do you have any other edit in mind for the docs?

lib/travis/build/script/dart.rb
```diff
@@ -88,60 +150,124 @@ def announce
...
+          dartfmt
         end
 
         private
 
+          def pub_run_test
+            args = task[:test]
+            unless args
+              sh.raw ':'
+              return
```
BanzaiMan: This still returns `nil`. We need something like:

```ruby
return sh.raw ':'
```

nex3: Why does the return value matter? Aren't all the `sh` commands side-effect-based?

BanzaiMan: Methods on `sh` (e.g., `sh.raw`, `sh.if`, etc.) return S expressions, which will be used to compile the bash script when `Travis::Build::Script#compile` runs. Having this method return `nil` means that under certain circumstances the build script will have no statement inside the `if` (or `else` or `elif`) clause, which is not allowed in bash.

I alluded to this (but not explicitly stated it) in an earlier comment. https://github.com/travis-ci/travis-build/pull/969#discussion_r103991442

nex3: Doesn't it [insert the s-expression into a list](https://github.com/travis-ci/travis-build/blob/master/lib/travis/shell/builder.rb#L38) rather than returning it?

BanzaiMan: Ruby's method return value is the last statement value, which, in this case is an `Array`, returned by `Array#insert`.

nex3: Right, but as far as I can tell that return value isn't actually used anywhere. The `Cmds` class, where the block passed to `sh.if` is ultimately used, [doesn't touch the return value at all](https://github.com/travis-ci/travis-build/blob/master/lib/travis/shell/ast.rb#L29)—it just sets itself as the recipient of the array of nodes.

BanzaiMan: The S expression (the `Cmd` object) will eventually find its way here: https://github.com/travis-ci/travis-build/blob/b2abcffaa7a73cfe5e30bbf585ea7d718456f1d6/lib/travis/shell/generator.rb#L15 and will be a part of the S expression that generates the bash script.

BanzaiMan: After a good night's sleep, I see that this should work, because the S expression is altered before `return`.

I apologize for the noise.

BanzaiMan:
This has been deployed to staging, though it may come and go, as we work on staging.

a14n:
@nex3 my test is ok except for [`"test": "--platform chrome"`](https://staging.travis-ci.org/a14n/dart-rational/jobs/547407) that should have been marked as _failed_ (not _passed_ because Chrome is not installed) because the tests failed. Don't know if the issue is in `test` package or the travis script.

BanzaiMan:
@a14n Do you have a build log to show this problem? I can take a look.

a14n:
The test failed and [pub run test --platform chrome](https://staging.travis-ci.org/a14n/dart-rational/jobs/547407#L231) should have returned an exit code != 0 but [The command "xvfb-run -s "-screen 0 1024x768x24" pub run test --platform chrome" exited with 0.](https://staging.travis-ci.org/a14n/dart-rational/jobs/547407#L255)

BanzaiMan:
Hmmm. That is indeed strange. Is it possible to run `pub run …` outside of xvfb?

a14n:
I made the test locally with a platform not install:

```
$ pub run test --platform phantomjs
00:05 +0 -1: compiling test/bigint_test.dart                                                       
  Failed to load "test/bigint_test.dart":
  Failed to run PhantomJS: No such file or directory
    Command: phantomjs /tmp/dart_test_en0OPX/script.js http://localhost:37662/gIA%2F%2FB5qoZ9cVjvTiJctyc%2FRS04hLlHo/packages/test/src/runner/browser/static/index.html?managerUrl=ws%3A%2F%2Flocalhost%3A37662%2FgIA%252F%252FB5qoZ9cVjvTiJctyc%252FRS04hLlHo%2F0.
  dart:async                                                     _SyncCompleter.complete
  package:package_resolver/src/current_isolate_resolver.dart 43  CurrentIsolateResolver.asSync.asSync_async_body>
  
00:06 +0 -2: compiling test/rational_test.dart                                                     
  Failed to load "test/rational_test.dart":
  Failed to run PhantomJS: No such file or directory
    Command: phantomjs /tmp/dart_test_en0OPX/script.js http://localhost:37662/gIA%2F%2FB5qoZ9cVjvTiJctyc%2FRS04hLlHo/packages/test/src/runner/browser/static/index.html?managerUrl=ws%3A%2F%2Flocalhost%3A37662%2FgIA%252F%252FB5qoZ9cVjvTiJctyc%252FRS04hLlHo%2F0.
  dart:async                                                     _SyncCompleter.complete
  package:package_resolver/src/current_isolate_resolver.dart 43  CurrentIsolateResolver.asSync.asSync_async_body>
  
00:06 +0 -2: Some tests failed.                                                                    

$ echo $?
1

```

You can see the exit code is `1`.

BanzaiMan:
🤔  It is not clear to me why this is so at the moment. I'll take another look later. Perhaps I'll have a moment of truth.

BanzaiMan:
Ah. I understand this now. There are a couple of empty `:` statements (which I asked @nex3 to add) between `xvfb` invocation and the command result checking, and that is resetting the result instead.

BanzaiMan:
Namely, `#dartanalyzer` and `#dartfmt` are executed no matter what the configuration. (`#pub_run_test` in some cases.)

BanzaiMan:
I think https://github.com/travis-ci/travis-build/commit/f5c77be11e4765876e28b4ddd944b6af7a574913 resolves this issue. This has been deployed to staging. Please test it.

a14n:
@BanzaiMan there're some issue in the generated script. See https://staging.travis-ci.org/a14n/dart-rational/builds/547794

a14n:
the script error : https://staging.travis-ci.org/a14n/dart-rational/jobs/547795#L231 

The issue of exit code looks fixed : https://staging.travis-ci.org/a14n/dart-rational/jobs/547798#L255

BanzaiMan:
Ah, `#pub_run_test` is different. Please rerun the ones that yielded the bash syntax error.

a14n:
[This new build](https://staging.travis-ci.org/a14n/dart-rational/builds/547814) works as intended ! Thanks !

BanzaiMan:
Is https://staging.travis-ci.org/a14n/dart-rational/jobs/547820#L233 correct? The build is exiting immediately, without any further execution. I'll make another comment on the code.

lib/travis/build/script/dart.rb
```diff
@@ -88,60 +150,124 @@ def announce
...
+
+            sh.if package_installed?('dart_style'), raw: true do
+              sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
+            end
+
+            sh.cmd 'unformatted=`dartfmt -n .`'
+            sh.if '! -z "$unformatted"' do
+              sh.echo "Files are unformatted:", ansi: :red
+              sh.echo "$unformatted"
+              sh.failure ""
```
BanzaiMan: I think you want `sh.terminate` here.

lib/travis/build/script/dart.rb
```diff
@@ -4,13 +4,53 @@ class Script
...
+          @task[:test] ||= true if !@task[:dartanalyzer] && !@task[:dartfmt]
+        end
+
         def configure
           super
 
-          if with_content_shell
+          if config[:with_content_shell]
+            if config.include?(:dart_task)
+              sh.failure "with_content_shell can't be used with dart_task."
```
BanzaiMan: Here, and other places where `sh.failure` is used, I think you mean `sh.terminate`, so that the exit status can be shown in the build logs.

BanzaiMan: Do keep in mind, though, that `sh.terminate` will also terminate the build immediately, without other phases (say, `after_failure`, `after_script`) running.

BanzaiMan:
Suggested changes: https://github.com/travis-ci/travis-build/compare/268f6de...ha-test-pr-969

a14n:
> Is https://staging.travis-ci.org/a14n/dart-rational/jobs/547820#L233 correct?

Yes it looks good to me

BanzaiMan:
I feel that not running `after_script`, `after_failure` is wrong.

If you insist on exiting immediately, that should be noted as an exception (on Dart builds) to the standard Travis build behavior in the docs.

lib/travis/build/script/dart.rb
nex3: I agree that `after_failure` and `after_script` should probably run, but in that case how should I emit errors like this?

nex3:
Added suggested changes.

BanzaiMan:
@nex3 I think you missed this commit https://github.com/travis-ci/travis-build/commit/dd662b07c952eea6ebca054e5d568b0c1c455fc4. Then we can test it in staging once again.

Fix the specs, and then we can merge and deploy.

BanzaiMan:
I can merge this, and make additional changes if both of you are busy with other things. If I don't get to deploying today, it'll be carried over to next week (Tuesday or later).

kevmoo:
@BanzaiMan @nex3 won't be available for a bit. If you're okay cleaning this up, please merge.

I'm happy to coordinate with @a14n if there is further work needed.

Thanks!

BanzaiMan:
Spec failures seem a little messy. I quite reason about the intents quickly, so I'd appreciate help here.

kevmoo:
@a14n might could you take a look?

a14n:
I can perhaps have a look (if I have time) but I cannot commit in Natalie branch to fix this PR

kevmoo:
@a14n fork the branch and create another PR? Or just point @BanzaiMan to your branch.

If you can...thanks so much!

a14n:
[Build fixed](https://travis-ci.org/a14n/travis-build/builds/212269239) in https://github.com/a14n/travis-build/tree/dart-revamp ([2 commits added](https://github.com/a14n/travis-build/commits/dart-revamp) compared to this PR).

@BanzaiMan let me know if you prefer to change the origin branch of this PR or close the PR in favor of another one.

kevmoo:
@a14n wanna send a parallel pull request (maybe marked as WIP, etc) just to see if it makes the bots green

@BanzaiMan how would you like to proceed?

BanzaiMan:
@kevmoo @a14n We can't change the head of this PR, so I reckon closing this and opening a new one might be best.

Thank you!

kevmoo:
@a14n – please open a parallel PR w/ your fix and reference this one.

If/when we land that one we can close this out.

Sound good?

a14n:
new PR #1001

BanzaiMan:
Closing.

PR #970, opened 2017-03-01
Use correct host name when writing ~/.netrc

BanzaiMan:
Use the hostname given in the payload

### DIFF
--- a/lib/travis/build/git/clone.rb
+++ b/lib/travis/build/git/clone.rb
@@ -79,7 +79,7 @@ module Travis
             sh.newline
             sh.echo "Using $HOME/.netrc to clone repository.", ansi: :yellow
             sh.newline
-            sh.raw "echo -e \"machine github.com\n  login #{data.token}\\n\" > $HOME/.netrc"
+            sh.raw "echo -e \"machine #{source_host_name}\n  login #{data.token}\\n\" > $HOME/.netrc"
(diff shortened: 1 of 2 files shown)


BanzaiMan:
See https://github.com/travis-ci/travis-build/pull/804/files#r102742878 for motivation.

lib/travis/build/git/clone.rb
```diff
@@ -88,14 +88,19 @@ def delete_netrc
           end
 
           def github?
+            source_host_name.downcase == 'github.com' || source_host_name.downcase.end_with?('.github.com')
```
svenfuchs: could this just be `source_host_name.downcase.end_with?('github.com')`? or `source_host_name.downcase =~ /\bgithub\.com/`

BanzaiMan: Neither seems to indicate the intent clear. `source_host_name.downcase.end_with?('github.com')` is incorrect, as `evilgithub.com` would match. I want the last 2 levels of `source_host_name` to match `github` and `com`, _*exactly*_.

Maybe `source_host_name.split('.')[-2,2] == ['github', 'com']`, but that seems rather unsightly.

svenfuchs: @BanzaiMan right, ignore me :)

PR #971, opened 2017-03-06
Add Rakefile for updating static files

BanzaiMan:
This PR adds `Rakefile`, which defines Rake tasks to update static files used in various builds; currently, `casher`, `nvm`, and `sbt`.


BanzaiMan:
Tested in https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/547087#L129-L130.

BanzaiMan:
@meatballhat has the brilliant idea to execute this Rake task at the `travis-build` app deployment time, so that these static files don't have to be stored in the git repository.

meatballhat:
✨ 🎩 

``` ruby
task 'assets:precompile' => :update_static_files
```

Rakefile
```diff
@@ -0,0 +1,84 @@
...
+task :nvm do
+  latest_tag = latest_tag_for 'creationix/nvm'
+  node_js_rb_path = File.expand_path('../lib/travis/build/script/node_js.rb', __FILE__)
+
+  logger.info "Latest nvm release is #{latest_tag}"
+  sed_cmd = %Q(sed -i "s,^\\(.*\\)NVM_VERSION\s*=.*$,\\1NVM_VERSION = #{latest_tag.gsub(/^v/,'')}," #{node_js_rb_path})
+
+  logger.info "Updating #{node_js_rb_path}"
+  `#{sed_cmd}`
+  fetch_githubusercontent_file "creationix/nvm/#{latest_tag}/nvm.sh"
```
ljharb: could you add `fetch_githubusercontent_file "creationix/nvm/#{latest_tag}/nvm-exec"` as well? https://github.com/creationix/nvm/blob/master/nvm-exec

README.md
```diff
@@ -12,7 +12,12 @@ form.
 
 ## Running test suites
 
-Run `bundle exec rspec spec`.
+Run
+
+```
+bundle exec rake update_static_files
+bundle exec rspec spec
```
meatballhat: Maybe combine these such that there's a task in the `Rakefile` for:

``` ruby
task default: %i(update_static_files spec)
```

Rakefile
```diff
@@ -0,0 +1,114 @@
...
+  to = File.join("..", public_files_dir, to)
+
+  FileUtils.mkdir_p public_files_dir
+
+  response = conn.get do |req|
+    logger.info "Fetching #{conn.url_prefix.to_s}#{from}"
+    req.url from
+  end
+
+
```
meatballhat: extra blank line?

Rakefile
```diff
@@ -0,0 +1,114 @@
...
+    File.write(dest, response.body)
+    logger.info "Setting mode 'a+rx' on #{dest}"
+    FileUtils.chmod "a+rx", dest
+    files << dest
+  else
+    fail "Could not fetch #{from}"
+  end
+end
+
+
```
meatballhat: extra blank line?

Rakefile
```diff
@@ -0,0 +1,114 @@
...
+  conn = Faraday.new('https://api.github.com')
+
+  response = conn.get do |req|
+    releases_url = "repos/#{repo}/releases"
+    logger.info "Fetching releases from #{conn.url_prefix.to_s}#{releases_url}"
+    req.url releases_url
+    oauth_token = ENV['GITHUB_OAUTH_TOKEN']
+    if oauth_token
+      req.headers['Authorization'] = "token #{oauth_token}"
+    end
```
meatballhat: Is there a case where it's OK for the `Authorization` header to be missing?  I would instead expect to see, e.g.:

``` ruby
    req.headers['Authorization'] = "token #{ENV.fetch('GITHUB_OAUTH_TOKEN')}"
```

Rakefile
```diff
@@ -0,0 +1,114 @@
...
+desc 'update nvm.sh'
+task :nvm do
+  latest_release = latest_release_for 'creationix/nvm'
+  node_js_rb_path = File.expand_path('../lib/travis/build/script/node_js.rb', __FILE__)
+
+  logger.info "Latest nvm release is #{latest_release}"
+  sed_cmd = %Q(sed -i "s,^\\(.*\\)NVM_VERSION\s*=.*$,\\1NVM_VERSION = '#{latest_release.gsub(/^v/,'')}'," #{node_js_rb_path})
+
+  logger.info "Updating #{node_js_rb_path}"
+  `#{sed_cmd}`
```
meatballhat: This munging of `./lib/travis/build/script/node_js.rb` seems out of place to me.  Could it be a separate target, maybe, so that the expectation is that it's performed manually?  I wouldn't want the source code to be changed during heroku deploy.

Rakefile
```diff
@@ -0,0 +1,114 @@
...
+task :update_static_files => [:casher, :gimme, :nvm, :sbt] do
+end
+
+desc 'add and commit updated static files'
+task :commit_static_files => [:update_static_files] do
+  logger.info "Adding #{files.join(" ")} to git staging area"
+  `git add #{files.join(" ")}`
+  logger.info "Creating a commit"
+  `git commit -m "Update static files"`
+end
```
meatballhat: Is this task needed anymore?

Rakefile
BanzaiMan: `Authorization: token` will fail the request.

Rakefile
BanzaiMan: No.

Rakefile
BanzaiMan: Agreed. I'll think of something.

Rakefile
BanzaiMan: On Heroku, it is less of a concern to have `GITHUB_OAUTH_TOKEN` defined, since the rate limit is less likely to apply.

bnferguson:
To recap a discussion we had in slack about the implications of this for Enterprise. As I understand it, this shouldn't affect Enterprise at all, but we will need make sure we add a step to the packaging process to run the rake task for this. 

Also we should maybe update the Dockerfile that @igorwwwwwwwwwwwwwwwwwwww made for this project so it compiles and sets up the project (and that I may use later for Enterprise). 

tl;dr - 👍 Should be good and super appreciate the heads up!

Rakefile
```diff
@@ -0,0 +1,96 @@
+require 'json'
+require 'faraday'
+require 'fileutils'
+require 'logger'
+require 'rubygems'
+require 'rspec/core/rake_task'
+
+RSpec::Core::RakeTask.new(:spec)
```
meatballhat: Have you tested this in a heroku deploy?  Given that the `rspec` gem is in the `:test` group, I thought that this usually resulted in `rspec/core/rake_task` being unavailable during heroku slug compilation, requiring some `rescue LoadError` and `if defined?(RSpec)`.

Rakefile
```diff
@@ -0,0 +1,96 @@
...
+end
+
+desc 'update sbt'
+task :sbt do
+  fetch_githubusercontent_file 'paulp/sbt-extras/master/sbt'
+end
+
+desc 'update static files'
+task :update_static_files => [:casher, :gimme, :nvm, :sbt] do
+end
```
meatballhat: I think the empty block can be omitted 😁

PR #966, opened 2017-02-22
Fetch sbt from the web app

BanzaiMan:

PR #965, opened 2017-02-21
Pull casher file from travis-build web app

BanzaiMan:
This PR adds a recent production version of `casher` to the local file storage, which will be served by the web app and fetched by `build.sh`. This allows the use of `casher` even when GitHub is down.

PR #972, opened 2017-03-06
Fetch gimme from the web app

BanzaiMan:
Continuing the work along the lines of #965 and #966, we fetch `gimme` from our web app when appropriate.

Assumes #971, but is mergeable without it.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,0 +1,84 @@
+require 'json'
+require 'faraday'
+require 'fileutils'
+require 'logger'
+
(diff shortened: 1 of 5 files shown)


BanzaiMan:
Tested in https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/547087#L129-L130.

BanzaiMan:
#971 has been substantially altered, so it's best to start anew.

PR #974, opened 2017-03-08
Wait for mysql to be really available

meatballhat:
up to the default travis_wait timeout, plus reworking how service overrides work
a smidge.

### DIFF
--- a/lib/travis/build/appliances/services.rb
+++ b/lib/travis/build/appliances/services.rb
@@ -15,16 +15,12 @@ module Travis
 
         def apply
           sh.fold 'services' do
-            if services.delete('mongodb')
-              sh.if "$(lsb_release -cs) != 'precise'" do
-                sh.cmd "sudo service mongod start", assert: false, echo: true, timing: true
-              end
-              sh.else do
-                sh.cmd "sudo service mongodb start", assert: false, echo: true, timing: true
-              end
-            end
-
             services.each do |name|
+              service_apply_method = "apply_#{name}"
+              if respond_to?(service_apply_method)
+                send(service_apply_method)
+                next
+              end
               sh.cmd "sudo service #{name.shellescape} start", assert: false, echo: true, timing: true
             end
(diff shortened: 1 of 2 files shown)


lib/travis/build/appliances/services.rb
```diff
@@ -35,6 +31,25 @@ def apply?
...
+        end
+
+        def apply_mysql
+          sh.raw <<~BASH
+            travis_mysql_ping() {
+              mysql &>/dev/null <<<'SHOW VARIABLES like "%version%"'
+            }
+          BASH
+          sh.cmd 'sudo service mysql start', assert: false, echo: true, timing: true
+          sh.cmd 'travis_wait travis_mysql_ping', assert: false, echo: false, timing: false
```
BanzaiMan: I am not sure if `travis_wait` is a good choice here. The smallest granularity that it gives is 1 minute, which might be too coerce. In addition, I am thinking about tweaking `travis_wait` a bit; have an idea, but have not worked on it.

A more straightforward thing to do is to check if the TCP port is open with some timeout.

```sh
until ruby -rsocket -e "TCPSocket.new('localhost',3306)" || test $count -lt 5; do sleep 1; let count=$count+1; done
```

BanzaiMan: See #975 for the `travis_wait` tweak.

meatballhat: @BanzaiMan How should I modify this PR, if at all? 😁

BanzaiMan: @meatballhat I suggest dropping the `travis_mysql_ping` function definition, and replacing this command `travis_wait travis_mysql_ping` with what I've suggested. (There should be a more succinct way to test TCP port 3306, though, than using `ruby`.)

emdantrim: any action on this? this PR is very stale :crying_cat_face:

meatballhat:
Superseded by #1121

PR #975, opened 2017-03-08
Tweak travis_wait

BanzaiMan:
Instead of redirecting I/O and sending commands passed to the background, we simply start a job which write audio bells every 60 seconds or so to STDOUT, so that the job we want to live long does not get killed.


BanzaiMan:
A simple test is found here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/547416

BanzaiMan:
Some GitHub issues that may benefit from this PR:

https://github.com/travis-ci/travis-ci/issues/5281
https://github.com/travis-ci/travis-ci/issues/5716
https://github.com/travis-ci/travis-ci/issues/6934
https://github.com/travis-ci/travis-ci/issues/7052
https://github.com/travis-ci/travis-ci/issues/7323

solarce:
What's the risk of this resulting in users hitting the 4mb log limit, if those users are counting on the existing `travis_wait` behavior of suppressing all output?

Could this behavior be made optional at first?

lib/travis/build/templates/header.sh
```diff
@@ -155,64 +155,31 @@ travis_terminate() {
...
-  while [ $count -lt $timeout ]; do
-    count=$(($count + 1))
-    echo -ne "Still running ($count of $timeout): $@\r"
-    sleep 60
-  done
+  _write_visual_bells $timeout &
 
-  echo -e "\n${ANSI_RED}Timeout (${timeout} minutes) reached. Terminating \"$@\"${ANSI_RESET}\n"
-  kill -9 $cmd_pid
+  travis_cmd "$*"
```
joepvd: Should likely be `"$@"`

BanzaiMan: `"$*"` is correct. With `"$@"`, `travis_cmd` only gets the first argument.

`travis_wait 15 sleep 30` results in just `sleep`, which is incorrect. (This was the rationale behind https://github.com/travis-ci/travis-build/pull/975/commits/3dad9c56c49412810673c2111a5a18b16ec1166d.)

meatballhat: @BanzaiMan I don't understand why `travis_cmd "$@"` results in only the first arg being passed.  I tend to agree with @joepvd, as "$*" passes arguments without (correctly) quoting.  Is this a case where there's a bug in another `travis_*` func that's assuming a single argument is the complete command line?

BanzaiMan: Because `travis_cmd` only takes one argument.

BanzaiMan: `travis_cmd` needs to take extra flags at the end, so that it can do other things echo command it is executing, set timing markers, etc.

`$*` is the right invocation here.

domenkozar:
While I haven't tested this solution, we did hit the quoting bug as well. Using `travis_wait 20 sleep 4600&` as a workaround

abitrolly:
So, what is required to merge this?

calvn:
@BanzaiMan any update as to when/if this will be merged? It'd be very useful to have output get streamed as the build is ran, rather than wait for `travis_wait` to timeout or exit on error.

BanzaiMan:
I'm going to revert this for now, in light of travis-ci/travis-ci#9663.

PR #980, opened 2017-03-09
Tweaks to the static file rake bits

meatballhat:


Rakefile
```diff
@@ -70,30 +66,42 @@ def latest_release_for(repo)
   end
 end
 
+task 'assets:precompile' => :update_static_files
+
+desc 'clean it up'
```
BanzaiMan: This needs a better description: "Clean up public/files", say.

PR #987, opened 2017-03-15
Compartmentalize job stages

BanzaiMan:
This PR wraps up each of the job stages into a bash function. The bash functions are thus defined in `~/.job_stages`, which will be `source`d during the build, and each of the appropriate stages are executed during the job.

As a side effect, this makes the debug job easier to deal with (e.g., a user can start a `tmate` session, then call `travis_run_install` to run all the install steps in one fell swoop).

### DIFF
--- a/lib/travis/build/stages.rb
+++ b/lib/travis/build/stages.rb
@@ -58,8 +58,16 @@ module Travis
       end
 
       def run
+        build_stages = []
         stages.each_slice(2) do |type, names|
-          names.each { |name| run_stage(type, name) }
+          names.each do |name|
+            run_stage(type, name)
+            build_stages << name
+          end
+        end
+
+        build_stages.each do |stg|
+          sh.raw "run_stage_#{stg}"
         end
       end
 
@@ -68,9 +76,12 @@ module Travis
       end
 
       def run_stage(type, name)
+        sh.raw "function run_stage_#{name}() {"
         type = :builtin if fallback?(type, name)
         stage = self.class.const_get(type.to_s.camelize).new(script, name)
-        stage.run
+        commands = stage.run
+        close = (commands.nil? || commands.empty?) ? ":\n}" : "}"
+        sh.raw close
       end
 
       def fallback?(type, name)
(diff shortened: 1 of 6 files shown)


lib/travis/build/stages.rb
```diff
@@ -58,24 +67,54 @@ def initialize(script, sh, config)
...
 
       def fallback?(type, name)
         type == :custom && !config[name]
       end
+
+      def wrap_in_func(stage)
+        if stage == :header
+          yield
+        else
+          sh.raw "cat <<'EOFUNC' >>$HOME/.build_stages"
```
meatballhat: I recall having issues with using a heredoc like this, as the body of the string can contain variables that resolve in the heredoc rather than when the body is evaluated.  Have you found that this isn't an issue?  I think my solution at the time was to base64 the heredoc contents and then decode just before writing to file.

BanzaiMan: With the quotation marks in the here-doc marker, no interpretation is performed. Example 19.7 in http://www.tldp.org/LDP/abs/html/here-docs.html.

joshk:
@BanzaiMan this is pretty cool, even just on a debug level.

Could we change this from `build_stages` to `job_stages` as 'build stages' will soon have a different meaning.

lib/travis/build/stages.rb
BanzaiMan: If you run specs locally, the `examples` directory will have generated bash scripts (and some snippets). Of course, these do not cover all the possible configurations, but they give some ideas about how the scripts may look.

cotsog:
This is indeed really cool!

I have a small suggestion to make about the naming of the commands to run each stage. Since we already have some functions prepended by `travis_`, could we do the same with these commands? e.g.

- `travis_run_before_install`
- `travis_run_install`
- `travis_run_before_script`
- `travis_run_script`

Let me know what you think.

Thanks!

BanzaiMan:
@cotsog Sure. Consistent names are good.

PR #999, opened 2017-03-23
Display `sudo` warning on containers only when `sudo` is invoked

BanzaiMan:
This PR replaces `/usr/bin/sudo` with a short bash script that displays a warning and quits. This reduces the noise when the user opts in with `sudo: false` **and** knows that `sudo` is unavailable.


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/555439

BanzaiMan:
It is probably desirable to exit with nonzero if `sudo` is invoked this way. The user expects `sudo` to work, but it doesn't.

PR #1002, opened 2017-03-24
R on OS X: use unsigned R image instead of signed one

lukasstadler:
this should solve https://github.com/travis-ci/travis-ci/issues/7370 by downloading the unsigned image instead of the signed one (which disappeared due to problems with the build process).

/cc @craigcitro @jimhester @s-u @rubak

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -109,9 +109,9 @@ module Travis
                 # output.
                 sh.cmd 'brew update >/dev/null', retry: true
 
-                # R-devel builds available at research.att.com
+                # (unsigned) R-devel builds available at research.att.com
                 if r_version == 'devel'
-                  r_url = "https://r.research.att.com/mavericks/R-devel/R-devel-mavericks-signed.pkg"
+                  r_url = "https://r.research.att.com/mavericks/R-devel/R-devel-mavericks.pkg"
 
                 # The latest release is the only one available in /bin/macosx
                 elsif r_version == r_latest
--- a/spec/build/script/r_spec.rb
+++ b/spec/build/script/r_spec.rb
@@ -71,7 +71,7 @@ describe Travis::Build::Script::R, :sexp do
   it 'downloads and installs R devel on OS X' do
     data[:config][:os] = 'osx'
     data[:config][:r] = 'devel'
-    should include_sexp [:cmd, %r{^curl.*r\.research\.att\.com/mavericks/R-devel/R-devel-mavericks-signed\.pkg},
+    should include_sexp [:cmd, %r{^curl.*r\.research\.att\.com/mavericks/R-devel/R-devel-mavericks\.pkg},
                          assert: true, echo: true, retry: true, timing: true]
   end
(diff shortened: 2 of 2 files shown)


jimhester:
The signed package seems to be available again, could you double check this and if so close this PR.

lukasstadler:
indeed - it is available again. thanks!

gvegayon:
I think that the problem is back again:

https://travis-ci.org/USCbiostats/aphylo/jobs/269734261#L43

```sh
The command "eval curl -fLo /tmp/R.pkg https://r.research.att.com/mavericks/R-devel/R-devel-mavericks-signed.pkg " failed. Retrying, 3 of 3.
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (60) SSL certificate problem: Invalid certificate chain
More details here: http://curl.haxx.se/docs/sslcerts.html
curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
The command "eval curl -fLo /tmp/R.pkg https://r.research.att.com/mavericks/R-devel/R-devel-mavericks-signed.pkg " failed 3 times.
The command "curl -fLo /tmp/R.pkg https://r.research.att.com/mavericks/R-devel/R-devel-mavericks-signed.pkg" failed and exited with 60 during .
```

Any ideas why this could be?

BanzaiMan:
@gvegayon Looks like an upstream issue to me. The URL gives me the same error on my machine (on AT&T, incidentally).

```sh-session
$ curl -I https://r.research.att.com/mavericks/R-devel/R-devel-mavericks-signed.pkg
curl: (60) SSL certificate problem: Invalid certificate chain
More details here: https://curl.haxx.se/docs/sslcerts.html

curl performs SSL certificate verification by default, using a "bundle"
 of Certificate Authority (CA) public keys (CA certs). If the default
 bundle file isn't adequate, you can specify an alternate file
 using the --cacert option.
If this HTTPS server uses a certificate signed by a CA represented in
 the bundle, the certificate verification probably failed due to a
 problem with the certificate (it might be expired, or the name might
 not match the domain name in the URL).
If you'd like to turn off curl's verification of the certificate, use
 the -k (or --insecure) option.
```

s-u:
That's and entirely different problem - which has nothing to do with the package itself. In this case the SSL cert for the server expired - I'm working on getting a new one. In the meantime you can safely use  http://r.research.att.com/mavericks/R-devel/R-devel-mavericks-signed.pkg because the package itself is signed.

gvegayon:
That sounds good! Thanks!

PR #1008, opened 2017-03-30
Skip customizable stage if `skip` or `ignore`

BanzaiMan:
This PR adds special tokens `skip` and `ignore`, which instructs the job to completely skip the customizable stage when they are used as the single token; e.g.,

```yaml
before_install: skip
```

Any other configuration (such as being in a sequence) is simply passed through:

```yaml
script:
  - skip # will execute the command `skip` (which may not succeed)
```

### DIFF
--- a/lib/travis/build/stages.rb
+++ b/lib/travis/build/stages.rb
@@ -74,15 +74,15 @@ module Travis
       def run
         define_header_stage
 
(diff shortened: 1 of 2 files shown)


svenfuchs:
Looks good to me 👍

PR #1015, opened 2017-04-06
Relax branch names when computing cache names

BanzaiMan:
Previously, we dropped everything that is not in the character class `[\w\.\_\-]`.
This meant that, if the branch name consists of characters outside of this
character class, we end up with an empty branch name, and passing invalid value to
casher, culminating in the mysterious exception in the build log:

```
attempting to download cache archive
/home/travis/.casher/bin/casher:63:in `block in fetch': undefined method `[]' for nil:NilClass (NoMethodError)
⋮
```

This PR changes this behavior.

By allowing a wider set of characters in the branch name to pass
to casher, we allow emoji-only branch names such as "🐡".

Technically speaking, it is not necessary to drop `/`, since it is
a legal character in git branch names.
However, retaining this common character in the cache name will
invalidate existing caches.


BanzaiMan:
Note that a white space was dropped from the branch names in certain specs.
The reasoning here is that the white space is not a valid character in git branch
names, and since the branch name got here, the value is legal and does not
contain white spaces.

BanzaiMan:
Tested in staging: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/730009#L438

The current result from production: https://travis-ci.org/BanzaiMan/travis_production_test/builds/445780850#L432

PR #1018, opened 2017-04-12
Don't fail build if curling of casher fails.

bnferguson:
Some config issues in Enterprise showed the issue of if we failed to pull down casher it'd fail the build. We'd prefer to failover to githubusercontent. Or just move on, simply not using caching. 

/cc @BanzaiMan - do I understand this issue correct and this fix?


bnferguson:
Testing this right now in an Enterprise build.

bnferguson:
Looks good. Doesn't failover to githubusercontent but does continue the build at least. Still issues when curling a self signed cert setup, but that's to be expected.  At least it doesn't fail now!

BanzaiMan:
@bnferguson

> Doesn't failover to githubusercontent

I wonder why that is. It may not be critical, but it would be great to figure out why.

PR #1023, opened 2017-04-20
Display error message when build script is not generated

BanzaiMan:
When users' `.travis.yml` has problems with the data structure (rather than invalid YAML), `travis-build` just generally throws an exception, and we display a generic error message:

    An error occurred while generating the build script.

(e.g., https://travis-ci.org/gligoran/cordova-set-version/jobs/223634347#L1)

This is not very helpful. Here, we `rescue` the Exception, and give some advice if we can.

### DIFF
--- a/lib/travis/build/script.rb
+++ b/lib/travis/build/script.rb
@@ -252,6 +252,7 @@ module Travis
           doc_path = exception.is_a?(Travis::Build::CompilationError) ? exception.doc_path : ''
 
           [
+            "",
             "There was an error in the .travis.yml file from which we could not recover.\n",
             msg,
             "",
--- a/lib/travis/build/script.rb
+++ b/lib/travis/build/script.rb
@@ -6,6 +6,7 @@ require 'rbconfig'
 
 require 'travis/build/addons'
 require 'travis/build/appliances'
+require 'travis/build/errors'
 require 'travis/build/git'
 require 'travis/build/helpers'
 require 'travis/build/stages'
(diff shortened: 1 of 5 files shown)


BanzaiMan:
Now has backtrace in a fold. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/574118#L11

lib/travis/build/script.rb
```diff
@@ -240,6 +246,37 @@ def debug_enabled?
...
+          if exception.is_a? Travis::Build::CompilationError
+            msg = [
+              exception.message
+            ]
+            doc_path = exception.doc_path
+          else
+            msg = [
+              "Unfortunately, we do not know much about this error.",
+              "",
+              "#{exception.class}: #{exception.message}"
```
meatballhat: Is there any risk of secret leakage in `exception.message`?

BanzaiMan:
Examples:

* [a generic error](https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/577974)
* [`addons.apt` definition error](https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/577709)
* [`deploy` config error](https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/574118)

lib/travis/build/script.rb
BanzaiMan: Here, `exception` is a builtin Ruby exception. I do not believe it will contain user input here.

BanzaiMan: Well, that's not necessarily true. In one example, it actually comes from RubyGems' argument error.

Without the error message, the argument class is useless. :-(

BanzaiMan:
We will display the Sentry error ID instead. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/578842#L11

PR #1024, opened 2017-04-20
Fail if dartfmt fails for some reason

kevmoo:
Closes https://github.com/travis-ci/travis-ci/issues/7632

### DIFF
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -236,6 +236,10 @@ MESSAGE
             end
 
             sh.cmd 'unformatted=`dartfmt -n .`'
+            # If `dartfmt` fails for some reason
+            sh.if '$? -ne 0' do
+              sh.failure ""
+            end
             sh.if '! -z "$unformatted"' do
               sh.echo "Files are unformatted:", ansi: :red
               sh.echo "$unformatted"


kevmoo:
CC @a14n @nex3 @BanzaiMan

lib/travis/build/script/dart.rb
```diff
@@ -236,6 +236,10 @@ def dartfmt
             end
 
             sh.cmd 'unformatted=`dartfmt -n .`'
+            # If `dartfmt` fails for some reason
+            sh.if '$? -ne 0' do
+              sh.failure ""
+            end
             sh.if '! -z "$unformatted"' do
               sh.echo "Files are unformatted:", ansi: :red
               sh.echo "$unformatted"
```
kevmoo: I'd love to trim the content of `unformatted` too. Any ideas?

lib/travis/build/script/dart.rb
```diff
@@ -236,6 +236,10 @@ def dartfmt
             end
 
             sh.cmd 'unformatted=`dartfmt -n .`'
```
BanzaiMan: `$unformatted` is set at run time by this assignment.

If you want to trim it, you'd have to come up with a `bash` way to trim it (with `tr`, `grep`, `head`, `tail`, `sed`, whatever).

kevmoo: Let's not worry about that for now. I just want to land this so folks stop getting a false positive from `dartfmt`

BanzaiMan:
OK. We'll deploy this Tuesday next week.

PR #1029, opened 2017-04-26
Do not use binary builds on precise

jimhester:
Fixes travis-ci/travis-ci#7654

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -324,7 +324,7 @@ module Travis
           return if packages.empty?
           packages = Array(packages)
           if config[:os] == 'linux'
-            unless config[:sudo]
+            unless config[:sudo] and config[:dist] != 'precise'
               sh.echo "R binary packages not supported with 'sudo: false', "\
                 ' falling back to source install'
               return r_install packages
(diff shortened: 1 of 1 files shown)


jimhester:
@BanzaiMan I am not sure if the condition as written is valid, in particular what is the default value of `config[:dist]` if it is not explicitly set in the `.travis.yml`?

This fix is fairly high priority to deploy, as all R builds on precise with `sudo: true` will fail when they try to retrieve the binary packages with the current code.

jimhester:
I don't believe the failing integration tests are related to my changes, but let me know if this is not the case.

lib/travis/build/script/r.rb
```diff
@@ -324,7 +324,7 @@ def r_binary_install(packages)
           return if packages.empty?
           packages = Array(packages)
           if config[:os] == 'linux'
-            unless config[:sudo]
+            unless config[:sudo] and config[:dist] != 'precise'
```
BanzaiMan: Eschew compound conditions with `unless`. It is easier to understand:

```ruby
if !config[:sudo] or config[:dist] == 'precise'
```
than
```ruby
unless config[:sudo] and config[:dist] != 'precise'
```

Also, this new criterion should be mentioned in the message below.

jimhester: Thanks for the review, just pushed with the requested changes!

PR #1035, opened 2017-04-27
Add precise_apt appliance

BanzaiMan:
To update apt sources for Precise at its EOL (2017-04-28)


meatballhat:
I'm assuming this is no longer relevant or necessary given we've been allowing precise VMs to continue booting for the past year-ish :sparkling_heart:

PR #1046, opened 2017-05-09
Add a Stage::Skip, call travis_result on script

svenfuchs:
The current strategy of skipping a stage does not play well with the `TRAVIS_TEST_RESULT`, which needs to be set in order for deployments to proceed.

### DIFF
--- a/lib/travis/build/stages/skip.rb
+++ b/lib/travis/build/stages/skip.rb
@@ -1,0 +1,13 @@
+require 'travis/build/stages/base'
+
+module Travis
+  module Build
+    class Stages
+      class Skip < Base
+        def run
+          sh.raw 'travis_result 0' if script?
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


svenfuchs:
Looks like this is working: https://staging.travis-ci.org/svenfuchs/test-2/builds/581288

The "Deploy to staging" stage has passed, and deployed.

lib/travis/build/stages/skip.rb
```diff
@@ -0,0 +1,13 @@
+require 'travis/build/stages/base'
+
+module Travis
+  module Build
+    class Stages
+      class Skip < Base
+        def run
```
BanzaiMan: We should add a friendly message that we are skipping this stage because of the special configuration; e.g.,

```ruby
sh.echo "Skipping the #{name} stage due to the configuration.", ansi: :green
```

PR #1052, opened 2017-05-12
Fix dynamic GHC installation

BanzaiMan:
Set up `$PATH` correctly, and terminate the build if `ghc` or `cabal` versions do not match the value defined in `.travis.yml`.


BanzaiMan:
Resolves https://github.com/travis-ci/travis-ci/issues/7752

PR #1054, opened 2017-05-16
Retry package installs on failure.

coderanger:
Because transient network failures are just not worth the headache. Not super common, but would be nice to drop it a few orders of magnitude :-)


BanzaiMan:
We tried this in #605. Didn't quite work. We are happy to take a look again, though.

meatballhat:
The current code performs retries of the `apt-get update` but _not_ the `apt-get install`, as the latter doesn't play nicely as @BanzaiMan references is #605, so I'm closing this out :sparkling_heart:   Please holler if you think it should be reopened!

PR #1055, opened 2017-05-16
Redirect stdin/stderr instead of setting up a new PTY to re-run the build script

svenfuchs:
This adds another strategy for hooking up the logs filtering logic which, i think, is a little more "unix'ish". Instead of setting up a new PTY in the filter, and the rerunning the build script inside of that PTY the new strategy just redirects stdin and stderr to the filter. The build script continues to run in the same process, but all output gets redirected to the filter.

Issues with the current PTY strategy:

* setting up a new PTY there are some inconsistencies across infrastructures with regards to, e.g., the terminal size
* grandchild processes can get detached which causes the build script to hang, and time out
* we add extra `\r` carriage returns to the log output

This also adds the ability to roll this out incrementally using travis-rollout. Since travis-build does not use a Redis instance we'll want to control this via env vars:

```
ROLLOUT=redirect_io
ROLLOUT_REDIRECT_IO_OWNERS=svenfuchs
ROLLOUT_REDIRECT_IO_PERCENT=25
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ gem 'rack-ssl', '~> 1.4'
 gem 'sentry-raven'
 gem 'sinatra', '~> 1.4'
 gem 'travis-config'
+gem 'travis-rollout', git: 'https://github.com/travis-ci/travis-rollout', ref: 'sf-refactor'
 gem 'travis-support', git: 'https://github.com/travis-ci/travis-support'
 
 gem 'faraday'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -12,6 +12,13 @@ GIT
     jemalloc (1.4.5)
 
 GIT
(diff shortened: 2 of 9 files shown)


bnferguson:
This PR introduced a bug into Enterprise installations causing builds to fail. It's very non-obvious as this only happens when a worker can't download the `filter.rb` from the local build server. When that happens it fails over to travis-ci.com and grabs that version. This is non-ideal as this file is not versioned for enterprise and can move like we see here. Unfortunately the way Enterprise installations use Self Signed Certificates makes remedying this tricky (either curl code in an insecure way or grab it from a secure, known, location). 

I think for the short term the easiest fix for our customers would be to keep the old `filter.rb` intact and use a new name for the improved one. This would maintain the API that older clients would have while still letting us roll out new functionality.

PR #1060, opened 2017-05-18
Migrate xbuild to msbuild

TOTBWF:
With the release of Mono 5, xbuild has been deprecated in favor of msbuild (Further details [here](http://www.mono-project.com/docs/about-mono/releases/5.0.0/#msbuild)). Luckily, the syntax is exactly the same, so updating just a matter of changing every reference of xbuild to msbuild.

### DIFF
--- a/lib/travis/build/script/csharp.rb
+++ b/lib/travis/build/script/csharp.rb
@@ -138,7 +138,7 @@ View valid versions of \"dotnet\" at https://docs.travis-ci.com/user/languages/c
           super
 
           sh.cmd 'mono --version', timing: true if is_mono_enabled
-          sh.cmd 'msbuild /version', timing: true if is_mono_enabled
+          sh.cmd "#{mono_build_cmd} /version", timing: true if is_mono_enabled
           sh.echo ''
 
           sh.cmd 'dotnet --info', timing: true if is_dotnet_enabled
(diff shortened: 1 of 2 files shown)


BanzaiMan:
@joshua-anderson @akoeplinger @nterry Opinions, please?

akoeplinger:
We can't do it this way since it'd break on older Mono versions where msbuild isn't available, we need to make it conditional on Mono >= 5.0.

You can look at `is_mono_after_4_4` for an example of such a check

TOTBWF:
Added checks to see if the mono version is greater than 5, and added tests to ensure that everything works

akoeplinger:
@BanzaiMan I have a few other changes that I'd like to get in either today or tomorrow, I guess this doesn't need to be deployed right away and can be tested together :)

TOTBWF:
Awesome, thanks 😄

TOTBWF:
Just checking in to see the progress on this, all of my builds have been failing for a while because of this 🙁

TOTBWF:
bump

akoeplinger:
@TOTBWF sorry for the delay, I forgot about this :( Rebased it in #1261.

btw. you shouldn't have been blocked by this since it only applies when the `solution` key in .travis.yml is specified which is just a shortcut for

```
script:
  - msbuild /p:Configuration=Release your_solution.sln
```

PR #1061, opened 2017-05-19
Move browserstack install/start logic to methods

BanzaiMan:
The use of `return` in this context is used as "skip the rest of
the logic for browserstack fold", but by using Ruby's `return`,
it actually breaks the resulting Bash script (with a syntax error).

By moving this portion into a method call, the use of "return" is
correctly reflected in the resulting valid Bash script.


BanzaiMan:
When browserstack access key decryption fails, the resulting bash script is invalid. This results in https://github.com/travis-ci/travis-ci/issues/7648.

Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/584657

PR #1063, opened 2017-05-24
Exit early on all HHVM+Precise combination

BanzaiMan:
Not just nightlies.

Resolves https://github.com/travis-ci/travis-ci/issues/7712 by documenting that it doesn't work.


BanzaiMan:
Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/587165

BanzaiMan:
The container image still has HHVM pre-installed, so it is usable there.

PR #1064, opened 2017-05-26
Upgrade to Composer 1.0.0 only by detecting version 1.0-dev

BanzaiMan:
`[^ ]*` is necessary because Composer 1.0-dev inserts
ANSI escape sequence for yellow.

Resolves https://github.com/travis-ci/travis-ci/issues/782.

### DIFF
--- a/lib/travis/build/script/php.rb
+++ b/lib/travis/build/script/php.rb
@@ -174,8 +174,12 @@ hhvm.libxml.ext_entity_whitelist=file,http,https
         end
 
         def composer_self_update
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Tested: 
* https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/587966 (using PHP 5.6, which is not affected)
* https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/587971 (with PHP 5.3; fails for Trusty, since we do not support it)

PR #1065, opened 2017-05-27
Integrate redactor in place of builtin filter script

meatballhat:


meatballhat:
@joecorcoran Egad sorry for the long delay.  Deploying this to staging should be enough 👍 ❤️

lib/travis/build/appliances/setup_filter.rb
```diff
@@ -33,19 +38,19 @@ def host
             app_host
           end
 
-          def download_url
-            "https://#{host}/filter.rb".untaint
+          def download_url(os_name)
+            {
+              'osx' => "https://#{host}/redactor_darwin_amd64".untaint,
```
BanzaiMan: Looks like we are missing `files` dir before `redactor_*` here and below.

BanzaiMan:
The conflicts should also be resolved. @joecorcoran and I decided not to try resolving them because it was a little more work than we wanted to take on.

With the suggested change in download URL, the build proceeds, and it was tested in https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/595184.

joecorcoran:
Confirm test works here too https://staging.travis-ci.org/joecorcoran/test/builds/595202

Would love to test this against a more complex build that gave the Ruby filter.rb issues. Can anyone point me at one?

joecorcoran:
Has the same issue as the early Ruby versions here, with a hanging process that times out: https://staging.travis-ci.org/joecorcoran/test/builds/595204

BanzaiMan:
I don't think this is no longer relevant. Is that true?

meatballhat:
Given the multiple conflicts and staleness and lack of interest and possible irrelevance, I'm sending this to the scrap heap :sparkling_heart:

PR #1072, opened 2017-06-04
Do not fail the build if `tlmgr update` returns non 0 error code

jimhester:
This should hopefully work around these errors, caused by the release of TeXLive 2017 yesterday. Complicating the matter the CTAN mirrors do not all seem to be updated yet, so I cannot simply update the texlive bundle to 2017 to fix this. However I think this will work in the short term, and prevent this breakage next year as well.

Fixes https://github.com/craigcitro/r-travis/issues/174, https://github.com/travis-ci/travis-ci/issues/7852


EllaKaye:
Thanks! That seems to be working fine again now.

PR #1078, opened 2017-06-09
Auto-accept EULA for ttf-mscorefonts-installer

joepvd:
https://staging.travis-ci.org/travis-ci/travis_staging_test/builds/593858


BanzaiMan:
I think treating `ttf-mscorefonts-installer` as a special case like this is not a good idea. Instead, we are looking into providing a more general solution for interacting with the `debconf-set-selections` command.

kevmoo:
I'd vote for a special-case solution that solves an immediate problem that can be removed when the general solution becomes available.

PR #1083, opened 2017-06-21
Allow docker job to fail

BanzaiMan:
This job requires secrets, and will fail on external PR builds.

Ideally, we should not run this job at all, but until we have such capability, we just allow it to fail.


BanzaiMan:
The feature I mention above is: https://github.com/travis-ci/travis-ci/issues/7181

PR #1090, opened 2017-06-28
remove neo4j apt list

rkh:


rkh:
Confirmed working on staging and org.production.

PR #1092, opened 2017-06-29
Return "default" if none of pre-installed rubies match regexp

BanzaiMan:
When looking for rubies that is desirable for `travis_internal_ruby`,
we only consider rubies that match the env `TRAVIS_BUILD_INTERNAL_RUBY_REGEX`.
As of 2017-06-29, this is one of 1.9.3, 2.0.x, 2.1.x, or 2.2.x.

If none of pre-installed rubies satisfies this assumption, the
local array rubies_array_len is empty, which leads to $i being -1.
This causes a confusing error message:

    /home/travis/.travis/job_stages: line 890: rubies_array_sorted: bad array subscript

This error message is benign, as the function will go on to echo
"default" in this case, but it is best not to show this.


BanzaiMan:
This _could_ potentially cause a problem, however, due to https://github.com/travis-ci/dpl/issues/513, when deployment is triggered with a provider that can't run on 2.3.x or 2.4.x.

PR #1094, opened 2017-07-03
Automate GHC and Go version aliases

meatballhat:
so that we can limit our involvement to bumping the `default` alias.

+ some general changes in Rakefile to support arbitrary modes without more positional args.


Rakefile
```diff
@@ -174,12 +189,102 @@ desc 'update tmate'
...
 desc 'update rustup'
 file 'public/files/rustup-init.sh' => 'public/files' do
-  fetch_githubusercontent_file "", "sh.rustup.rs", "rustup-init.sh"
+  fetch_githubusercontent_file '', host: 'sh.rustup.rs', to: 'rustup-init.sh'
+end
+
+desc 'update raw gimme versions'
+file 'public/files/gimme-versions-binary-linux' => 'tmp' do
+  fetch_githubusercontent_file(
+    'travis-ci/gimme/master/.testdata/sample-binary-linux',
```
tianon: Ahhh, this is the line I couldn't find on my previous pass through this file -- just to confirm, this essentially makes https://github.com/travis-ci/gimme/blob/31c7d80fddc8058c490d36ff4da1d70dd5a93aa8/.testdata/sample-binary-linux the source-of-truth for which version of Go is the "latest" of each release track?

Rakefile
```diff
@@ -174,12 +189,102 @@ desc 'update tmate'
...
+
+desc 'update gimme versions'
+file 'public/files/gimme-versions-binary-linux.json' => 'public/files/gimme-versions-binary-linux' do
+  raw = File.read('public/files/gimme-versions-binary-linux').split(/\n/).reject do |line|
+    line.strip.empty? || line.strip.start_with?('#')
+  end
+
+  raw.sort! do |a, b|
+    a.split('.').map(&:to_i) <=> b.split('.').map(&:to_i)
+  end
```
tianon: Clever way to do a version sort! :heart_eyes:

Rakefile
```diff
@@ -174,12 +189,102 @@ desc 'update tmate'
...
+    major = fullparts.first
+    out.merge!(
+      full_version => full_version,
+      "#{major}" => full_version,
+      "#{major}.x" => full_version,
+      "#{major}.x.x" => full_version,
+      "#{fullparts[0]}.#{fullparts[1]}" => full_version,
+      "#{fullparts[0]}.#{fullparts[1]}.x" => full_version
+    )
+  end
```
tianon: Isn't this going to end up at the very end with `1.x` pointing at `1.2.2` rather than `1.8.3`?  What do we do about things like `1.9beta1`?  Just ignore them?

meatballhat: It _would_ if it were reversed after sort.  Having some tests would be cool, though.

The `make` target over in `gimme` that's responsible for generating this list explicitly ignores versions matching `(rc|beta)`.

tianon: Oh duh, I forgot how sorts work this morning. :innocent: :+1: :heart:

Rakefile
```diff
@@ -174,12 +189,102 @@ desc 'update tmate'
...
+end
+
+desc 'update gimme versions'
+file 'public/files/gimme-versions-binary-linux.json' => 'public/files/gimme-versions-binary-linux' do
+  raw = File.read('public/files/gimme-versions-binary-linux').split(/\n/).reject do |line|
+    line.strip.empty? || line.strip.start_with?('#')
+  end
+
+  raw.sort! do |a, b|
+    a.split('.').map(&:to_i) <=> b.split('.').map(&:to_i)
```
BanzaiMan: Perhaps use `#version_for` and `Gem::Version.new` for consistency?

Rakefile
```diff
@@ -174,12 +189,102 @@ desc 'update tmate'
...
+      "#{major}.x" => full_version,
+      "#{major}.x.x" => full_version,
+      "#{fullparts[0]}.#{fullparts[1]}" => full_version,
+      "#{fullparts[0]}.#{fullparts[1]}.x" => full_version
+    )
+  end
+
+  raise StandardError, 'no go versions parsed' if out.empty?
+
+  out.merge!(
```
BanzaiMan: Why is this `#merge!` necessary?

Rakefile
```diff
@@ -174,12 +189,102 @@ desc 'update tmate'
...
+    to: 'ghc-versions.html',
+    mode: 0o644
+  )
+end
+
+desc 'update ghc versions'
+file 'public/files/ghc-versions.json' => 'public/files/ghc-versions.html' do
+  out = {}
+  File.read('public/files/ghc-versions.html')
+      .scan(%r[<a href="[^"]+">(?<version>\d[^<>]+)/</a>]i)
```
BanzaiMan: I am not a big fan of scanning HTML files with Regular Expressions. Surely, there is a better source of authority than this.

Rakefile
meatballhat: Because go versions don't follow consistent rules 😁

Rakefile
meatballhat: If you know of one, I'll gladly switch to it.  The raw HTML in this case is a directory index from ... maybe nginx?... so it's fairly slim HTML and this regex feels pretty safe to me.

lib/travis/build/rake_tasks.rb
```diff
@@ -0,0 +1,395 @@
...
+
+      def file_update_rustup
+        fetch_githubusercontent_file(
+          '', host: 'sh.rustup.rs', to: 'rustup-init.sh'
+        )
+      end
+
+      def file_update_raw_go_versions
+        fetch_githubusercontent_file(
+          'travis-ci/gimme/master/.testdata/sample-binary-linux',
```
BanzaiMan: How is this file getting generated?

meatballhat: This comes directly from the Google Storage API for the bucket with Go releases: https://github.com/travis-ci/gimme/blob/31c7d80fddc8058c490d36ff4da1d70dd5a93aa8/Makefile#L64-L82

tianon: FYI, that's going to need to handle API pagination once 1.9 actually releases (see https://github.com/travis-ci/gimme/issues/93#issuecomment-308575631).

PR #1096, opened 2017-07-05
Run `npm ls` for npm 5+

BanzaiMan:
Starting with version 5, npm no longer displays dependency tree
when you run `npm install`.

As there is no single command that can produce the dependency tree for both
npm <= 4 and 5+, we check for the npm version, and run `npm ls` if
necessary.

This resolves https://github.com/travis-ci/travis-ci/issues/7991.

### DIFF
--- a/lib/travis/build/script/node_js.rb
+++ b/lib/travis/build/script/node_js.rb
@@ -6,6 +6,8 @@ module Travis
 
         YARN_REQUIRED_NODE_VERSION = '4'
 
+        NPM_QUIET_TREE_VERSION = '5'
+
         def export
           super
           if node_js_given_in_config?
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/598323#L432

sonicdoe:
Just FYI, this change just broke our builds on Travis CI. The issue is that `npm ls` exits with a status code of `1` if it finds unmet peer dependencies.

Granted, we should take care of the error itself, too, however, there might also be cases where it would be fine to have unmet peer dependencies.

StoneCypher:
ya this broke my build too

https://travis-ci.org/StoneCypher/jssm/jobs/250411327

BanzaiMan:
@StoneCypher Are you reporting an outstanding breakage we should look at, or a historic one that's fixed already by #1098?

StoneCypher:
#1098 probably fixes my problem, but as it hasn't yet been deployed, I'm uncertain

BanzaiMan:
@StoneCypher Thanks for the clarification. Let us know if you continue to see problems.

PR #1097, opened 2017-07-05
Run `npm disable prefix` before `npm install`

BanzaiMan:
This can prevent nvm from installing desired Node.js runtime


BanzaiMan:
This was meant to fix https://github.com/travis-ci/travis-ci/issues/8008, but it does not. Closing.

PR #1099, opened 2017-07-05
Update firefox dev edition URL

jrbenny35:
This just landed so this PR updates the ```product``` variable based on the new URL.

https://bugzilla.mozilla.org/show_bug.cgi?id=1357378


BanzaiMan:
Thanks for the PR. It appears to me, then, that the comment above no longer applies. If you agree, could you also delete it as well?

jrbenny35:
Was there something blocking this @BanzaiMan ?

BanzaiMan:
Sorry about the delay.

jrbenny35:
No Problem, thanks!

PR #1100, opened 2017-07-05
Run `rvm use` in configure

BanzaiMan:
RVM resets certain environment variables to ensure that it can
set up Ruby environment correctly.
One such environment variable is $BUNDLE_PATH. If `.travis.yml`
attempts to set $BUNDLE_PATH in env, this is no longer effective
in the subsequent portions of the build. This is confusing.

We resolve this by moving `rvm …` commands to configure, so that
`export` is effective.

### DIFF
--- a/lib/travis/build/script/shared/rvm.rb
+++ b/lib/travis/build/script/shared/rvm.rb
@@ -24,7 +24,7 @@ module Travis
           sh.export 'TRAVIS_RUBY_VERSION', config[:rvm], echo: false if rvm?
         end
 
-        def setup
+        def configure
           super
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/shared/rvm.rb
```diff
@@ -24,7 +24,7 @@ def export
           sh.export 'TRAVIS_RUBY_VERSION', config[:rvm], echo: false if rvm?
         end
 
-        def setup
+        def configure
```
cotsog: Do we need to change occurences where `setup` was called before?

PR #1102, opened 2017-07-07
Map lang 'minimal' to the Generic script

edmorley:
Previously it returned the fallback of Ruby.
See travis-ci/docs-travis-ci-com#1267.

Also adds specs for Travis::Build.by_lang(), since the previous specs were removed in #320 (presumably unintentionally).


edmorley:
Many thanks 👍

PR #1104, opened 2017-07-10
Fix nix.rb to use wget

matthewbauer:
The current install script still uses curl and that was causing a problem for me see:
https://travis-ci.org/matthewbauer/nixpkgs-config/jobs/252089435
and https://github.com/travis-ci/travis-ci/issues/7636.

/cc @domenkozar

### DIFF
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -40,8 +40,19 @@ module Travis
           super
 
           sh.fold 'nix.install' do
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Just curious… how is `wget` an improvement over `curl`?

domenkozar:
@matthewbauer the problem lies in old openssl version (possibly patched). I recommend using xcode 8.3 that fixes the problem instead of rewriting the install script

matthewbauer:
ok i can give that a try for now. regardless it does seem like we should support "retry" in Nix tarball downloads for when our connection breaks, but that should probably be changed there.

PR #1106, opened 2017-07-11
Use Dart SDK's formatter by default, not custom.

matanlurey:
As discussed on https://github.com/dart-lang/dart_style/issues/643.

The previous behavior meant if you depended on the dart formatter package (`dart_style`) you automatically ran formatter checks using the version you pinned to. This is not the behavior any user expects.

To get this behavior, you can add the following to `.travis.yml`:

```yaml
- dartfmt: custom
```

### DIFF
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -230,11 +230,10 @@ MESSAGE
           def dartfmt
             args = task[:dartfmt]
 
-            if args.is_a?(String)
-              sh.echo "dartfmt arguments aren't supported.", ansi: :red
-            end
-
-            sh.if package_direct_dependency?('dart_style'), raw: true do
+            # If specified `-dartfmt: custom` and there is a dependency on `dart_style` we
+            # will use the custom (pinned) version of dart_style to run formatting checks
+            # instead of the SDK.
+            sh.if args == 'custom' and package_direct_dependency?('dart_style'), raw: true do
               sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
             end
 
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -234,7 +234,7 @@ MESSAGE
             # will use the SDK version of dart_style to run formatting checks instead of
             # the custom pinned version.
             if args != 'sdk'
-              if args.is_a?(String)
+              if !args.nil?
                 sh.echo "dartfmt only supports 'sdk' as an optional argument value.", ansi: :red
               end
               sh.if package_direct_dependency?('dart_style'), raw: true do
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -233,8 +233,15 @@ MESSAGE
             # If specified `-dartfmt: custom` and there is a dependency on `dart_style` we
             # will use the custom (pinned) version of dart_style to run formatting checks
             # instead of the SDK.
-            sh.if args == 'custom' and package_direct_dependency?('dart_style'), raw: true do
-              sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
+            if args == 'custom'
+              sh.if package_direct_dependency?('dart_style'), raw: true do
+                sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
+              end
+              sh.else do
+                sh.failure "Since 'custom' was used as the dartfmt argument, you should directly depend on dart_style in pubspec.yaml"
+              end
+            elsif args.is_a?(String)
+              sh.echo "dartfmt only supports 'custom' as an optional argument value.", ansi: :red
             end
 
             sh.cmd 'unformatted=`dartfmt -n .`'
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -230,15 +230,14 @@ MESSAGE
           def dartfmt
             args = task[:dartfmt]
 
-            # If specified `-dartfmt: custom` and there is a dependency on `dart_style` we
-            # will use the custom (pinned) version of dart_style to run formatting checks
-            # instead of the SDK.
-            if args == 'custom'
+            # If specified `-dartfmt: sdk` and there is a dependency on `dart_style` we
+            # will use the SDK version of dart_style to run formatting checks instead of
+            # the custom pinned version.
+            if args != 'sdk'
               sh.if package_direct_dependency?('dart_style'), raw: true do
+                sh.echo 'Using the provided dart_style package to run format instead of the SDK.'
+                sh.echo "You may specify '- dartfmt: sdk' in order to use the SDK version instead"
                 sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
-              end
-              sh.else do
-                sh.failure "Since 'custom' was used as the dartfmt argument, you should directly depend on dart_style in pubspec.yaml"
               end
             elsif args.is_a?(String)
               sh.echo "dartfmt only supports 'custom' as an optional argument value.", ansi: :red
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -240,7 +240,7 @@ MESSAGE
                 sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
               end
             elsif args.is_a?(String)
-              sh.echo "dartfmt only supports 'custom' as an optional argument value.", ansi: :red
+              sh.echo "dartfmt only supports 'sdk' as an optional argument value.", ansi: :red
             end
 
             sh.cmd 'unformatted=`dartfmt -n .`'
--- a/lib/travis/build/script/dart.rb
+++ b/lib/travis/build/script/dart.rb
@@ -234,13 +234,14 @@ MESSAGE
             # will use the SDK version of dart_style to run formatting checks instead of
             # the custom pinned version.
             if args != 'sdk'
+              if args.is_a?(String)
+                sh.echo "dartfmt only supports 'sdk' as an optional argument value.", ansi: :red
+              end
               sh.if package_direct_dependency?('dart_style'), raw: true do
                 sh.echo 'Using the provided dart_style package to run format instead of the SDK.'
                 sh.echo "You may specify '- dartfmt: sdk' in order to use the SDK version instead"
                 sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
               end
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/dart.rb
```diff
@@ -230,11 +230,10 @@ def dartanalyzer
...
-            if args.is_a?(String)
-              sh.echo "dartfmt arguments aren't supported.", ansi: :red
-            end
-
-            sh.if package_direct_dependency?('dart_style'), raw: true do
+            # If specified `-dartfmt: custom` and there is a dependency on `dart_style` we
+            # will use the custom (pinned) version of dart_style to run formatting checks
+            # instead of the SDK.
+            sh.if args == 'custom' and package_direct_dependency?('dart_style'), raw: true do
               sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
```
kevmoo: Uh...wouldn't `custom` be passed to the args here?

lib/travis/build/script/dart.rb
```diff
@@ -230,11 +230,10 @@ def dartanalyzer
...
 
-            if args.is_a?(String)
-              sh.echo "dartfmt arguments aren't supported.", ansi: :red
-            end
-
-            sh.if package_direct_dependency?('dart_style'), raw: true do
+            # If specified `-dartfmt: custom` and there is a dependency on `dart_style` we
+            # will use the custom (pinned) version of dart_style to run formatting checks
+            # instead of the SDK.
+            sh.if args == 'custom' and package_direct_dependency?('dart_style'), raw: true do
```
a14n: This doesn't work. `sh.if` takes a String as parameter and generates a shell if. Here `args == 'custom'` should rather be used alone with a ruby `if`.

Something like that should work:

```ruby
if args == 'custom'
  sh.if package_direct_dependency?('dart_style'), raw: true do
    sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
  end
  sh.else do
    sh.failure "As 'custom' is used as dartfmt argument you should directly depend on dart_style in pubspec.yaml"
  end
elsif args.is_a?(String)
  sh.echo "dartfmt only support 'custom' as optional argument.", ansi: :red
end
```

nex3:
I'm opposed to this change. Not only is it backwards-incompatible, I don't think it's the correct default. I understand that @matanlurey ran into a case where `dart_style` was being depended on as a library rather than as a tool, but that doesn't mean it's the majority case. *Any* package may want to depend on `dart_style` as a tool to get the latest and greatest formatting, whereas only a few packages are likely to use it as a library.

kevmoo:
I know of zero packages that are using `dart_style` as a dependency to get the "latest and greatest" – but at least 2 who have hit this issue.

nex3:
> I know of zero packages that are using `dart_style` as a dependency to get the "latest and greatest" – but at least 2 who have hit this issue.

Where are you looking? I just ran a quick script against recently-published Dart packages and found a bunch that have a dev dependency depend on `dart_style`, including `opentracing`, `w_transport`, `fluri`, `dart_language_server`, `w_common`, `w_module`, `w_flux`, `dart_to_js_script_rewriter`, `platform_detect`, `transformer_utils`, `built_redux`, `cbor`, `recompose_dart`, `sass4dart`, `state_machine`, `money`, `fortune_1000_companies`, `angular_ast`, `term_glyph`, `seltzer`, `scissors`, `random_users`, `superheros`, `pact_consumer_dart`, `w_service`, `docs_analyzer`, `era_geojson`, `syncbase`, `dev_compiler`, `v23discovery`, `ng2_autogrow_textarea`, and `darpule`.

kevmoo:
...now I have examples 😄

matanlurey:
I'm happy to immediately after send every author of a package with this constraint a PR updating `.travis.yml` to use `-dartfmt: custom`. I imagine 95% of them won't care (no impact) and I personally know the author and 5% will accept the PR.

nex3:
More people are depending on `dart_style` as a tool for their package than as a library. This will probably happen more over time as the formatter improves without being pushed into the stable SDK, since it's frozen during the Dart 2 push. This is compelling evidence that the current default, which assumes `dart_style` is being used as a tool, is correct.

I understand that your experience was with a package that was using `dart_style` as a library, and that that caused unpleasant behavior for you. But your personal experience isn't necessarily an indicator of the correct default for everyone—we need to decide that based on information about how people actually use the tool.

matanlurey:
That isn't compelling enough evidence for me to punish the other 99.9% of package authors if they decide to take a dependency on `dart_style`. We are writing more and more code generators both internally and externally, and they rely on `dart_style` to emit formatted code, but they do not use `dart_style`'s pinned version to run formatting checks.

Anyway this is a ridiculous argument to have on the public travis PR. I'm requesting to land this PR ASAP and we can chat after-the-fact on what the correct behavior for Dart 2.0 should be, perhaps with the Dart 2.0 team instead of just us two guessing.

nex3:
>  That isn't compelling enough evidence for me to punish the other 99.9% of package authors if they decide to take a dependency on dart_style.

Where are you getting this 99.9% statistic? Why do you think that in general, a user taking a dependency on the formatter would prefer *not* to use the version of the formatter they depended on, especially since the majority of packages that have such a dependency now seem to use it for formatting their own package?

> We are writing more and more code generators both internally and externally, and they rely on dart_style to emit formatted code, but they do not use dart_style's pinned version to run formatting checks.

This may be true, but we're writing even more packages that use `dart_style` as a tool, and we'll have motivation to do continue doing so throughout the near future.

> Anyway this is a ridiculous argument to have on the public travis PR. I'm requesting to land this PR ASAP and we can chat after-the-fact on what the correct behavior for Dart 2.0 should be, perhaps with the Dart 2.0 team instead of just us two guessing.

You're proposing a backwards-incompatible change that's opposed by the maintainer of the Dart Travis infrastructure. I think that warrants a fair amount of discussion.

If you want to land something quickly, I'd be happy to approve a CL that keeps the existing default behavior and allows users to opt in to using the SDK's formatter.

nex3:
Hey, I think we got a bit carried away talking past each other here. I definitely hear you that the current behavior doesn't work for your use-case, and your idea of adding configuration is great. If we allow users to use the SDK's formatter with `dartfmt: sdk`, you'll be able to do that for your packages and existing users won't be affected.

matanlurey:
Ok, I've landed the changes for `- dartfmt: sdk`, with the caveat that it at least now it echos to console saying that it used `dart_style` and how to change the behavior back to using the SDK.

lib/travis/build/script/dart.rb
```diff
@@ -230,12 +230,17 @@ def dartanalyzer
...
+            # will use the SDK version of dart_style to run formatting checks instead of
+            # the custom pinned version.
+            if args != 'sdk'
+              sh.if package_direct_dependency?('dart_style'), raw: true do
+                sh.echo 'Using the provided dart_style package to run format instead of the SDK.'
+                sh.echo "You may specify '- dartfmt: sdk' in order to use the SDK version instead"
+                sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
+              end
+            elsif args.is_a?(String)
+              sh.echo "dartfmt only supports 'custom' as an optional argument value.", ansi: :red
```
a14n: s/custom/sdk/

lib/travis/build/script/dart.rb
```diff
@@ -230,12 +230,17 @@ def dartanalyzer
...
+            # the custom pinned version.
+            if args != 'sdk'
+              sh.if package_direct_dependency?('dart_style'), raw: true do
+                sh.echo 'Using the provided dart_style package to run format instead of the SDK.'
+                sh.echo "You may specify '- dartfmt: sdk' in order to use the SDK version instead"
+                sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
+              end
+            elsif args.is_a?(String)
+              sh.echo "dartfmt only supports 'sdk' as an optional argument value.", ansi: :red
             end
```
a14n: I don't think the conditions are ok.
If I use `dartfmt: sdk` the first `if` will be `false` and the `elsif` true then the message _dartfmt only supports 'sdk' as an optional argument value_ will be displayed.
If I use `dartfmt: something` the first `if` will be `true`.
If I use `dartfmt` it seems to be ok.

matanlurey: Thanks, fixed. PTAL.

BanzaiMan:
This has been deployed to staging for testing.

matanlurey:
Thanks @BanzaiMan!

lib/travis/build/script/dart.rb
```diff
@@ -230,12 +230,18 @@ def dartanalyzer
...
-            end
-
-            sh.if package_direct_dependency?('dart_style'), raw: true do
-              sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
+            # If specified `-dartfmt: sdk` and there is a dependency on `dart_style` we
+            # will use the SDK version of dart_style to run formatting checks instead of
+            # the custom pinned version.
+            if args != 'sdk'
+              if args.is_a?(String)
+                sh.echo "dartfmt only supports 'sdk' as an optional argument value.", ansi: :red
```
a14n: Should it be a `sh.failure` ?

matanlurey:
@a14n It was an echo before. Probably if we are prioritizing backwards compatibility it should remain an echo.

a14n:
Using a `sh.failure` here is not backwards incompatible as the old version didn't accept a value for `dartfmt` entry.

Nevertheless it looks good to me.

matanlurey:
@BanzaiMan Let me know if there are any changes necessary to get this merged!

BanzaiMan:
I'll deploy it tomorrow.

lib/travis/build/script/dart.rb
```diff
@@ -230,12 +230,18 @@ def dartanalyzer
...
-              sh.echo "dartfmt arguments aren't supported.", ansi: :red
-            end
-
-            sh.if package_direct_dependency?('dart_style'), raw: true do
-              sh.raw 'function dartfmt() { pub run dart_style:format "$@"; }'
+            # If specified `-dartfmt: sdk` and there is a dependency on `dart_style` we
+            # will use the SDK version of dart_style to run formatting checks instead of
+            # the custom pinned version.
+            if args != 'sdk'
+              if args.is_a?(String)
```
nex3: Shouldn't this be `!args.nil?`? We don't want people passing `dartfmt: 12` or whatever.

BanzaiMan: Should this be addressed before deploying?

matanlurey:
Updated @a14n @BanzaiMan

PR #1107, opened 2017-07-11
Restore filter redirection

joepvd:


lib/travis/build/appliances/setup_filter.rb
```diff
@@ -49,11 +49,14 @@ def owner_login
             fi
           ),
           redirect_io: %(
-            exec > >(
-              %{curl}
-              %{exports}
-              ruby ~/filter.rb %{args}
-            ) 2>&1
+            if [[ -z "$TRAVIS_FILTERED" ]]; then
```
svenfuchs: I believe this env var won't be set anywhere. Do we need this condition?

lib/travis/build/templates/header.sh
```diff
@@ -166,6 +166,9 @@ travis_result() {
 
 travis_terminate() {
   set +e
+  [[ -n $TRAVIS_FILTERED && -e /dev/fd/9 ]] \
```
svenfuchs: Should we add a comment here that explains what this does?

lib/travis/build/appliances/setup_filter.rb
joepvd: The pty strategy sets this variable (for unrelated reasons). This gets used in the two tests in `travis_terminate` to gauge whether it should be attempted to dismantle the filter. Not sure how defensiveness and the creation of a new env var relate.

PR #1108, opened 2017-07-11
Keep `rvm current` STDOUT

BanzaiMan:
When setting up `casher`, ensure that we retain the output from `rvm
current`. If we sent it to `/dev/null`, we effectively run `rvm use`,
which will respect `.ruby-version` if it is present, and if this
version is pre-installed, that version is selected, even though we had
previously switched to a different version.

This resolves https://github.com/travis-ci/travis-ci/issues/8069

### DIFF
--- a/lib/travis/build/script/shared/directory_cache/base.rb
+++ b/lib/travis/build/script/shared/directory_cache/base.rb
@@ -102,7 +102,7 @@ module Travis
           end
 
           def run_rvm_use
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Before https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/601386

Afterhttps://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/601395 (in particular https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/601397#L472)

PR #1111, opened 2017-07-13
Introduce hide_secrets_in_debug config

BanzaiMan:
If set (in public setting), skip secure env vars when running in
debug mode, so that the log can be public.

### DIFF
--- a/lib/travis/build/appliances/env.rb
+++ b/lib/travis/build/appliances/env.rb
@@ -6,6 +6,8 @@ module Travis
       class Env < Base
         MSG = "Setting environment variables from %s"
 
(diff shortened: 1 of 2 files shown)


BanzaiMan:
Tested in staging: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/600857#L171-L172

BanzaiMan:
Needs specs.

BanzaiMan:
Notice that this would not work well when we merge .org and .com. After the merge, each repository would have to know its visibility, and we need to take that into account when we decide to keep or drop secrets.

BanzaiMan:
Closing for now.

PR #1116, opened 2017-07-14
Add AddonConfigError

BanzaiMan:
For a general catch-all error message for all addons


BanzaiMan:
See https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/603124 for an example error.

BanzaiMan:
Previously, it would have manifested this way: https://travis-ci.org/wardpeet/lighthouse/jobs/253691020

lib/travis/build/addons.rb
```diff
@@ -56,6 +56,8 @@ def addon(name, config)
           const = self.class.const_get(name.to_s.camelize)
           const.new(script, sh, data, config) if const && run_addon?(const)
         rescue NameError
+        rescue
+          raise Travis::Build::AddonConfigError.new(name)
```
meatballhat: Does this need to be a bare `rescue`?  Would `rescue StandardError => e` with `e` _at least_ being logged be acceptable?

BanzaiMan: If we do, we probably don't want to rescue the exception at all, since a new exception will hide `e`. I would also want to avoid putting a lot of logic in `Travis::Build::CompilationError`.

BanzaiMan:
Closing for now.

PR #1120, opened 2017-07-17
Allow lang archives to be served from non-S3 servers

BanzaiMan:
So that Travis Enterprise can pull archives from somewhere other
than our S3 buckets.

Notice that we have not publicized our archive URL scheme (bucket
names, prefix structure, etc.), so releasing this feature to
Enterprise requires documenting just such a scheme.

Also note that it is entirely up to the Enterprise users to provide
the language archives. They can either copy our archives, or compile
their own runtimes for use on Travis Enterprise.
(Requirements for Travis Enterprise compatibility is also currently
undocumented.)


schultyy:
😍

BanzaiMan:
We will need to pick this up again, since we want to serve archives from GCS as well.

BanzaiMan:
This was basically done by https://github.com/travis-ci/travis-build/pull/1656.

PR #1124, opened 2017-07-20
Add shebang line to /usr/bin/sudo replacement script

BanzaiMan:
So that when it is invoked directly no syntax error is shown


BanzaiMan:
See https://github.com/travis-ci/travis-ci/issues/8129 for motivation.

PR #1130, opened 2017-07-26
Use archives for specific OTP releases for recent Elixir versions

BanzaiMan:
Use Elixir archives which are built specifically for the OTP release in use. See https://github.com/hexpm/bob/blob/master/README.md#elixir-builds for more details.

While some older combinations may work, we restrict our choices
to Elixir 1.4.0+ and OTP 19.0+.

### DIFF
--- a/lib/travis/build/script/elixir.rb
+++ b/lib/travis/build/script/elixir.rb
@@ -26,11 +26,11 @@ module Travis
           super
 
           sh.fold "elixir" do
+            archive = elixir_archive_name(elixir_version, otp_release)
             sh.if "! -f #{HOME_DIR}/.kiex/elixirs/elixir-#{elixir_version}.env" do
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/611570

BanzaiMan:
https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/611579 shows that the current `v1.5-otp-20.zip` archive contains `1.5.0-rc.2`, so we should wait until this is fixed.

See https://github.com/hexpm/bob/issues/2.

BanzaiMan:
The archive issue has been fixed.

whitfin:
@BanzaiMan where is the appropriate place to add support for these archives on `osx` builds? Right now it appears that they only work on Linux builds?

PR #1131, opened 2017-07-27
Skip update_static_files Rake task during the required jobs

BanzaiMan:
This task is sensitive to GitHub API rate limit, and often fails
during the external PR builds because of it.

We move it to a separate job, which is allowed to fail. This has
an added benefit of not running the Docker build stage during the
external PR builds, because the new "test rake" stage will fail
then.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ jobs:
       dist: trusty
       env:
       script: ./script/validate-example-payloads-with-docker
-    - stage: "test rake"
+    - stage: "test update_static_files Rake task"
       script: "bundle exec rake update_static_files"
     - stage: ":ship: it to Quay.io"
       sudo: required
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -23,19 +23,24 @@ stages:
...
     - stage: ":ship: it to Quay.io"
       sudo: required
       dist: trusty
       env:
       script: ./script/docker-build-and-push
 
-script: bundle exec rake && ./script/validate-bash-syntax
+script:
+  - mkdir -p public/files && touch public/files/nvm.sh
+  - bundle exec rake lib/travis/build/addons/sauce_connect/sauce_connect.sh spec && ./script/validate-bash-syntax
```
cotsog: Why is `lib/travis/build/addons/sauce_connect/sauce_connect.sh` added here?

BanzaiMan: This Rake task generates `sauce_connect.sh`, which a spec relies on. See https://travis-ci.org/travis-ci/travis-build/jobs/258148161#L575 for a failure.

cotsog: Gotcha. Thanks for explaining.

PR #1132, opened 2017-07-28
retry fix to wget for fragile MacTEX downloads

graemeblair:
MacTeX download from CTAN breaks some types of R builds (oldrel on Mac OS X often) about a third of the time. 

Add `wget --retry-on-http-error=403 --waitretry=3 --tries=5` to the command for downloading MacTeX from CTAN to address fragile downloads. See https://github.com/travis-ci/travis-ci/issues/6702 for discussion.

Thanks to @aaronrudkin for the specific code and @BanzaiMan for the place to drop it in.

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -450,8 +450,8 @@ module Travis
             mactex = 'BasicTeX.pkg'
             # TODO(craigcitro): Confirm that this will route us to the
             # nearest mirror.
-            sh.cmd 'curl -fLo http://mirror.ctan.org/systems/mac/mactex/'\
-                   "#{mactex} -O \"/tmp/#{mactex}\"", retry: true
+            sh.cmd 'curl -fLo \"/tmp/#{mactex}\" --retry 3 http://mirror.ctan.org/systems/mac/mactex/'\
+                   '#{mactex}'
 
             sh.echo 'Installing OS X binary package for MacTeX'
             sh.cmd "sudo installer -pkg \"/tmp/#{mactex}\" -target /"
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -450,7 +450,7 @@ module Travis
             mactex = 'BasicTeX.pkg'
             # TODO(craigcitro): Confirm that this will route us to the
             # nearest mirror.
-            sh.cmd 'wget http://mirror.ctan.org/systems/mac/mactex/'\
+            sh.cmd 'wget --retry-on-http-error=403 --waitretry=3 --tries=5 http://mirror.ctan.org/systems/mac/mactex/'\
                    "#{mactex} -O \"/tmp/#{mactex}\""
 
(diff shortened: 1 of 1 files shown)


graemeblair:
@craigcitro @jimhester.

michaellevy:
This fix would be much appreciated for us. We're getting these random build fails pretty frequently.

craigcitro:
This looks reasonable to me, but @jimhester should weigh in as the decisive vote.

craigcitro:
(In particular, should we just do this for most/all the tex-related downloads?)

jimhester:
This looks fine, however a few questions I have in general

1. Should we switch this to use `curl` like the other requests? This is the only place we are using `wget`.
2.  Should the retries be done in the call as done here or with a `:retry true` parameter to `sh.cmd` like elsewhere in the code?

craigcitro:
Yeah, I think switching to `curl` and doing retries at the travis layer both make sense.

graemeblair:
Thanks @craigcitro @jimhester. I've updated the PR to use curl and to use `, retry: true` with `sh.cmd` as you suggest.

BanzaiMan:
Instead of `retry: true`, I should mention that `curl` has the `--retry` flag (e.g., `--retry 3`), which might be a better option.

graemeblair:
Thanks @BanzaiMan. @jimhester, you had suggested using `retry: true`; what do you think of switching to `curl --retry`?

BanzaiMan:
`retry: true` would retry immediately after a failure, while `curl --retry …` would wait progressively longer. The former only looks at the command's exit status, while the latter is a bit more intelligent about HTTP status code.

jimhester:
If @BanzaiMan thinks `curl --retry` is what we should use then 👍  from me!

graemeblair:
Thanks, @jimhester and @BanzaiMan. I've switched it to using `curl --retry 3`. This should be good to go.

graemeblair:
Thanks @BanzaiMan and all!

lib/travis/build/script/r.rb
```diff
@@ -450,8 +450,8 @@ def setup_latex
             mactex = 'BasicTeX.pkg'
             # TODO(craigcitro): Confirm that this will route us to the
             # nearest mirror.
-            sh.cmd 'wget http://mirror.ctan.org/systems/mac/mactex/'\
-                   "#{mactex} -O \"/tmp/#{mactex}\""
+            sh.cmd 'curl -fLo \"/tmp/#{mactex}\" --retry 3 http://mirror.ctan.org/systems/mac/mactex/'\
+                   '#{mactex}'
```
jimhester: Pretty sure the single quotes here should be double quotes, single quotes are not expanded by the shell.

BanzaiMan:
Oops. Sorry for overlooking this. We'll fix it shortly.

BanzaiMan:
(Incidentally, it is Ruby that is interpolating the string here.)

PR #1135, opened 2017-08-04
Add JAVA_HOME to the PATH

kdaniels:
...to temporarily fix travis-ci/travis-cookbooks#875

### DIFF
--- a/lib/travis/build/script/shared/jvm.rb
+++ b/lib/travis/build/script/shared/jvm.rb
@@ -18,6 +18,7 @@ module Travis
           CLEANUPS.each do |find_arg|
             sh.raw "find #{find_arg[:directory]} -name #{find_arg[:glob]} -delete 2>/dev/null"
           end
+          sh.export 'PATH', '$JAVA_HOME:$PATH'
         end
 
         def install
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/shared/jvm.rb
```diff
@@ -18,6 +18,7 @@ def setup
           CLEANUPS.each do |find_arg|
             sh.raw "find #{find_arg[:directory]} -name #{find_arg[:glob]} -delete 2>/dev/null"
           end
+          sh.export 'PATH', '$JAVA_HOME:$PATH'
```
BanzaiMan: It is `$JAVA_HOME/bin`, not `$JAVA_HOME`, which should be added.

kdaniels: Ahhh, good catch! Will fix, thanks!

emdantrim:
We were trying to test these changes in staging but ran into some issues getting a test job to actually run in staging. We'll come back to this next week.

kdaniels:
Tested and seems to work on staging: https://staging.travis-ci.org/kdaniels/obstinate-otter/jobs/615393#L429

PR #1136, opened 2017-08-08
update to latest nightly mac build on att research

maxheld83:
cc @craigcitro @jimhester

It appears that `https://r.research.att.com/el-capitan/R-devel/R-devel-el-capitan.pkg` is now the latest nightly build on att research.

The old `mavericks/` builds seem to stop around 2017-03-23.

In line with this, the latest build from `https://r.research.att.com/mavericks/R-devel/R-devel-mavericks-signed.pkg` now (2017-08-08) now reports to be `R 3.5.0 Under development (unstable) (2017-03-23 r72389)`, which is somewhat behind the curve.

So I'm guessing the latest nightlies are now only build for mavericks.

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -111,7 +111,7 @@ module Travis
 
                 # R-devel builds available at research.att.com
                 if r_version == 'devel'
-                  r_url = "https://r.research.att.com/mavericks/R-devel/R-devel-mavericks-signed.pkg"
+                  r_url = "https://r.research.att.com/el-capitan/R-devel/R-devel-el-capitan.pkg"
 
                 # The latest release is the only one available in /bin/macosx
                 elsif r_version == r_latest


maxheld83:
yikes, sorry about the failed CI checks; I have no idea what those mean and how/why my one-line diff might have caused them.

BanzaiMan:
@maxheld83 There is a spec to check that the generated S expression contains the data. https://github.com/maxheld83/travis-build/blob/0a366ecdf15a3985c9aa741ed53869f0ebca2971/spec/build/script/r_spec.rb#L74

maxheld83:
@BanzaiMan sorry, not sure I understand what that means.
I now *also* updated that spec to point to the el cap build in https://github.com/travis-ci/travis-build/pull/1136/commits/49b44aae484e72cbcae70d3e3959b1127a67b6c5.

Perhaps this passes now.

PR #1138, opened 2017-08-15
Add runtime kernel version to system info fold

kdaniels:
As discussed, adding separate runtime kernel version, separate from build kernel version

### DIFF
--- a/lib/travis/build/appliances/show_system_info.rb
+++ b/lib/travis/build/appliances/show_system_info.rb
@@ -24,7 +24,7 @@ module Travis
             end
             sh.echo "Build id: #{Shellwords.escape(data.build[:id])}"
             sh.echo "Job id: #{Shellwords.escape(data.job[:id])}"
-            sh.echo "Runtime kernel version: #{`uname -r`.strip.untaint}"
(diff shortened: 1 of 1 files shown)


lib/travis/build/appliances/show_system_info.rb
```diff
@@ -24,6 +24,7 @@ def header
             end
             sh.echo "Build id: #{Shellwords.escape(data.build[:id])}"
             sh.echo "Job id: #{Shellwords.escape(data.job[:id])}"
+            sh.echo "Runtime kernel version: #{`uname -r`.strip}"
```
BanzaiMan: I think you need `.untaint` here.

lib/travis/build/appliances/show_system_info.rb
```diff
@@ -24,6 +24,7 @@ def header
             end
             sh.echo "Build id: #{Shellwords.escape(data.build[:id])}"
             sh.echo "Job id: #{Shellwords.escape(data.job[:id])}"
+            sh.echo "Runtime kernel version: #{`uname -r`.strip.untaint}"
```
BanzaiMan: Whee! Good catch!

PR #1144, opened 2017-08-18
Trap ERR to help with `set -e`

joepvd:


edmorley:
Is this redundant now #1162 has merged?

PR #1147, opened 2017-08-23
Run `rebar3` if it is found

BanzaiMan:
Otherwise, run `rebar`.

This effectively resolves https://github.com/travis-ci/travis-ci/issues/6506.

On Trusty images, `rebar3` is pre-installed, and it should be preferred to `rebar`.

### DIFF
--- a/lib/travis/build/script/erlang.rb
+++ b/lib/travis/build/script/erlang.rb
@@ -35,7 +35,7 @@ module Travis
             sh.cmd './rebar compile && ./rebar skip_deps=true eunit'
           end
           sh.elif rebar_configured do
-            sh.if "command -v rebar3" do
+            sh.if "-n $(command -v rebar3)" do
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/erlang.rb
```diff
@@ -24,7 +24,12 @@ def install
             sh.cmd './rebar get-deps', fold: 'install', retry: true
           end
           sh.elif rebar_configured do
-            sh.cmd 'rebar get-deps', fold: 'install', retry: true
+            sh.if "command -v rebar3" do
+              sh.cmd 'rebar3 get-deps', fold: 'install', retry: true
```
ddeboer: Actually, in rebar3 this command has become obsolete. You can [just call `rebar3 compile`](https://www.rebar3.org/v3/docs/basic-usage#section-building) (which is already done below):

PR #882, opened 2016-11-10
Support git sparseCheckout, close travis-ci/travis-ci#5286

PeterDaveHello:
This will add support for git sparseCheckout!

By default, git will checkout all the files from the repository to the filesystem, by sparseCheckout, we can decide to checkout partial of the files, so this will help decrease the number of files git will checkout, which will totally save both disk sapce, cpu time, build time!

There are only few different steps between git clone and git, but would help a lot for a huge repo :smile:

The config will have a `sparseCheckout` under `git` section to point to the **file** contains sparseCheckout config like this:

```yaml
git:
  depth: 10
  sparseCheckout: .travis.sparseCheckoutList
```

Please fell free to test and feedback!

ref:
 - https://git-scm.com/docs/git-read-tree
 - https://github.com/cdnjs/cdnjs/blob/master/documents/sparseCheckout.md

PR #1149, opened 2017-08-25
Support sparseCheckout

BanzaiMan:
This PR resolves merge conflicts in https://github.com/travis-ci/travis-build/pull/882, and addresses the naming concern raised therein.


lib/travis/build/git/clone.rb
```diff
@@ -22,12 +22,22 @@ def apply
...
-                  sh.echo "Failed to clone from GitHub.", ansi: :red
-                  sh.echo "Checking GitHub status (https://status.github.com/api/last-message.json):"
-                  sh.raw "curl -sL https://status.github.com/api/last-message.json | jq -r .[]"
+              if sparse_checkout
+                sh.cmd "git init #{dir}", assert: true, retry: true
+                sh.cmd "git -C #{dir} config core.sparseCheckout true", assert: true, retry: true
+                sh.cmd "echo #{sparse_checkout} >> #{dir}/.git/info/sparseCheckout", assert: true, retry: true
+                sh.cmd "git -C #{dir} remote add origin #{data.source_url}", assert: true, retry: true
+                sh.cmd "git -C #{dir} pull origin #{branch} #{pull_args}", assert: true, retry: true
+                sh.cmd "cat #{sparse_checkout} >> #{dir}/.git/info/sparseCheckout", assert: true, retry: true
```
BanzaiMan: @PeterDaveHello Is this `cat` correct? I don't think there is a file with this name exists at this point in execution.

BanzaiMan:
@PeterDaveHello I should note here that cdnjs is probably not a very good repository to test this feature. I went through the steps on my machine (assuming that `cat` I mentioned in the comment earlier is really `echo`), and I ended up with a directory with 100 GB of stuff in it. We do not offer VM large enough to hold that much data.

If you would like to move forward, please find a different, smaller repository to test it.

BanzaiMan:
For posterity, here's the build that runs out of disk space on GCE (which has the largest disk capacity we offer): https://staging.travis-ci.org/BanzaiMan/cdnjs/builds/617985

lib/travis/build/git/clone.rb
BanzaiMan: Indeed, it fails when the execution actually gets to this spot: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/617995

BanzaiMan:
Further test: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/617997

PeterDaveHello:
@BanzaiMan if the feature works, it won't use that large space :)

BanzaiMan:
@PeterDaveHello Well, that may be true, but I'd much prefer to have a complete end-to-end test done. My test is rather limited and not very indicative of its usefulness.

PeterDaveHello:
Thanks @BanzaiMan !

lib/travis/build/git/clone.rb
```diff
@@ -22,12 +22,20 @@ def apply
...
           def clone_or_fetch
             sh.if "! -d #{dir}/.git" do
-              sh.cmd "git clone #{clone_args} #{data.source_url} #{dir}", assert: true, retry: true
-              if github?
-                sh.if "$? -ne 0" do
-                  sh.echo "Failed to clone from GitHub.", ansi: :red
-                  sh.echo "Checking GitHub status (https://status.github.com/api/last-message.json):"
-                  sh.raw "curl -sL https://status.github.com/api/last-message.json | jq -r .[]"
+              if sparse_checkout
+                sh.echo "Cloning with sparse checkout specificd with #{sparse_checkout}", ansi: :yellow
```
cotsog: Nit: typo `specificd` => `specified`

lib/travis/build/git/clone.rb
```diff
@@ -121,6 +139,15 @@ def source_host_name
 
             host
           end
+
+          def warn_github_status
```
cotsog: Nice!

PR #1152, opened 2017-08-28
Ensure nvm files are executable

BanzaiMan:
This resolves https://github.com/travis-ci/travis-ci/issues/8159

`nvm.sh` exists on the build image with proper permissions, and it already has the execute bit set.


BanzaiMan:
Test: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/611422#L437

The command itself is failing because it is bogus, but you can see that the execute bit is set and respected. (Compare with the `permission denied` error message which was reported.)

PR #1156, opened 2017-08-29
Simple fix for consistency with r cmd build

muschellij2:
R language pull request: 
@craigcitro and @jimhester 

Problem: In https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/r.rb#L220, it printed that `R CMD build` was being run with `R_BUILD_ARGS`, which could be set globally.  

This global argument wasn't used in https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/r.rb#L221, but rather `#{config[:r_build_args]}`, which I changed.  

Solution: Changed https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/r.rb#L220 to use `#{config[:r_build_args]}`.

Example: https://travis-ci.org/muschellij2/baker/jobs/269764874, and minimal travis yaml:
```
language: r
warnings_are_errors: true
sudo: required

r_build_args: ""
env:
 global:
   - R_BUILD_ARGS="--no-manual --no-build-vignettes"
```


jimhester:
LGTM

PR #1159, opened 2017-08-31
[WIP] DO NOT MERGE Run addons conditionally

BanzaiMan:
Much like the deployment is triggered based on conditions, we add the conditional logic to other addons.

This PR aims to resolve https://github.com/travis-ci/travis-ci/issues/1788.

The shapes of addon configurations may vary wildly, and presenting a coherent `if`/`on` value might be a challenge. We will note such difficulties as this PR is being developed. In all cases, though, the case is taken to not break the existing build configurations.

Initially, we have MariaDB:

Before:
```yaml
mariadb: '10.2'
```

After:
```yaml
mariadb:
  version: '10.2'
  if:
    branch: '10.2'
```


BanzaiMan:
Hmm. Testing on environment variables are problematic, as they are defined after some addons (e.g., `mariadb`) have already run.

BanzaiMan:
This should be re-done using https://github.com/travis-ci/travis-conditions.

PR #1160, opened 2017-09-05
    Change Dockerfile in order to make it generic

rpsene:
This commits adds the required packages which need to
    be installed for building travis-build. It brings flexilibily
    for other architectures, requiring only the change of the name
    of Docker image.

Signed-off-by: Rafael Peria de Sene <rpsene@br.ibm.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,25 @@
-FROM ruby:2.3.4
+FROM ubuntu:latest
+#FROM armv7/armhf-ubuntu
+#FROM ppc64le/ubuntu:latest
 
 LABEL maintainer Travis CI GmbH <support+travis-app-docker-images@travis-ci.com>
 
-# throw errors if Gemfile has been modified since Gemfile.lock
+RUN apt-get -qq update && apt-get -qq upgrade -y && apt-get -qq install apt-utils
+
+RUN apt-get -qq install -y wget git ruby ruby-dev build-essential clang libffi-dev
+
+RUN gem install ffi
+
+RUN gem install bundler
+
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Hello. Could you explain what problem this PR is trying to solve?

rpsene:
@BanzaiMan This Dockerfile is tailored for x86_64 (a.k.a Intel). When building travis-build for other architectures it fails given that the base ruby image won't be available for ppc64le or arm. This PR aims to add the capability to create the necessary ruby runtime from scratch using as base a simple Ubuntu image. 
As you can see at the top of Dockerfile, users will only change the necessary Docker image accordingly.

BanzaiMan:
I see. It would be good to have a working build https://travis-ci.org/travis-ci/travis-build/jobs/272148725#L1509, and perhaps shorten the build time. (It currently takes 90 seconds to run this step https://travis-ci.org/travis-ci/travis-build/jobs/271662353#L535; and it is taking 67% longer and failing).

edmorley:
The official ruby docker image also does a lot more than what has been added to the Dockerfile by the PR:
https://github.com/docker-library/ruby/blob/d2a85bf5a3799c0592140ff66b269c823b813998/2.3/jessie/Dockerfile

Perhaps asking them to add support for other architectures might be simpler? As far as I'm aware other official images (eg the Python one) already do this.

BanzaiMan:
Sorry for the long delay. We've made changes to Dockerfile in recent weeks. If this is still relevant, do review these changes, and open a new PR.

Again, sorry for the delay, and we thank you for your contribution and understanding.

PR #1161, opened 2017-09-05
Move Rust install step to after cache is set up

BanzaiMan:
Rust sets up its binaries in `$HOME/.cargo`, so if cache is in use,
certain binaries cached there will remain that of old installation,
which could be a surprise.

By moving the installation logic to after the cache, we ensure that
rustup-init.sh result is not overwritten by the cache.

This resolves https://github.com/travis-ci/travis-ci/issues/8124.


BanzaiMan:
There is no tasks getting executed between `setup` and `setup_cache` for Rust builds, so this juxtaposition of steps should have no impact.

PR #1162, opened 2017-09-07
Update "travis_retry" to work with "set -e"

tianon:
If we rely on `$?` to detect whether the user's command failed, then if the user enabled `set -e` before invoking `travis_retry`, it won't actually retry, and will bail immediately upon the first failure.  If we instead just use `if` or `&&` directly on the user's command, then `set -e` is appeased, and we get the retry logic we're expecting. :smile: :tada:

### DIFF
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -247,9 +247,7 @@ travis_retry() {
     [ $result -ne 0 ] && {
       echo -e "\n${ANSI_RED}The command \"$@\" failed. Retrying, $count of 3.${ANSI_RESET}\n" >&2
     }
-    "$@"
-    result=$?
-    [ $result -eq 0 ] && break
+    "$@" && { result=0 && break; } || result=$?
     count=$(($count + 1))
     sleep 1
   done


tianon:
To illustrate, I'll use the following simple script:
```bash

(
	travis_retry bash -c 'echo hi; exit 1'
)

(
	set -e
	travis_retry bash -c 'echo hi; exit 1'
)
```

Before this change:
```
hi
hi
hi

The command "bash -c echo hi; exit 1" failed 3 times.

hi
```

After this change:
```
hi
hi
hi

The command "bash -c echo hi; exit 1" failed 3 times.

hi
hi
hi

The command "bash -c echo hi; exit 1" failed 3 times.

```

tianon:
Doh, right you are -- corrected change pushed, and here's an updated test to show that it works:
```bash
(
	travis_retry bash -c 'echo hi; exit 1'
)

(
	set -e
	travis_retry bash -c 'echo hi; exit 1'
)

(
	travis_retry bash -c 'echo hi; exit 0'
)

(
	set -e
	travis_retry bash -c 'echo hi; exit 0'
)
```

```console
hi

The command "bash -c echo hi; exit 1" failed. Retrying, 2 of 3.

hi

The command "bash -c echo hi; exit 1" failed. Retrying, 3 of 3.

hi

The command "bash -c echo hi; exit 1" failed 3 times.

hi

The command "bash -c echo hi; exit 1" failed. Retrying, 2 of 3.

hi

The command "bash -c echo hi; exit 1" failed. Retrying, 3 of 3.

hi

The command "bash -c echo hi; exit 1" failed 3 times.

hi
hi
```

BanzaiMan:
`result` gets either zero or one, which seems incorrect to me. If the command does exit with anything other than zero or one, we should preserve that value.

tianon:
Fair point; updated again with another approach which captures the proper exit code. :+1:

(Output of my example test is the same as in https://github.com/travis-ci/travis-build/pull/1162#issuecomment-327958953)

tianon:
Any chance I could convince someone to take another look at this and see if it's OK now? :pray: :innocent:

BanzaiMan:
We can test this in staging next week. Please ensure that you have a repo enabled on https://staging.travis-ci.org and ping me in Slack when you are ready.

tianon:
Used the same example from https://github.com/travis-ci/travis-build/pull/1162#issuecomment-327958953 (https://github.com/tianon/docker-python/commit/c8755d515e12dd4eb4344c6d45b1201e916019d7): https://staging.travis-ci.org/tianon/docker-python/builds/637395 :+1: :metal:

PR #1171, opened 2017-09-12
Revert "alias :on to :if (in Deploy::Script)"

svenfuchs:
This reverts commit c1c93918386fcc1672722aa39ba81a69519690c2.

We are going to introduce `if` using https://github.com/travis-ci/travis-conditions later once we have https://github.com/travis-ci/travis-yml in place.


BanzaiMan:
I don't think this is necessary now.

PR #1177, opened 2017-09-13
Disable _JAVA_OPTIONS and MALLOC_ARENA_MAX on sudo-enabled trusty

meatballhat:
Signed-off-by: Dan Buch <dan@travis-ci.org>


lib/travis/build/appliances/fix_sudo_enabled_trusty.rb
```diff
@@ -0,0 +1,34 @@
...
+          end
+        end
+
+        def apply?
+          !data.disable_sudo?
+        end
+
+        private
+
+        def sh_is_linux?
```
BanzaiMan: What is the purpose of `sh_` prefix?

meatballhat: Mmmmm I vaguely recall meaning to indicate that the method is generating some shell goop as opposed to answering in ruby boolean.  This bit is copy pasta from the other appliance for fixing container-based trusty.

BanzaiMan: ah, I see. I agree that it would be a good idea to standardize some function naming convention. I suppose we can start from here.

PR #1178, opened 2017-09-14
Assert all commands in the `script` step, and warn if the second (and later) command fails

BanzaiMan:
Users often find it confusing that the first failure in the `script` step does not stop the build. This PR changes this.

### DIFF
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -83,6 +83,11 @@ ${ANSI_RED}For more information, see https://docs.travis-ci.com/user/encryption-
   fi
 
   if [[ -n "$assert" ]]; then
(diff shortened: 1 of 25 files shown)


lib/travis/build/templates/header.sh
```diff
@@ -83,8 +83,14 @@ ${ANSI_RED}For more information, see https://docs.travis-ci.com/user/encryption-
   fi
 
   if [[ -n "$assert" ]]; then
+    if [[ $travis_build_step = script && $travis_func_index > 0 ]]; then
+      echo -e "${ANSI_RED} We show an error message because this is the second (or later) command in 'script' and it failed.${ANSI_CLEAR}"
```
joshk: Sorry, what is this message trying to explain to the user?

BanzaiMan:
This is still relevant, but needs an overhaul.

PR #1179, opened 2017-09-19
nix: silence and retry curl

matthewbauer:
This will setup curl to automatically silence itself and also retry when something fails. This is just a small tweak to the Nix script that tries to fix a problem we've been having lately with Nix projects.

### DIFF
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -19,6 +19,8 @@ module Travis
         def configure
           super
 
+          sh.cmd "echo '-sS --retry 3' > ~/.curlrc"
+
           # Nix needs to be able to exec on /tmp on Linux
           # This will emit an error in the container but
           # it's still needed for "trusty" Linux.


BanzaiMan:
This applies the same options to all invocations of `curl`, right? Not just those related to nix. It should be removed once the run time is configured.

matthewbauer:
> It should be removed once the run time is configured.

We can definitely do that. The problem is that curl is used by Nix pretty extensively after it's been installed by the default ```script``` command ```nix-build```. Issues like travis-ci/travis-ci#7636 sometimes happen after Nix has been installed and ```--retry 5``` seems like a pretty sensible. In addition, curl's progress bar is pretty useless in Travis logs anyway. I don't know of any way to configure Nix's usage of Curl without just configuring global curl.

BanzaiMan:
Thanks for the input. I'll leave it up to the nix community maintainers:  @domenkozar @garbas and @matthewbauer (you!).

matthewbauer:
Ok....

/cc @domenkozar @garbas

Does this sound okay to make it the default on all Travis builds?

domenkozar:
Afaik this won't solve https://github.com/travis-ci/travis-ci/issues/7636, because the retrying mechanism does not apply to the SSL layer.

I'm 0+ for merging this, it changes the expected behaviour, although for the good.

domenkozar:
Maybe we should set it to 3? Usually, that's enough.

matthewbauer:
> Maybe we should set it to 3?

Changed to that in latest commit.

PR #1180, opened 2017-09-19
Truncate commit message to 32KB

BanzaiMan:
When the commit message is too long, the assignment can result in
mysterious runtime errors.

We avoid this by truncating the message to a reasonable length.

We choose 32K because GitHub's web UI limits its display of
git commit message to something around 29K.

This resolves https://github.com/travis-ci/travis-ci/issues/8443.


BanzaiMan:
`| head -c NNNNN` works on both Linux and Mac.

PR #1186, opened 2017-09-28
Speedup containers

joepvd:


joepvd:
Closing, as the docker container should be started with additional permissions to allow remounting (`--privileged`). There would have been a 4 second speedup on a non-busy host.

PR #1190, opened 2017-10-09
Do not upgrade brew unless specifically requested

joepvd:


slodki:
This is replaced with #1467 IMHO

meatballhat:
Superseded / now-irrelevant!

PR #1192, opened 2017-10-10
Don't remove mongo32 sources anymore and refresh gpg key

bogdanap:
We've decided to remove mongo from sources.list yesterday, since their GPG key was expired.
The issue (https://jira.mongodb.org/browse/SERVER-31459) has since been fixed and we just need to refresh the key. 
 
- [x]  Tested on staging: https://staging.travis-ci.org/bogdanap/autoscale-tester/builds/631323


BanzaiMan:
Deploy at will, once confirmed in staging.

PR #1193, opened 2017-10-11
CRAN builds with gfortran 6.1 on R 3.4

jeroen:
@jimhester

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -89,11 +89,14 @@ module Travis
                 # Dependencies queried with `apt-cache depends -i r-base-dev`.
                 # qpdf and texinfo are also needed for --as-cran # checks:
                 # https://stat.ethz.ch/pipermail/r-help//2012-September/335676.html
+                optional_apt_pkgs = ""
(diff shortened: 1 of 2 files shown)


jeroen:
Actually this is wrong, the fortran file is a `dmg` not a `pkg` :(

jeroen:
Update: added `hdutil` to extract the dmg file.

lib/travis/build/script/r.rb
```diff
@@ -25,6 +25,7 @@ class R < Script
           # Heavy dependencies
           pandoc: true,
           latex: true,
+          fortran: false,
```
jimhester: If we have an option for this, it definitely has to default to `true`. It also should work for non MacOS builds as well (just excluding `gfortran` from https://github.com/travis-ci/travis-build/pull/1193/files#diff-5b7e1aa01f1c6f058df6e0b9e2486c62R93 would do it I think)

jeroen:
Thanks for the feedback. PR has been updated to default to `fortran:true` and also respect this option on Linux.

jimhester:
LGTM, @BanzaiMan can you do the official review?

PR #1194, opened 2017-10-11
Append to PATH whether or not /etc/profile.d/pyenv.sh exists

meatballhat:
Removes an error that is visible in job output as:

```
sed: can't read /etc/profile.d/pyenv.sh: No such file or directory
```


meatballhat:
:+1: this is working well in staging

PR #1203, opened 2017-10-14
haxe: use the latest stable version defined by the haxe.org website as default

andyli:
Define a new value `stable` as the default haxe version.
The actual version will be found out from https://haxe.org/website-content/downloads/versions.json.

### DIFF
--- a/lib/travis/build/script/haxe.rb
+++ b/lib/travis/build/script/haxe.rb
@@ -16,15 +16,16 @@ module Travis
   module Build
     class Script
       class Haxe < Script
+
         DEFAULTS = {
-          haxe: '3.4.3',
+          haxe: 'stable',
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/haxe.rb
```diff
@@ -124,6 +125,20 @@ def script
 
         private
 
+          def haxe_versions
+            require 'faraday'
+            JSON.parse(Faraday.get("https://haxe.org/website-content/downloads/versions.json").body.to_s)
```
BanzaiMan: Perhaps we should fall back to something reasonable if the HTTP request fails.

andyli: Yes, it makes sense. I've just added that!

PR #1205, opened 2017-10-17
Fix .NET specification given by a float

BanzaiMan:
See a manifestation of this error in https://travis-ci.org/UniversityOfNottingham/UoN.AspNetCore.VersionMiddleware/jobs/288560592

A spec that fails: https://travis-ci.org/travis-ci/travis-build/jobs/289255837


BanzaiMan:
This is wrong. C# builds expect `config[:dotnet]` to be a singular string everywhere.

PR #1233, opened 2017-10-19
Reorder appliances

BanzaiMan:
Separate the execution of `apt-get update` and remove MongoDB 3.2 key before the first invocation of `apt-get update`.


BanzaiMan:
Deploying next week.

PR #1236, opened 2017-10-25
Skip overwriting ~/.pearrc for PHP 5.3 and older

BanzaiMan:
PHP 5.3 and older have problems with the `[]` array syntax, so we
need to avoid it.

Resolves https://github.com/travis-ci/travis-ci/issues/8639.

### DIFF
--- a/lib/travis/build/script/php.rb
+++ b/lib/travis/build/script/php.rb
@@ -31,8 +31,8 @@ module Travis
             sh.if "$? -ne 0" do
               install_php_on_demand(version)
             end
-            sh.else do
-              unless php_5_3_or_older?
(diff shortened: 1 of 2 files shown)


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/637772

5.3.3 is pre-installed on the GCE Precise image, but `.pearrc` is not written.

BanzaiMan:
@bogdanap The spec is definitely intentional. It has a bad description of it. I'll update it shortly.

DiegoPino:
@BanzaiMan thanks for fixing this so quick. 👏

PR #1237, opened 2017-10-25
Check bash syntax of functions defined in ~/.travis/job_stages

BanzaiMan:


cotsog:
Have we considered using a linter like https://github.com/koalaman/shellcheck ? Could be interesting to hook it up somewhere in the `travis-build` CI.

BanzaiMan:
@costog I have not. It would be interesting to add such checks in the future, but not with this PR.

PR #1238, opened 2017-10-26
Quietly exit when `-v` is passed

BanzaiMan:
So that this can be invoked without drawing attention
on the first invocation.


BanzaiMan:
This will allow us to invoke `sudo` when it is __anticipated__ that it fails, but we do not want the warning to appear; for example, https://github.com/travis-ci/casher/pull/25.

PR #1239, opened 2017-10-26
Update naming scheme for Julia binaries

ararslan:
The naming scheme changed recently, which broke Travis builds on macOS for Julia nightly. This PR also bumps 'release' to point to 0.6 instead of 0.5.

Fixes https://github.com/JuliaLang/julia/issues/24308

cc @staticfloat

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -94,14 +94,14 @@ module Travis
               ext = 'linux-x86_64.tar.gz'
               nightlyext = 'linux64.tar.gz'
             when 'osx'
-              osarch = 'osx/x64'
-              ext = 'osx10.7+.dmg'
-              nightlyext = 'osx.dmg'
+              osarch = 'mac/x64'
+              ext = 'mac64.dmg'
+              nightlyext = ext
             end
             case config[:julia].to_s
             when 'release'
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/julia.rb
```diff
@@ -91,19 +91,21 @@ def julia_url
             case config[:os]
             when 'linux'
               osarch = 'linux/x64'
+              nightlyosarch = osarch
               ext = 'linux-x86_64.tar.gz'
               nightlyext = 'linux64.tar.gz'
             when 'osx'
               osarch = 'osx/x64'
+              nightlyosarch = 'mac/x64'
```
staticfloat: Instead of doing this, we should just copy the `osx` directory to `mac` on the release side of things and keep the `osx` folder around for a while until we get tired of it and drop the old links.  But we should strive for consistency when we can.

omus: I agree, this will make it much easier for tooling to support this change without requiring logic changes.

ararslan: Done.

BanzaiMan:
Does this work on older releases, too?

staticfloat:
Yes, we have copied the older releases over to mimic this same naming convention.

ararslan:
Thanks all!

PR #1240, opened 2017-10-27
Fix last fold in R scripts

jimhester:
If the last command is in a fold the fold will be expanded. In the
common case dumping the output logs is the last command, but we don't
want the fold expanded. So simply add a dummy command after dumping the
fold.

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -258,6 +258,10 @@ module Travis
           # information.
           dump_log("out")
 
+          # If a fold (like dump log) is the last command it is not folded, so
+          # add a dummy command
+          sh.cmd 'TRUE', echo: false
+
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/r.rb
```diff
@@ -258,6 +258,10 @@ def script
           # information.
           dump_log("out")
 
+          # If a fold (like dump log) is the last command it is not folded, so
+          # add a dummy command
+          sh.cmd 'TRUE', echo: false
```
BanzaiMan: I don't think this works. If you want a command that does nothing, you can try `true`, `;`, or `:`.

jimhester:
Actually now that I look at a few more build reports I think the fold is working as-is. The report I was looking at to prompt this PR must have had something else wrong.

PR #1241, opened 2017-10-28
Install dpl `< 1.9` if `travis_internal_ruby` returns `1.9*`

BanzaiMan:
This is for transitioning `dpl` to drop MRI 1.9 support and add 2.3 (and later) support.

We examine the output of `travis_internal_ruby`, and if it matches the glob pattern `1.9*`,
we run `gem install dpl -v '< 1.9'`.

When `dpl` 1.9.0 is released and drops MRI 1.9 support, those images with `travis_internal_ruby` *not* matching `1.9*` (Trusty, practically speaking) will be able to use `dpl` 1.9.0.


BanzaiMan:
It seems that `travis_internal_ruby` returns some variation of `2.2.x` on all images in production at the moment.

(This is not to say that this PR is meaningless.)

BanzaiMan:
A thorough test: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/639839

All of these jobs have `travis_internal_ruby` returning `2.2.x`,(e.g, https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/639848#L537) and we run `gem install dpl` (`--pre`, since it has `edge: true`).

BanzaiMan:
https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/639881 forces `1.9` to be returned for `travis_internal_ruby()`.

lib/travis/build/addons/deploy/script.rb
```diff
@@ -134,17 +134,13 @@ def run
...
 
-            def install(edge = config[:edge])
-              edge = config[:edge]
-              if edge.respond_to? :fetch
-                src = edge.fetch(:source, 'travis-ci/dpl')
-                branch = edge.fetch(:branch, 'master')
-                build_gem_locally_from(src, branch)
+            def install
+              sh.if "$(rvm use $(travis_internal_ruby) do ruby -e \"puts RUBY_VERSION\") = 1.9*" do
+                cmd(dpl_install_command(true), echo: false, assert: !allow_failure, timing: true)
```
cotsog: Nit: Could we consider not passing a bare `true` here? Either pass a variable with a more descriptive name e.g.

```
installDplPre1_9 = true
cmd(dpl_install_command(installDplPre1_9)...
```

Or pass the version to install instead e.g.

```
cmd(dpl_install_command('1.8')...
```

In this case the function could also be renamed to something like `dpl_install_version_command(version = 'latest')`

My 2¢

PR #1243, opened 2017-11-01
Update GHC and Go version aliases

meatballhat:
These files are generated via:

``` bash
bundle exec rake public/version-aliases/go.json public/version-aliases/ghc.json
```

The mappings are used via `Travis::Build.config` to allow humans to put things like this in their `.travis.yml`:

``` yaml
go:
- 1.8.x
- 1.x
```

``` yaml
ghc:
- 6.x
- 8.2.x
```


public/version-aliases/go.json
```diff
@@ -14,10 +14,10 @@
   "1.6.x": "1.6.4",
   "1.7.6": "1.7.6",
   "1.7.x": "1.7.6",
-  "1.8.3": "1.8.3",
-  "1.8.x": "1.8.3",
-  "1.9": "1.9",
```
soulshake: Is `1.9` removed intentionally?

meatballhat: I'm not sure why there was a `1.9` :arrow_right: `1.9` "alias" in there before.  If someone specifies a literal `1.9`, then no alias will match, meaning they will get `1.9`.

meatballhat:
Looking 👌 https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/builds/641205

PR #1244, opened 2017-11-02
Alias `none` to `generic`

strugee:


meatballhat:
Sorry about the long delay, @strugee! If you are still interested in seeing the `language: none` handled like this, the patch should use an explicit alias mapping instead of a symlinked file. Thank you! 💖

PR #1246, opened 2017-11-07
Update smalltalkCI invocation

fniephaus:
Hi @BanzaiMan,
Please let me know when this is in staging :)

Many thanks,
Fabio


BanzaiMan:
Thanks for your patience. This is in staging now.

fniephaus:
Thanks, @BanzaiMan. Looking good so far, so please deploy in prod.

Just wondering why there are blank lines like this:
image
[Here](https://github.com/fniephaus/travis-build/blob/ea639f8a199f82438b28dce8baf402195a3c043b/lib/travis/build/script/smalltalk.rb#L60-L65) is the corresponding code. Is it because of the `echo: false`s in between?

PR #1247, opened 2017-11-13
Add steps to include private auth steps hex.pm

BanzaiMan:
This PR adds a step to run `mix hex.organization auth acme` commands with the auth keys given in `hex_org_keys`.

`hex_org_keys` is either a scalar or an array of auth keys.


joshk:
@BanzaiMan this looks interesting, although I don't think we should add another config item to the .travis.yml file. Could we detect this from an env var instead? and we can document how to set this up in the Repository settings page?

PR #1248, opened 2017-11-16
Fix hhvm apt repo key

cotsog:
For context => https://hhvm.com/blog/2017/11/14/gpg-key-migration.html


lib/travis/build/appliances/fix_hhvm_source.rb
```diff
@@ -0,0 +1,23 @@
+
+require 'travis/build/appliances/base'
+
+module Travis
+  module Build
+    module Appliances
+      class FixHhvmSource < Base
+        def apply
+          sh.if "$(command -v lsb_release)" do
```
meatballhat: If the purpose of this guard is to ensure it only runs on systems with APT, then I think it instead be something like `$(apt-key add --version &>/dev/null)`.

cotsog: @meatballhat: I copied this code from similar appliances. This appliance could definitely be improved. Is it ok if we do this in a second step? I'd like to be able to ship this ASAP and need to run in 25 minutes. Thanks!

acnagy: looking around as well, I'm with @cotsog - agreed, we should change it, but next

PR #1250, opened 2017-11-17
apply hhvm fix not for precise

joepvd:
On Precise builds, the following error:
`Error: 'deb [ arch=amd64 ] http://dl.hhvm.com/ubuntu precise main' invalid`

As HHVM is not targeting precise anymore, the apt-repository only needs to be updated when the dist is not precise, see https://github.com/travis-ci/travis-ci/issues/7827

This works as expected:
https://staging.travis-ci.org/travis-ci/travis_staging_test/builds/656494

### DIFF
--- a/lib/travis/build/appliances/fix_hhvm_source.rb
+++ b/lib/travis/build/appliances/fix_hhvm_source.rb
@@ -6,7 +6,7 @@ module Travis
     module Appliances
       class FixHhvmSource < Base
         def apply
-          sh.if "$(command -v lsb_release)" do
+          sh.if "$(command -v lsb_release) && $(lsb_release -cs) != precise" do
             command = <<-EOF
             if [[ -d /var/lib/apt/lists && -n $(command -v apt-get) ]]; then
               grep -l -i -r hhvm /etc/apt/sources.list.d | xargs sudo rm -vf
@@ -14,7 +14,7 @@ module Travis
             EOF
             sh.cmd command, echo: false
             sh.cmd "apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0xB4112585D386EB94", echo: false, assert: false, sudo: true
-            sh.cmd 'add-apt-repository "deb [ arch=amd64 ] http://dl.hhvm.com/ubuntu $(lsb_release -sc) main"', echo: false, assert: false, sudo: true
+            sh.cmd 'add-apt-repository "deb [ arch=amd64 ] https://dl.hhvm.com/ubuntu $(lsb_release -sc) main"', echo: false, assert: false, sudo: true
           end
         end
       end


lib/travis/build/appliances/fix_hhvm_source.rb
```diff
@@ -6,18 +6,18 @@ module Build
...
+          sh.if "$(command -v lsb_release) && $(lsb_release -cs) != precise" do
             command = <<-EOF
             if [[ -d /var/lib/apt/lists && -n $(command -v apt-get) ]]; then
               grep -l -i -r hhvm /etc/apt/sources.list.d | xargs sudo rm -vf
             fi
             EOF
             sh.cmd command, echo: false
             sh.cmd "apt-key adv --recv-keys --keyserver hkp://keyserver.ubuntu.com:80 0xB4112585D386EB94", echo: false, assert: false, sudo: true
-            sh.cmd 'add-apt-repository "deb [ arch=amd64 ] http://dl.hhvm.com/ubuntu $(lsb_release -sc) main"', echo: false, assert: false, sudo: true
+            sh.cmd 'add-apt-repository "deb [ arch=amd64 ] https://dl.hhvm.com/ubuntu $(lsb_release -sc) main"', echo: false, assert: false, sudo: true
```
bnferguson: did you test on trusty as well again for this change to https?

joepvd: yeah, a trusty sudo run was included in the build referenced above

bnferguson: Dope! I just realized I could have looked at the builds. 🤦‍♂️

bnferguson:
One Enterprise customer is now seeing: 

```
$ sudo apt-get update -qq
W: GPG error: http://dl.hhvm.com precise Release: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY B4112585D386EB94
W: Failed to fetch http://dl.hhvm.com/ubuntu/dists/precise/main/binary-i386/Packages  404  Not Found [IP: 52.85.131.239 80]
E: Some index files failed to download. They have been ignored, or old ones used instead.
The command "sudo apt-get update -qq" failed and exited with 100 during .
Your build has been stopped.
```

Is this potentially a side effect of this PR?

joepvd:
hhvm dropped support for precise. See: https://github.com/travis-ci/travis-ci/issues/7827

bnferguson:
Yeah, I remembered that being the case but should an `apt-get update` cause this? I'm trying to get more clarification from the customer about if this is an HHVM build or if this is just doing an `apt-get update`. I think @acnagy has a reproduction on her server.

bnferguson:
(it may be that they're trying to do HHVM builds - as they were mentioning the work around didn't work for them.... and that makes sense since precise wouldn't be there)

acnagy:
@bnferguson I do have a reproduction on my server - it's for PHP builds. I'll slack you the build logs. Also, as another note - looks like this patch requires sudo? I'm not sure that's always an option for enterprise customers - that's blocking just a regular `apt-get update` on a non-PHP build for me

PR #1252, opened 2017-11-17
Remove fix i386 arch

joepvd:
This is redundant since https://github.com/travis-ci/packer-templates/commit/01534a3047a71aff97e894c87a5e5217a40cd1c9


BanzaiMan:
Has it been tested?

bogdanap:
I think that since we still have Trusty images in use (all the deprecated groups) that don't have i386 by default, this still needs to be in here unfortunately, unless we explicitly decide not to support them anymore.

PR #1253, opened 2017-11-20
Drop `localhost` from `::1` in /etc/hosts

BanzaiMan:


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/659015#L466

BanzaiMan:
Preferably, we should understand how we got here in the first place.

PR #1257, opened 2017-11-21
Add error message for badly-formed `deploy.on`

BanzaiMan:
See https://github.com/travis-ci/travis-ci/issues/8786.

### DIFF
--- a/lib/travis/build/errors.rb
+++ b/lib/travis/build/errors.rb
@@ -32,6 +32,16 @@ module Travis
       end
     end
 
+    class DeployConditionError < DeployConfigError
+      def initialize(msg = "The \\`deploy.on\\` should be a hash (dictionary).")
+        super
+      end
+
+      def doc_path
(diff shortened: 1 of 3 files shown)


lib/travis/build/errors.rb
```diff
@@ -32,6 +32,16 @@ def doc_path
       end
     end
 
+    class DeployConditionError < DeployConfigError
+      def initialize(msg = "The \\`deploy.on\\` should be a hash (dictionary).")
```
cotsog: Maybe add *key* i.e.

`"The \\`deploy.on\\` key should be a hash (dictionary).`

BanzaiMan: This form is consistent with other error messages that we have. Also, strictly speaking, it is the _value_ that should be a hash, not the key itself (which is `on`)

cotsog: How about removing `The` ?

Or "The `deploy.on` *value*..."

PR #1258, opened 2017-11-22
Wrap `brew` command with `do_brew()` function

BanzaiMan:
To ensure that `brew` commands are run with Ruby 2.3 or later

### DIFF
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -273,6 +273,22 @@ vers2int() {
   printf '1%03d%03d%03d%03d' $(echo "$1" | tr '.' ' ')
 }
 
+_ensure_ruby_23() {
+  if ! (rvm list | grep 2\.[3-9] 2>&1 >/dev/null); then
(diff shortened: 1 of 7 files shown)


lib/travis/build/templates/header.sh
```diff
@@ -273,6 +273,22 @@ vers2int() {
   printf '1%03d%03d%03d%03d' $(echo "$1" | tr '.' ' ')
 }
 
+_ensure_ruby_23() {
+  if ! (rvm list | grep 2\.[3-9] 2>&1 >/dev/null); then
```
mhinz: Nitpick: `! (rvm list | grep 2\.[3-9])` -> `rvm list | grep -v 2\.[3-9]`

BanzaiMan: No. Yours would return `true` if any line from `rvm list` output does not match `2\.[3-9]`.

mhinz: Oh, of course. Sorry for the noise.

PR #1262, opened 2017-11-28
Wrap `brew` so that it runs with Ruby 2.3

BanzaiMan:
This replaces #1258.

Resolves https://github.com/travis-ci/travis-ci/issues/8552.


BanzaiMan:
`brew` seems to be working now, so this is no longer necessary.

PR #1264, opened 2017-11-30
Do not remove SUID bits for chrome-sandbox

joepvd:
Assumptions have changed for `chrome-sandbox` to have suid bits. This
PR ensures those bits are not removed if set.

(edit by AJ)
Addresses https://github.com/travis-ci/travis-ci/issues/8836


BanzaiMan:
What problem was it causing, and has this PR been tested?

PR #1267, opened 2017-12-06
Delay setting user-defined env vars after built-ins

BanzaiMan:
So that the user-defined environment variables can make use of
default ones, such as `TRAVIS_RUBY_VERSION`.

Resolves https://github.com/travis-ci/travis-ci/issues/8858

### DIFF
--- a/lib/travis/build/script/c.rb
+++ b/lib/travis/build/script/c.rb
@@ -7,11 +7,11 @@ module Travis
         }
 
         def export
-          super
           sh.export 'CC', compiler
           if data.cache?(:ccache)
             sh.export 'PATH', "/usr/lib/ccache:$PATH"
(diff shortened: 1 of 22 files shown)


BanzaiMan:
https://github.com/travis-ci/travis-ci/issues/8858#issuecomment-349685368

> Hmmm. On a second thought, setting built-in env vars first will allow them to be overridden by users, and that does not seem like a good idea.

lightmare:
bash allows read-only variables
`declare -rx FOO=bar` instead of `export FOO=bar`
subsequent attempt to set/unset FOO will cause an error

joshk:
@lightmare oh, that is pretty cool.

@BanzaiMan thoughts?

PR #1268, opened 2017-12-06
Revert "Use `echo` instead of `cat`"

annulen:
This reverts commit fa7d7c8abb0b5f4a6f71b51672819512d859b1bb.

We need to use cat instead of echo here, because we need to check
out files listed in #{sparse_checkout}, not #{sparse_checkout} file
itself (which is already available).

### DIFF
--- a/lib/travis/build/git/clone.rb
+++ b/lib/travis/build/git/clone.rb
@@ -30,7 +30,7 @@ module Travis
                 sh.cmd "git -C #{dir} remote add origin #{data.source_url}", assert: true, retry: true
                 sh.cmd "git -C #{dir} pull origin #{branch} #{pull_args}", assert: false, retry: true
                 warn_github_status
-                sh.cmd "echo #{sparse_checkout} >> #{dir}/.git/info/sparseCheckout", assert: true, retry: true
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Reverting by itself is probably wrong. We need to check if the file exists, and deal with the case where it does not exist.

annulen:
If file does not exist, it's not a valid .travis.yml, is it?

BanzaiMan:
@annulen It is. YAML has no bearing on existence of another file.

annulen:
Oh, you mean that following process might not fail with such severely broken checkout, and user might think everything is fine, right?

annulen:
OK then. What is the correct way to terminate CI job inside clone_or_fetch?

annulen:
>YAML has no bearing on existence of another file

Sure, YAML may be valid, but configuration is not correct. User should either point `sparse_checkout` to existing file, or don't use this option

BanzaiMan:
Maybe it is obvious to those who use this feature that the file should exist? In that case, failing the command (`cat file`) —and just reverting— is acceptable outcome.

BanzaiMan:
This has been deployed.

annulen:
Error with missing file was actually caused by bug, fixed in #1270

PR #1269, opened 2017-12-06
Use full fingerprint when retrieving MongoDB 3.2 OpenPGP key

jwilk:
32-bit OpenPGP key IDs should not used, because [it's computationally trivial to generate a new key with chosen ID](https://evil32.com/).


BanzaiMan:
Thanks for opening the PR. I should point out that the key value you present is different from what's given in https://docs.mongodb.com/v3.2/tutorial/install-mongodb-on-ubuntu/, however.

BanzaiMan:
Superseded by #1290.

PR #1270, opened 2017-12-06
File name is .git/info/sparse-checkout, not sparseCheckout

annulen:
Using sparseCheckout leads to empty working directory when
sparse_checkout option is active.


BanzaiMan:
Is there a document describing what is wrong with the previous commands? My previous tests were obviously limited and lacking, so I would like to understand the issue better.

BanzaiMan:
(Trivially, specs need fixing.)

annulen:
Yes, https://git-scm.com/docs/git-read-tree#_sparse_checkout

annulen:
>$GIT_DIR/info/sparse-checkout is used to define the skip-worktree reference bitmap.

Not $GIT_DIR/info/sparseCheckout

PR #1272, opened 2017-12-13
Add no_world_writable_dirs appliance

BanzaiMan:
To ensure that no dir in `$PATH` is world-writable.

It turns out that the current clang 5.0.0 archive is faulty
and its `bin` is world-writable, causing
https://github.com/travis-ci/travis-ci/issues/8892.
We work around this issue in general by checking each dir in `$PATH`.

GNU `chmod` has a handy `--changes` flag that prints only the changes,
but this flag is not available for BSD `chmod` (on the Mac), so we
use `-vv` and filter the output instead.


BanzaiMan:
Tested: https://staging.travis-ci.org/cotsog/travis_ci_staging_regular/builds/675845

PR #1275, opened 2017-12-14
Filter `yarn global bin` output for `/`

BanzaiMan:
It turns out that `yarn` writes undersiable error message to
STDOUT if `node` or `nodejs` is not found on `$PATH`.
Adding this string is problematic, as it contains spaces.

Resolves https://github.com/travis-ci/travis-ci/issues/8921.

### DIFF
--- a/lib/travis/build/appliances/ensure_path_components.rb
+++ b/lib/travis/build/appliances/ensure_path_components.rb
@@ -6,14 +6,14 @@ module Travis
       class EnsurePathComponents < Base
 
         COMPONENTS = [
-          '$(yarn global bin)'
+          '$(yarn global bin | grep /)'
         ]
 
         def apply
           COMPONENTS.each do |pc|
             sh.cmd <<-EOF
 pc=#{pc}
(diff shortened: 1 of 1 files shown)


BanzaiMan:
See also: https://github.com/yarnpkg/yarn/pull/5094

lib/travis/build/appliances/ensure_path_components.rb
```diff
@@ -6,14 +6,14 @@ module Appliances
...
-          '$(yarn global bin)'
+          '$(yarn global bin | grep /)'
         ]
 
         def apply
           COMPONENTS.each do |pc|
             sh.cmd <<-EOF
 pc=#{pc}
-if [[ -z $(echo :$PATH: | grep :$pc:) ]]; then export PATH=$PATH:$pc; fi
+if [[ -n $pc && -z $(echo :$PATH: | grep :$pc:) ]]; then export PATH=$PATH:$pc; fi
```
joepvd: This grep might leak into the build log. Wouldn't `=~` work as a shell test?

BanzaiMan:
Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/676880#L609

PR #1284, opened 2018-01-09
Add `--retry 2` to `curl` invocations

BanzaiMan:
Another approach might be to write `~/.curlrc` with an appliance,
but that may feel like a bit of magic.

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -132,7 +132,7 @@ module Travis
             unless whitelisted.empty?
               sh.export 'DEBIAN_FRONTEND', 'noninteractive', echo: true
               whitelisted.each do |source|
-                sh.cmd "curl -sSL #{source['key_url'].untaint.inspect} | sudo -E apt-key add -", echo: true, assert: true, timing: true if source['key_url']
+                sh.cmd "curl --retry 2 -sSL #{source['key_url'].untaint.inspect} | sudo -E apt-key add -", echo: true, assert: true, timing: true if source['key_url']
                 if source['sourceline'].start_with? 'ppa:'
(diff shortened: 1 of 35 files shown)


BanzaiMan:
Some of the recent network issues might be mitigated by simply retrying.

ryndaniels:
@BanzaiMan would we potentially want a `--retry-max-time` in any of these, to keep the overall execution time from taking too long in the case of network issues, or is there a sensible default that would work?

cotsog:
Looks good to me.

Would it be worth it to try to centralize this change e.g. if we ever want to increase the number of retry in the future or remove it? 🤔

BanzaiMan:
@ryndaniels Hmm. `--retry-max-timeout` makes sense, but how long? I do not exactly have a good sense of this value.

Centralized management of these options makes sense, and we would be doing that with `~/.curlrc`, which, to the best of my knowledge`, is nonexistent on our images at this time.

ryndaniels:
@BanzaiMan Hmmm. Looking deeper into the docs on the options, maybe `-m` for each request's max time makes more sense. I believe this defaults to 5 minutes (defined in `lib/connect.h`). I'm not sure I have a great sense of this either, but if there is some limit for "if the total build script execution time is longer than X, it will cause problems with something else" that plus the number of curls could give us a starting place.

I'd also be fine with leaving that alone for now, especially if we centralize curl command definitions so it would be easier to change later. Is there a way to define a new curl command in the `travis-build` ruby code somewhere rather than dropping a `.curlrc` file on all the images?

BanzaiMan:
I suppose we can write a short function:

```bash
function curl() {
  /usr/bin/curl --retry 2 -s "$@" # `-s`, so that `curl BLAH | grep FOO` works without noise
}
```

maybe.

PR #1285, opened 2018-01-09
Add RedefineCurl appliance

BanzaiMan:
For a convenient place to define common `curl` flags

Supersedes #1284.


BanzaiMan:
Other conflicting options such as `-v` (against `-s`) should also win.

PR #1287, opened 2018-01-10
Add D DUB cache

wilzbach:
- `~/.dub` is used for all downloaded packages
- `~/dlang` is used by the install script for the compiler binaries

CC @MartinNowak @s-ludwig

I guess this page would need to be updated as well afterwards: https://docs.travis-ci.com/user/caching/

----------------------

Also I think it would be worthwhile to cache the `install.sh` script.
What would be the best way to do so? It by default downloads itself to
`~/dlang/install.sh`, so I guess we could just check for the existence in
`setup` or do you have a better method?


spec/build/script/d_spec.rb
```diff
@@ -57,4 +57,14 @@
                            assert: true, echo: true, timing: true]
     end
   end
+
+  context "when cache is configured" do
```
jacob-carlborg: Use single quotes.

BanzaiMan:
> Also I think it would be worthwhile to cache the `install.sh` script.

Why?

PR #1289, opened 2018-01-11
[Enterprise] update for enterprise 2.1.13

acnagy:
Updating the `enterprise-2.1` branch with the latest, especially for mongodb gpg key funtimes


bnferguson:
Ahhh noticed what the issue was. It wasn't a missing Ruby version, it was that this branch was branched off of `master`. There are a couple modifications in `enterprise-2.1` that we need (antipattern, I know). So the workflow is typically branch from `enterprise-*` branch, `merge master`, and then push up the new branch and make a PR. This way if we have to hotfix something only in one enterprise version it gets carried on. 

Typically we want to make our changes in `master` so this won't matter and we don't have to keep hotfixing future `enterprise-*` branches but in some cases it can make sense.

PR #1293, opened 2018-01-17
Add in Dep for Golang builds

DarthHater:
Hi there,

Out of some need for my own projects, I started a branch to add in dep (seems fairly consistent with godep).

I am not done with this yet, FYI, but wanted to get a PR up so I could figure out if there was anything stupid I'm forgetting. I notice you have `app_host` for getting binaries for example, I imagine that's to not get punished over hitting github a million times?

### DIFF
--- a/lib/travis/build/script/go.rb
+++ b/lib/travis/build/script/go.rb
@@ -70,7 +70,7 @@ module Travis
             sh.echo 'Using Go 1.5 Vendoring, not checking for Godeps'
           end
           sh.else do
-            sh .if '-f Godeps/Godeps.json' do
+            sh.if '-f Godeps/Godeps.json' do
               sh.export 'GOPATH', '${TRAVIS_BUILD_DIR}/Godeps/_workspace:$GOPATH', retry: false
               sh.export 'PATH', '${TRAVIS_BUILD_DIR}/Godeps/_workspace/bin:$PATH', retry: false
 
@@ -80,6 +80,10 @@ module Travis
                   sh.cmd 'godep restore', retry: true, timing: true, assert: true, echo: true
                 end
               end
+            end
+            sh.elif '-f Gopkg.toml' do 
+              fetch_dep
+              sh.cmd 'dep ensure', retry: true, timing: true, assert: true, echo: true
             end
           end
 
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/go.rb
```diff
@@ -81,6 +81,10 @@ def install
                 end
               end
             end
+            sh.elif '-f Gopkg.toml' do 
+              fetch_dep
+              sh.cmd 'dep ensure', retry: true, timing: true, assert: true, echo: true
```
BanzaiMan: Where is this `dep` command supposed to come from?

lib/travis/build/script/go.rb
```diff
@@ -196,6 +200,21 @@ def update_gimme
...
+            sh.mkdir "$HOME/gopath/bin", echo: false, recursive: true
+
+            sh.if "$TRAVIS_OS_NAME = osx" do
+              sh.cmd "curl -sL -o #{dep} https://github.com/golang/dep/releases/download/v0.3.2/dep-darwin-amd64", echo: false
+            end
+            sh.elif "$TRAVIS_OS_NAME = linux" do
+              sh.cmd "curl -sL -o #{dep} https://github.com/golang/dep/releases/download/v0.3.2/dep-linux-amd64", echo: false
+            end
+
+            sh.cmd "chmod +x #{dep}"
```
BanzaiMan: oh, this.

lib/travis/build/script/go.rb
DarthHater: `fetch_dep` I believe

lib/travis/build/script/go.rb
```diff
@@ -196,6 +200,21 @@ def update_gimme
             sh.cmd %Q'gimme #{DEFAULTS[:go]} &>/dev/null'
           end
 
+          def fetch_dep
+            dep = "$HOME/gopath/bin/dep"
+
+            sh.mkdir "$HOME/gopath/bin", echo: false, recursive: true
+
+            sh.if "$TRAVIS_OS_NAME = osx" do
+              sh.cmd "curl -sL -o #{dep} https://github.com/golang/dep/releases/download/v0.3.2/dep-darwin-amd64", echo: false
```
BanzaiMan: Hard coding the version number is not wise.

DarthHater: Agree, I was wondering about that in regards to my previous `app_host` question, it seems that the binaries are rehosted somewhere and then referred to there.

lib/travis/build/script/go.rb
BanzaiMan: Is this utility universally used? If not, it seems rather strange to let the presence of `Gopkg.toml` drive its use.

DarthHater: `dep` is likely to be merged into the go toolchain, it's a more mature stab at how to handle dependencies:

https://github.com/golang/dep/wiki/Roadmap

A lot of people are starting to use it versus godep (which is currently on Travis-CI). The `Gopkg.toml` is basically like dep's version of a `package.json` or `pom.xml` file. If it's there, it means you are using `dep`.

lib/travis/build/script/go.rb
AlekSi: There is no URL for "latest stable version". We _could_ use `master` branch, but personally, I'm against it – it will make bug reporting much harder.

DarthHater: Yeah, even the homebrew formula for `dep` is just done against the newest version:

https://github.com/Homebrew/homebrew-core/blob/master/Formula/dep.rb#L4

At any rate, curious if there is a place y'all park binaries to keep them from getting downloaded from remotes. I'd love to wrap this PR up if possible, if only because it would be immediately useful for my project, hah.

meatballhat:
@DarthHater Hello and sorry about the long delay! Do you think that this work is still relevant and up to date and should still be considered for merge?

DarthHater:
It might be for others! The go ecosystem still uses Dep quite a bit.

PR #1298, opened 2018-01-26
Fix local dpl installation when cwd is not $TRAVIS_BUILD_DIR

BanzaiMan:
When execution gets to deployment and PWD is not `$TRAVIS_BUILD_DIR`, installing `dpl` with the locally built `dpl-*.gem` file fails.

This PR resolves this issue.

Failure with cf6d29c9c: https://travis-ci.org/BanzaiMan/travis_production_test/builds/333491097#L540
Success with fc63b337: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/704197#L533

### DIFF
--- a/lib/travis/build/addons/deploy/script.rb
+++ b/lib/travis/build/addons/deploy/script.rb
@@ -141,6 +141,9 @@ module Travis
             end
 
             def install
+              if install_local?(config[:edge])
+                sh.cmd "pushd $TRAVIS_BUILD_DIR >&/dev/null", echo: false
+              end
               sh.if "$(rvm use $(travis_internal_ruby) do ruby -e \"puts RUBY_VERSION\") = 1.9*" do
                 cmd(dpl_install_command(WANT_18), echo: false, assert: !allow_failure, timing: true)
               end
@@ -148,6 +151,9 @@ module Travis
                 cmd(dpl_install_command, echo: false, assert: !allow_failure, timing: true)
               end
               sh.cmd "rm -f dpl-*.gem", echo: false, assert: false, timing: false
+              if install_local?(config[:edge])
+                sh.cmd "popd >&/dev/null", echo: false
+              end
             end
 
             def run_command(assert = !allow_failure)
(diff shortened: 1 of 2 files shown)


lib/travis/build/addons/deploy/script.rb
```diff
@@ -142,12 +142,12 @@ def run
 
             def install
               sh.if "$(rvm use $(travis_internal_ruby) do ruby -e \"puts RUBY_VERSION\") = 1.9*" do
-                cmd(dpl_install_command(WANT_18), echo: false, assert: !allow_failure, timing: true)
+                cmd(dpl_install_command(WANT_18), echo: true, assert: !allow_failure, timing: true)
```
cotsog: Is echoing enabled on purpose here or it's a remainder of your tests?

lib/travis/build/addons/deploy/script.rb
```diff
@@ -142,12 +142,12 @@ def run
...
               end
               sh.else do
-                cmd(dpl_install_command, echo: false, assert: !allow_failure, timing: true)
+                cmd(dpl_install_command, echo: true, assert: !allow_failure, timing: true)
```
cotsog: Same as above.

lib/travis/build/addons/deploy/script.rb
BanzaiMan: It's done on purpose; so that it is a little clearer what is happening.

bnferguson:
Seeing problems in Enterprise too, but don't quite look like this. More like; 

```
ERROR:  While executing gem ... (OptionParser::AmbiguousOption)

    ambiguous option: -*.gem

The command "rvm $(travis_internal_ruby) --fuzzy do ruby -S gem install dpl -v '< 1.9' -*.gem --local --pre" failed and exited with 1 during .
```

Looks like we may need to roll another release with updated `travis-build`?  Thoughts?

BanzaiMan:
@bnferguson For the record: I think that is a particular bug where `travis_internal_ruby` returns `1.9*` and `edge` is defined in the config. Our hosted services do not do this, so we never saw it in production. At any rate, the error should be fixed with this PR.

bnferguson:
Ahhh thanks for the clarification. Good to know why we didn't see it in hosted before.

PR #1299, opened 2018-01-31
Run tests with yarn if yarn.lock exists

xPaw:
Was moving to Yarn and we noticed tests are still being run from npm.


BanzaiMan:
I'm not sure if this should go in here. https://yarnpkg.com/en/docs/cli/test says that this is effective only when `script` is defined in `package.json`. While it is not impossible to detect this, it seems to me that the user should know to override `script` in such a case.

xPaw:
>  https://yarnpkg.com/en/docs/cli/test says that this is effective only when script is defined in package.json. 

That's the case for npm as well: https://docs.npmjs.com/cli/test so it doesn't change that behaviour.

I know it's possible to change `script` yourself, but I think it would be a little nicer for Travis to do it automatically.

BanzaiMan:
Oh! How right you are! In that case, this makes sense. Have you looked into the spec failures?

xPaw:
These failed due to `the server responded with status 403`, I don't think that's due to my changes?

BanzaiMan:
I see. The error comes from GitHub API rate limiting, I think. :-/

I'll retry.

astorije:
@BanzaiMan, it's a slightly different concern (though we're affected in thelounge/lounge#1987 as well) but GH issues are not enabled on this, so let me ask here: when switching to Yarn, does the npm registry deploys correctly use `yarn publish` instead of `npm publish`?
As you can see, we're using it [here](https://github.com/thelounge/lounge/blob/master/.travis.yml#L26-L37) and I prefer to make sure before we complete thelounge/lounge#1987).

BanzaiMan:
@astorije Are you asking about [deployment with `npm`](https://docs.travis-ci.com/user/deployment/npm)? If so, the answer is "no". It only uses the [`npm publish` command](https://github.com/travis-ci/dpl/blob/v1.8.47/lib/dpl/provider/npm.rb#L33-L35).

BanzaiMan:
@xPaw Please open a corresponding docs PR. Thanks!

xPaw:
@BanzaiMan Wasn't sure how to write it nicely, but here's the pr: https://github.com/travis-ci/docs-travis-ci-com/pull/1698

astorije:
> It only uses the [`npm publish` command](https://github.com/travis-ci/dpl/blob/v1.8.47/lib/dpl/provider/npm.rb#L33-L35).

@xPaw, do you think you could look into Yarnifying the `npm` deploys as well? 🙏
Also, you'll need to rebase this branch with `master`.

xPaw:
@astorije That would have to be a new deploy process, publishing with Yarn when deploy is explicitly named `npm` would be odd. I don't want to look into that.

astorije:
Nah, in my understanding of the docs and other deployments, it's named after the provider, not the tool being used. That's why for Ruby the provider is called `rubygems` and not `gem`, etc.
So it's safe(r) to do the same thing you did here at deployment time, because we both know mixing up npm/yarn commands does not always end well (and failing a deployment is the last thing we did...).

BanzaiMan:
Unless there is a good reason to use `yarn` _*instead of*_ `npm` to deploy, it seems rather unnecessary.

xPaw:
@BanzaiMan The reason would be avoiding npm CLI (as yarn is implemented separately, second opinion and all that).

astorije:
@BanzaiMan, additionally it would be using yarn instead of npm when already installing and testing with yarn. Use a single CLI tool for the entire chain.

xPaw:
Just FYI, `yarn publish` can't be used in scripts just yet because it expects interactivity. https://github.com/yarnpkg/yarn/issues/610

BanzaiMan:
Deployment is all about *deploying* the code. It does not matter what tool is used (`npm` or `yarn` or plain API calls to accomplish the task. If the existing machinery accomplishes the task, adding support for a new tool is a liability, not a feature.

So far, I have not seen anything to indicate that using `yarn` brings any benefit to our use case.

astorije:
@BanzaiMan, it would be good measure to use the same tool on the entire chain. If the entire chain minus deploying is done with one tool, and deploying (the crucial piece) is done with another one, then issues could potentially be discovered only at deploy time.
That said, as @xPaw mentioned and linked to, `yarn publish` is not totally ready for prod usage yet, so that's a moot point for now.

Going back to the current PR, any chance this could get in? @xPaw, would you mind rebasing your branch on `master`?

xPaw:
@BanzaiMan so is this PR good to go?

BanzaiMan:
Yup. Planning to merge and deploy tomorrow. Thanks!

PR #1300, opened 2018-02-01
Increase verbosity of apt-get update in build.sh

soulshake:
Addresses debugging needs in https://github.com/travis-ci/travis-ci/issues/9112 (maybe)


bogdanap:
I think this change would be useful, but am wondering if we can maybe put it under some debug flag of some sorts, so it doesn't affect all builds.
@BanzaiMan is there an easy way to do this?

soulshake:
@bogdanap I don't think this needs to be pushed to production if there's a way for me to use a specific version of travis-build on some test jobs. Is that possible?

BanzaiMan:
Looks like `apt` is a new command with Trusty (and not present in Precise), so we'll have to check that it exists before running. For this PR, though, it looks like we can just call `apt-get` uniformly.

BanzaiMan:
On Precise, the command `apt` exists, but it's the Java Annotation Processing Tool. https://docs.oracle.com/javase/7/docs/technotes/guides/apt/ (and it exists at the same location `/usr/bin/apt`). Not sure what to do with it. I feel that `apt-get` is a safer choice.

BanzaiMan:
Closing this in favor of a more general attempt of the same, #1343.

PR #1303, opened 2018-02-05
Tweak `sed` regex to match localhost line in `/etc/hosts`

BanzaiMan:
The space between `127.0.0.1` and `localhost` could be a tab,
which causes the previous regex to miss this line, and preserve
the line we do not want.

This resolves https://github.com/travis-ci/travis-ci/issues/9167.

### DIFF
--- a/lib/travis/build/appliances/put_localhost_first.rb
+++ b/lib/travis/build/appliances/put_localhost_first.rb
@@ -6,7 +6,7 @@ module Travis
       class PutLocalhostFirst < Base
         def apply
           # gather names mapped to 127.0.0.1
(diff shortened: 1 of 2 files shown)


BanzaiMan:
This resolves https://github.com/travis-ci/travis-ci/issues/9167.

BanzaiMan:
Before: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/711180#L3558
After: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/711189#L517

BanzaiMan:
Looks like we have a duplicate `localhost` on EC2 and Mac, but it *should* be functional.

PR #1307, opened 2018-02-08
Upgrade Go versions

itchyny:
Regarding the security upgrade: https://github.com/golang/go/issues/23672
Depends on https://github.com/travis-ci/gimme/pull/124


itchyny:
Now blocked by https://github.com/travis-ci/gimme/pull/125.

meatballhat:
@itchyny The rake tests should not be changed, fwiw, as the input is stubbed 👍

itchyny:
Oops, thank you. I dropped the commit and force-pushed. I don't still understand how the two repositories are related though...

PR #1310, opened 2018-02-19
Update Go versions

shibayu36:
As Go version in gimme was updated by https://github.com/travis-ci/gimme/pull/127 , I just ran `bundle exec rake public/version-aliases/go.json` .


public/version-aliases/go.json
```diff
@@ -18,6 +18,8 @@
   "1.8.x": "1.8.7",
   "1.9.4": "1.9.4",
   "1.9.x": "1.9.4",
+  "1.10": "1.10",
```
BanzaiMan: Unfortunately, this one is not going to work, because we only see `1.1` as a value passed.

public/version-aliases/go.json
```diff
@@ -18,6 +18,8 @@
...
+  "1.10.x": "1.10",
   "1.2": "1.2.2",
   "go1": "go1"
-}
+}
```
BanzaiMan: It's a tiny thing, but please restore the newline here.

public/version-aliases/go.json
shibayu36: Sorry, but I don't understand yet.  Do you mean that if we write following in .travis.yml

```
language: go
language: go
go:
  - 1.10
```

then travis-build recognize it as not version `1.10` but `1.1` ?

public/version-aliases/go.json
shibayu36: I fixed it by https://github.com/travis-ci/travis-build/pull/1310/commits/2bcf8f7fc4e785643485fa535e35367e500d1bff .

`bundle exec rake public/version-aliases/go.json` has removed the newline here.

public/version-aliases/go.json
AlekSi: Unfortunately, yes. Parser interprets it as a float number, not as a string, and 1.1, 1.10, and 1.100 are all the same numbers.
There is an ongoing work in gimme itself to recognize `.x` aliases. After it is done, and a new version of gimme is released, my understanding is that this file can be removed.
@BanzaiMan please correct me if I'm wrong.

AlekSi: Forgot to mention how to fix the problem now:
```yaml
language: go
go:
  - "1.10"
```

shibayu36: Thanks, I understood!

Is there anything I should do to solve this problem?

shibayu36: I think we should specify a character string for version, like perl https://docs.travis-ci.com/user/languages/perl/  ...

BanzaiMan: That won't work in Go's case, because `1.1` is a also valid version. In contrast, in the case of Perl (5.20) and Node.js (0.10), their numerical counterparts (Perl 5.2) and (Node.js 0.1) are invalid versions, so we can map these values to the ones that work.

BanzaiMan: Yes, the Go documentation should use strings instead of floats.

PR #1311, opened 2018-02-20
add golang 1.10.x version alias

igorwwwwwwwwwwwwwwwwwwww:
I wasn't sure if we also need the actual `1.10.0` version in there as a key, as it seems to be working fine without.


igorwwwwwwwwwwwwwwwwwwww:
Looks like this is a duplicate of #1310 -- closing!

PR #1312, opened 2018-02-22
Build and install dpl provider gems from source

BanzaiMan:
Starting with `dpl` 1.9.0, each deployment provider code will be split into separate gems, so that each provider's dependencies can be managed independently of other providers. `dpl` 1.9.0+ will detect the absence of provider code, and it will attempt to install the corresponding version of the provider gem on the fly.

When building and installing `dpl` from the source, however, we need to install the provider gem from the same source, since the correct gems are probably not available on the source.


lib/travis/build/addons/deploy/script.rb
```diff
@@ -203,7 +204,7 @@ def options
             end
 
             def warning_message(message)
-              sh.echo "Skipping a deployment with the #{config[:provider]} provider because #{message}", ansi: :yellow
+              sh.echo "Skipping a deployment with the #{provider} provider because #{message}", ansi: :yellow
```
cotsog: Ruby n00b here but should we refer to `provider` as `@provider`? 🤔

lib/travis/build/addons/deploy/script.rb
```diff
@@ -219,6 +220,9 @@ def build_gem_locally_from(source, branch)
               sh.cmd("git checkout #{branch}",                              echo: true,  assert: !allow_failure, timing: true)
               sh.cmd("git show-ref -s HEAD",                                echo: true,  assert: !allow_failure, timing: true)
               cmd("gem build dpl.gemspec",                                  echo: true,  assert: !allow_failure, timing: true)
+              sh.if("-f dpl-#{provider}.gemspec") do
```
cotsog: Same question as above here. I guess not.

lib/travis/build/addons/deploy/script.rb
BanzaiMan: Using the accessor makes it easier to change presentation later if need be, with `def provider`.

cotsog: Got it! Thanks!

PR #1317, opened 2018-02-27
Revert "Run 'apt-get update' last in list of scripts"

bogdanap:
My latest PR (https://github.com/travis-ci/travis-build/pull/1316) seems to have made the problem more widespread - I can at least reproduce it with all my test builds, which I couldn't before.

And while this is interesting for the investigation, I think it's better to revert it in order to reduce the disruption to our users.

### DIFF
--- a/lib/travis/build/script.rb
+++ b/lib/travis/build/script.rb
@@ -164,6 +164,7 @@ module Travis
           apply :update_apt_keys
           apply :fix_hhvm_source
           apply :update_mongo_arch
+          apply :apt_get_update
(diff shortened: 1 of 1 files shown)


bogdanap:
This is a bit weird - there seems to have been a wave of builds where the initial issue was reproducible, but now it's back to the previous state, where it only happens now and then.
I think I'll leave things as they are for now, unless customers actually start complaining.

bogdanap:
It probably doesn't make sense to keep this around anymore.

PR #1318, opened 2018-03-02
Add hack language support

cotsog:
Please notify @fredemmott [in this ticket](https://secure.helpscout.net/conversation/508977913/65718/) when it's released.


lib/travis/build/script/php.rb
```diff
@@ -100,7 +100,11 @@ def cache_slug
         private
 
         def version
-          config[:php].to_s
+          if config[:hhvm]
+            "hhvm-#{config[:hhvm].to_s}"
```
fredemmott: What would be the equivalent of:

```
php:
 - hhvm
 - hhvm-nightly
```

i.e. how would I specify that one of the things that I want to run is the latest version of HHVM?

cotsog: @fredemmott Thanks for chiming in!

How about if we supported `latest` in that case e.g. the equivalent of your snippet with `language: hack` would be

```yaml
hack:
  - latest
  - nightly
```

What do you think?

Note: the current state of the PR doesn't support this yet.

fredemmott:
*bump* the next release of HHVM will not support PHP.

BanzaiMan:
@fredemmott Thanks for the updates. We appreciate it. Do you know when the next HHVM is released?

fredemmott:
Due 11th of February

fredemmott:
*bump*

fredemmott:
HHVM 4.0 is released, an d can not execute Composer.

For now, we're going to have to revert to docker-based testing.

joshk:
@BanzaiMan @cotsog could we help get this merged and deployed?

Should this be separated from the PHP lang defaults? eg. its own class? `Script::Hack`

fredemmott:
> Should this be separated from the PHP lang defaults? eg. its own class? `Script::Hack`

Yep - HHVM 4 can not execute Composer, or most other PHP code. In 4.1, I'm likely disabling support for `<?php` as it basically indicates the file won't work anyway

BanzaiMan:
I'm closing this PR, because Hack looks to be different enough from PHP, and we will need a separate class to deal with the differences.

PR #1320, opened 2018-03-05
Wait for network prior to continuing execution

meatballhat:
The intent of this change is to ensure that execution waits until the network is available enough to make an HTTP request of the `travis-build` API.

From [this job](https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/builds/713898):

screenshot_2018-03-05_18-01-44

and the corresponding heroku logs from `travis-build-staging`:

```
2018-03-05T22:58:58.885064+00:00 heroku[router]: at=info method=GET path="/empty.txt?job_id=713899&repo=meatballhat/yolo-octo-adventure?count=1" host=build-staging.travis-ci.org request_id=0cf61427-783a-42bc-9891-e980c89f23d9 fwd="35.225.134.118" dyno=web.1 connect=1ms service=44ms status=200 bytes=244 protocol=https
```

### DIFF
--- a/lib/travis/build/templates/header.sh
+++ b/lib/travis/build/templates/header.sh
@@ -270,12 +270,12 @@ travis_download() {
   local dst="${2}"
 
   if curl --version &>/dev/null; then
-    curl -fsSL --retry=9 --retry-delay=1 -o "${dst}" "${src}" 2>/dev/null
+    curl -fsSL --connect-timeout=5 --retry=9 --retry-delay=1 -o "${dst}" "${src}" 2>/dev/null
     return $?
   fi
 
   if wget --version &>/dev/null; then
-    wget --tries=10 --wait=1 -q "${src}" -O "${dst}" 2>/dev/null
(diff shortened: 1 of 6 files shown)


lib/travis/build/templates/header.sh
```diff
@@ -265,6 +265,37 @@ travis_fold() {
...
+}
+
+travis_wait_for_network() {
+  local result=0
+  local count=1
+  local url="<%= app_host %>/empty.txt"
+  url="${url}?job_id=${TRAVIS_JOB_ID}"
+  url="${url}&repo=${TRAVIS_REPO_SLUG}"
+
+  while [ $count -le 9 ]; do
```
BanzaiMan: Instead of using a custom counter, maybe we can use `curl --retry=9 --retry-delay=1` and `wget --tries=10 --wait=1`.

BanzaiMan: I just read these flags from their man pages. They aren't tested.

PR #1325, opened 2018-03-07
Replace `crystal deps` with `shards install`

straight-shoota:
`crystal deps` was just a wrapper for `shards install` and will be removed from
the compiler in the next release (see crystal-lang/crystal#5544).

/cc @asterite, @jhass, @waj, and @will


BanzaiMan:
Does the new command work now, or do we have to wait?

straight-shoota:
Has been working for a long time. `crystal deps` did just call `shards install`. It was only kept for backwards compatibility but will now be removed.

PR #1326, opened 2018-03-09
When using local `dpl` for deployment, build and install provider gem before invoking `dpl`

BanzaiMan:
`dpl` 1.9.0 now installs provider gem dynamically. It does this via 

    install_cmd = "gem install dpl-#{provider_gem_name || opt} -v #{ENV['DPL_VERSION'] || DPL::VERSION}"

https://github.com/travis-ci/dpl/blob/66ea4d5c7e6fe80f76945db9ceb5e761dc6c4642/lib/dpl/provider.rb#L83

which means fetching from rubygems.org.

For local `dpl`, then, we need to build and install the provider gem beforehand in order to test the new `dpl` code.

This resolves https://github.com/travis-ci/dpl/issues/770.


BanzaiMan:
Test: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/714253#L563

The deployment uses this commit: https://github.com/travis-ci/dpl/commit/93d2eed8c8bb2090aa586fa40dee0706c124a5c5#diff-1c6ad0d4f864c8bfb7d1e62b8afef0f4R18

and it shows: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/714253#L563

PR #1340, opened 2018-03-17
Fix for more valid dotnet versions

JustArchi:
`2.1.101` is completely valid SDK version, as well as `2.1.300-preview1-008174`

cc @joshua-anderson @akoeplinger @nterry


akoeplinger:
@BanzaiMan this should be good to be merged and deployed to production

PR #1343, opened 2018-03-25
For a more resilient apt operations, add retries

BanzaiMan:
This replaces #1300, by using the configuration file rather than flags.


BanzaiMan:
We may consider randomly adding this configuration (rather than switching it on for all the jobs) to test it in production first.

BanzaiMan:
Test: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/715006#L7 (The addon works.)

The idea would be to allow interested users to opt-in in order to guarantee that we could get some feedback, while we raise the randomness threshold as we go along.

BanzaiMan:
Here's a test with `TRAVIS_BUILD_APT_RETRIES_PERCENTAGE=100` and MariaDB addon, showing that the file is written before the addon runs. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/715010#L415

lib/travis/build/addons/apt_retries.rb
```diff
@@ -0,0 +1,34 @@
...
+      class AptRetries < Base
+        SUPER_USER_SAFE = true
+
+        def before_configure?
+          config
+        end
+
+        def before_configure
+          sh.echo "Configuring default apt-get retries", ansi: :yellow
+          set_retries = <<-EOF
```
meatballhat: We can safely use `<<~HEREDOC`, iirc, so that this whole string may be safely indented.  The APT configs don't care about blankspace either, fwiw.

lib/travis/build/addons/apt_retries.rb
```diff
@@ -0,0 +1,34 @@
...
+  Retries "5";
+  https {
+    Timeout "30";
+  };
+};
+EOS
+            sudo mv 99apt /etc/apt/apt.conf.d
+          fi
+          EOF
+          sh.cmd set_retries
```
meatballhat: Why define the string separately and then pass to `sh.cmd`?

lib/travis/build/appliances/apt_get_update.rb
```diff
@@ -5,13 +5,33 @@ module Build
...
+  https {
+    Timeout "30";
+  };
+};
+EOS
+    sudo mv 99apt /etc/apt/apt.conf.d
+  fi
+fi
+            EOF
+            sh.cmd set_retries
```
meatballhat: Same questions/comments apply here.  I guess I don't understand this pattern :confused:

lib/travis/build/appliances/apt_get_update.rb
```diff
@@ -5,13 +5,33 @@ module Build
     module Appliances
       class AptGetUpdate < Base
         def apply
+          retry_under = Travis::Build.config.apt.retries.percentage.to_i || 0 # if under this, we add retry config
```
meatballhat: What is the intent with this rollout percentage?

lib/travis/build/addons/apt_retries.rb
BanzaiMan: There is no reason, really.

lib/travis/build/addons/apt_retries.rb
BanzaiMan: I was not aware of `<<~`.

lib/travis/build/appliances/apt_get_update.rb
BanzaiMan: "X % of all jobs will use retries set by the appliances"

lib/travis/build/addons/apt_retries.rb
meatballhat: One fewer reading/editing dereference SGTM :skull:

lib/travis/build/appliances/apt_get_update.rb
meatballhat: @BanzaiMan I understand the mechanics, but what is your intent regarding the speed of rollout, how we're tracking this, why not all at once, etc?

BanzaiMan: We don't have a good handle on how often the issue is manifesting. Only user feedback. We won't know how effective this is, and it seems prudent to me that we do not go to the deep end without testing it on a smaller scale.

We can drop the appliances idea, until the opt-in users confirm better result, and go all in, if the scheduling the rollout seems cumbersome.

lib/travis/build/addons/apt_retries.rb
```diff
@@ -0,0 +1,34 @@
...
+            cat <<-EOS > 99apt
+Acquire {
+  ForceIPv4 "1";
+  Retries "5";
+  https {
+    Timeout "30";
+  };
+};
+EOS
+            sudo mv 99apt /etc/apt/apt.conf.d
```
meatballhat: Can the basename be something specific to travis-build, such as `99-travis-build-retries`?

PR #1344, opened 2018-03-26
Add support for debug Firefox builds using latest-debug

davehunt:
I'm not sure how to test this, beyond running `bundle exec rake`. This should allow users to install the latest debug Firefox builds from Mozilla's CI. This will be useful for anyone reporting issues with running automation against Firefox.


BanzaiMan:
@davehunt Thanks for the PR. Is there a web page somewhere explaining the debug builds you mention?

davehunt:
I couldn't find anything concise. They're builds with the `--enable-debug` option set, which is covered here: https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/Configuring_Build_Options

PR #1345, opened 2018-03-27
Suppress rubygems post update messages

BanzaiMan:
As there are more RubyGems are released, the messages are getting
longer and longer, mostly unnecessarily.

We should suppress the output.


BanzaiMan:
Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/715042#L432

PR #1347, opened 2018-04-02
Support more recent versions of OracleJDK and OpenJDK

BanzaiMan:
Using https://github.com/sormuras/bach/blob/master/install-jdk.sh, we install and set up Java environment that `jdk_switcher` is unable to set up.


BanzaiMan:
This resolves https://github.com/travis-ci/travis-ci/issues/9368.

BanzaiMan:
Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/715793

sormuras:
How about adding two generic `Early-Access` constants that always refer to the latest JDKs available?

- `openjdk-ea` would now refer to `openjdk-11-ea+LATEST`, and 2018.10 it will switch to `openjdk-12-ea+LATEST`
- `oraclejdk-ea` same logic applies here

`install-jdk.sh` supports that be omitting the `-F` option.

BanzaiMan:
@sormuras Are you saying this mapping for the flags works?

```ruby
            args_for = {
              'oraclejdk-ea' => '-L BCL',
              'openjdk-ea'   => '-L GPL',
              'openjdk9'     => '-F 9  -L GPL',
              'oraclejdk10'  => '-F 10 -L BCL',
              'openjdk10'    => '-F 10 -L GPL',
              'oraclejdk11'  => '-F 11 -L BCL',
              'openjdk11'    => '-F 11 -L GPL'
            }
```

sormuras:
> Are you saying this mapping for the flags works?

Aye. The [current default](https://github.com/sormuras/bach/blob/master/install-jdk.sh#L40) values are:

```
VERSION=2018-04-03

JDK_FEATURE='11'
JDK_BUILD='?'
JDK_LICENSE='GPL'
```

BanzaiMan:
Great. `*-ea` has been deployed to .org staging now. Please test it.

sormuras:
Great! 👍 

I'll update those defaults when the JDK release train stops at the next stations, which are 2018.10, 2019.03, ...

At the moment, this will be done manually. Which implies an update of `install-jdk.sh` on your environment every six month. But I have an idea how to automate this as well.

BanzaiMan:
The current process is to fetch the latest `install-jdk.sh` at the application deployment time. This process is currently manual also, but should generally happen when this repository gets an update.

sormuras:
> Please test it.

https://staging.travis-ci.org/sormuras/sormuras.github.io/builds/716080

`openjdk-ea` resolved to `openjdk-11-ea+7` ✔️ 
`oraclejdk-ea` resolved to `oraclejdk-11-ea+7` ✔️

sormuras:
The env var `PATH` includes the binary directory twice: `PATH = /home/travis/jdk-11/bin:/home/travis/jdk-11/bin:...`
It does not hurt, but could be nicer.

BanzaiMan:
That's because [`install-jdk.sh`](https://github.com/sormuras/bach/blob/e33830117a8534d1826889d44cf55ba7d285328a/install-jdk.sh#L129) manipulates `$PATH` and so does `travis-build`. We can skip the latter.

sormuras:
Although the *push* build failed here, the test project reports "all green": https://staging.travis-ci.org/sormuras/sormuras.github.io/builds/716215

sormuras:
> But I have an idea how to automate this as well.

Updated `install-jdk.sh` to version: `2018-04-04`

Now, `-F ?` (or omitting it) leads to a look-up of the current _early access_ release.
And `-D` activates the _dry-run_ mode, which only gathers and prints all variable values.

Caveat: when the currently developed JDK goes from `-ea` state to _release candidate_ the scripts still needs a manual fix. Will take care of that ... soon.

sormuras:
> Will take care of that ... soon.

Done. Released `install-jdk.sh` version: `2018-04-05` along with [install-jdk-tests.sh](https://github.com/sormuras/bach/blob/master/install-jdk-tests.sh).

lib/travis/build/script/shared/jdk.rb
```diff
@@ -4,7 +4,17 @@ class Script
       module Jdk
         def configure
           super
-          sh.cmd "jdk_switcher use #{config[:jdk]}", assert: true, echo: true, timing: false if uses_jdk?
+          if uses_jdk?
+            if use_install_jdk?(config[:jdk])
+              download_install_jdk
+              sh.if "-f install-jdk.sh" do
+                sh.cmd "source ./install-jdk.sh #{install_jdk_args config[:jdk]}", echo: true, assert: true
```
meatballhat: Instead of `source`'ing, can this directly run `./install-jdk.sh {args}`?  I see that the project uses `source` in at least one place, but I'm assuming this is because something defined within is being used elsewhere, which I don't believe is the case for us.

sormuras: How are the updated environment variables, `JAVA_HOME` and `PATH` are set/modified by `install-jdk.sh`, propagated back to the calling shell then? Or am I missing something?

BanzaiMan: I think the main function of `install_jdk.sh` is to download archives, and set `JAVA_HOME`. If `install_jdk.sh` can print this value out instead (say, with a flag?), we can use it; e.g.,

    export JAVA_HOME=$(./install-jdk.sh FLAGS)

sormuras: That would require every current user of `install-jdk.sh` to change the call like you suggested and take care of setting `JAVA_HOME` and adopt `PATH` accordingly. Those users will lose the _fire and forget_ feature, like in this _direct_ usage style: https://github.com/junit-team/junit5/blob/master/.travis.yml

At the moment `install-jdk.sh` already prints some information while running. Would that interfer with _can print this value out instead_? May bash-fu is limited here. PR is welcome. 😉

sormuras: See current print out here: https://travis-ci.org/sormuras/bach/jobs/362500362#L444

```
. ./install-jdk.sh -F 11 -L GPL
build.sh (2018-04-05)
  FEATURE = 11
  LICENSE = GPL
  ARCHIVE = openjdk-11-ea+7_linux-x64_bin.tar.gz
      URL = https://download.java.net/java/early_access/jdk11/7/GPL/openjdk-11-ea+7_linux-x64_bin.tar.gz [200]

--2018-04-05 07:49:43--  https://download.java.net/java/early_access/jdk11/7/GPL/openjdk-11-ea+7_linux-x64_bin.tar.gz
Resolving download.java.net (download.java.net)... 23.46.60.117
Connecting to download.java.net (download.java.net)|23.46.60.117|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 181183158 (173M) [application/x-gzip]
Saving to: â€˜openjdk-11-ea+7_linux-x64_bin.tar.gzâ€™


 0% [                                       ] 0           --.-K/s              
 5% [>                                      ] 9,251,383   41.6MB/s             
10% [===>                                   ] 19,297,463  44.7MB/s             
13% [====>                                  ] 25,174,016  38.8MB/s             
20% [======>                                ] 37,141,447  43.3MB/s             
26% [=========>                             ] 48,242,688  44.6MB/s             
32% [===========>                           ] 59,033,262  45.7MB/s             
38% [=============>                         ] 69,214,208  45.8MB/s             
44% [================>                      ] 79,727,934  46.2MB/s             
49% [==================>                    ] 90,185,728  46.5MB/s             
55% [====================>                  ] 100,671,488 46.5MB/s             
62% [=======================>               ] 113,254,400 47.4MB/s             
68% [=========================>             ] 123,740,160 47.4MB/s             
75% [============================>          ] 136,163,653 48.3MB/s             
81% [==============================>        ] 146,808,832 48.5MB/s             
86% [================================>      ] 157,294,592 48.6MB/s  eta 0s     
93% [===================================>   ] 169,176,509 49.0MB/s  eta 0s     
99% [=====================================> ] 180,363,264 49.8MB/s  eta 0s     
100%[======================================>] 181,183,158 49.9MB/s   in 3.5s   

2018-04-05 07:49:47 (49.3 MB/s) - â€˜openjdk-11-ea+7_linux-x64_bin.tar.gzâ€™ saved [181183158/181183158]

openjdk-11-ea+7_linux-x64_bin.tar.gz: gzip compressed data, from Unix, last modified: Fri Mar 30 03:14:13 2018
openjdk 11-ea 2018-09-18
OpenJDK Runtime Environment (build 11-ea+7)
OpenJDK 64-Bit Server VM (build 11-ea+7, mixed mode)
```

Would it suffice if I included a `JAVA_HOME = <...>` line?

sormuras: About to add an `echo ${JAVA_HOME}` as the last script command. With this, it possible to:

`JAVA_HOME=$(./install-jdk.sh -L GPL | tail -n 1)`

No more `source`-ing.

sormuras: Done via https://github.com/sormuras/bach/commit/d75685e693200cc0384972d8ff0497877385e4f7

Here is the result of `install-jdk-tests.sh`:
```java
install-jdk-tests.sh -- testing main variants in dry-run mode...

+ ./install-jdk.sh -D -L GPL -F 9
install-jdk.sh (2018-04-05)
  FEATURE = 9
  LICENSE = GPL
  ARCHIVE = openjdk-9.0.4_linux-x64_bin.tar.gz
      URL = https://download.java.net/java/GA/jdk9/9.0.4/binaries/openjdk-9.0.4_linux-x64_bin.tar.gz [200]

+ ./install-jdk.sh -D -L BCL -F 9
install-jdk.sh (2018-04-05)
  FEATURE = 9
  LICENSE = BCL
  ARCHIVE = jdk-9.0.4_linux-x64_bin.tar.gz
      URL = http://download.oracle.com/otn-pub/java/jdk/9.0.4+11/c2514751926b4512b076cc82f959763f/jdk-9.0.4_linux-x64_bin.tar.gz [302]

+ ./install-jdk.sh -D -L GPL -F 10
install-jdk.sh (2018-04-05)
  FEATURE = 10
  LICENSE = GPL
  ARCHIVE = openjdk-10_linux-x64_bin.tar.gz
      URL = https://download.java.net/java/GA/jdk10/10/binaries/openjdk-10_linux-x64_bin.tar.gz [200]

+ ./install-jdk.sh -D -L BCL -F 10
install-jdk.sh (2018-04-05)
  FEATURE = 10
  LICENSE = BCL
  ARCHIVE = jdk-10_linux-x64_bin.tar.gz
      URL = http://download.oracle.com/otn-pub/java/jdk/10+46/76eac37278c24557a3c4199677f19b62/jdk-10_linux-x64_bin.tar.gz [302]

+ ./install-jdk.sh -D -L GPL -F 11
install-jdk.sh (2018-04-05)
  FEATURE = 11
  LICENSE = GPL
  ARCHIVE = openjdk-11-ea+7_linux-x64_bin.tar.gz
      URL = https://download.java.net/java/early_access/jdk11/7/GPL/openjdk-11-ea+7_linux-x64_bin.tar.gz [200]

+ ./install-jdk.sh -D -L BCL -F 11
install-jdk.sh (2018-04-05)
  FEATURE = 11
  LICENSE = BCL
  ARCHIVE = jdk-11-ea+7_linux-x64_bin.tar.gz
      URL = https://download.java.net/java/early_access/jdk11/7/BCL/jdk-11-ea+7_linux-x64_bin.tar.gz [200]
```

Happy reviewing!

BanzaiMan:
Relying on the last line to deliver critical information seems rather brittle to me.

sormuras:
Would a `-X` flag work? It would surpress all output, except for `echo "${JAVA_HOME}"`. If the script fails before it could download and extract the JDK archive an error message is printed and a non-zero exit code returned. Sounds better?

sormuras:
I'd favor an explicit option that takes the final destination of the JDK as an argument:

```
JAVA_HOME=/usr/opt/jdk
./install-jdk.sh -J ${JAVA_HOME} -F ? -L GPL
PATH=${JAVA_HOME}/bin:${PATH}
```
`install-jdk.sh` then `mv`-es the extracted files to that destination folder.

What do you think, @BanzaiMan?

bbelovic:
@sormuras @BanzaiMan Any update on this pull request? When can we expect it on production and not only on staging environment?

sormuras:
Updated `install-jdk.sh` to support JDK 10.0.1 and archived feature releases.

BanzaiMan:
Sorry for the delay. I’ve been traveling and have not had a chance to refine and deploy it. Hopefully later this week.

sormuras:
Or this week? ;)

BanzaiMan:
`--silent` suppresses all output from `install-jdk.sh`. This is not very helpful, since we want to see what the script is doing. Perhaps it is agreeable to have an option to send installation progress to `STDERR`, and emit the `JAVA_HOME` value to `STDOUT`, so that we can retain the output from the installation process. How does that sound?

sormuras:
What about passing the destination directory as an option? That is already implemented and it emits all default output to stdout (no source-ing required):
```
JAVA_HOME=~/jdk 
./install-jdk.sh --target $JAVA_HOME
PATH=$JAVA_HOME/bin:$PATH
```

sormuras:
Pushed new version of `install-jdk.sh`. With the `--target` option everything's clear to roll, methinks.

BanzaiMan:
Looks good. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/720268#L449

sormuras:
Great! Shall I run the complete JDK matrix on staging, too?

sormuras:
Did it anyway: all 10 jobs report  ✔️ using `install-jdk.sh 2018-04-30` with `--target $JAVA_HOME`

Stephan202:
First off: this is really cool :)

Maybe I missed it in this PR, but it appears there is no caching. If Travis CI users start using this feature at a large scale, won't that cause issues upstream? There may also be an opportunity to speed things up further, but my primary concern is with `https://download.java.net` simply blocking Travis CI IPs.

BanzaiMan:
@Stephan202 Thanks for raising the issue. That is, indeed, a valid concern.

@sormuras How is `install_jdk.sh` designed to behave if the `--target` already has a valid (in some sense of the word) JDK installation? If it bails out early, then the download is going to be an issue while we wait for the next build images to contain the newer JDKs.

This also poses a question as to how it handles on-the-fly installations in case there are newer JDKs available than the ones already installed; this is a worthwhile considerations for, say, EA releases of the next generation JDKs.

sormuras:
The `install-jdk.sh` script was designed to just download and extract the latest-and-greatest JDK into a clean (CI) environment. If the `--target` directory already exists, the script exits before downloading any large archive.

I don't want to support more complex scenarios within the script. There are better solutions available, but I don't how if/whether Travis CI may make use of them: http(s) caching proxy techniques like http://www.squid-cache.org does is one option. The archived and releases JDK uri are very stable, the EA releases change every 1-3 weeks.

Switching to a more complete Java Version Manager like https://github.com/shyiko/jabba is an option, too.

sormuras:
`install-jdk.sh` does use the `--timestamping --continue` options of `wget` to minimize the download activities already. So, if a matching `.tar.gz` archive is present, only an up-to-date check will be performed.

sormuras:
@BanzaiMan Could the downloaded `.tar.gz` file from a previous build be cached or be "injected" by some other means?

sormuras:
@BanzaiMan Ping. :)

sormuras:
Smoke-test is green, too: 10x ✅

https://staging.travis-ci.org/sormuras/sormuras.github.io/builds/721219

BanzaiMan:
@sormuras Do you have a list of archives `install_jdk.sh` can pull? (exclude `ea` releases for the purpose of this question)

sormuras:
Not explicitly. But it compiles like this:

**OpenJDK**

- early-access OpenJDK build from http://jdk.java.net (/11 at the moment)
- current OpenJDK from http://jdk.java.net (/10 at the moment)
- latest archived OpenJDKs from http://jdk.java.net/archive (openjdk-9.0.4_linux-x64_bin.tar.gz)

**Oracle JDK***

- early-access Oracle JDK build from http://jdk.java.net (/11 at the moment)
- Oracle jdk-10.0.1_linux-x64_bin.tar.gz
- Oracle jdk-9.0.4_linux-x64_bin.tar.gz

spec/build/script/shared/jdk.rb
```diff
@@ -36,6 +40,19 @@
     end
   end
 
+  context "jdk is set to oraclejdk10" do
```
sormuras: `oraclejdk10` or `11`? The code below uses `11`

sormuras:
`11-ea+13` is now `11-ea+14` :)

sormuras:
> Instead, maybe install_jdk.sh can: [...]

`install-jdk.sh` uses `--timestamping` and `--continue`. That prevents downloading already present archives. So, if the user caches the `JAVA_HOME` directory, everything's fine.

> As for the macOS support: could you elaborate what you mean by "if you unset -f cd first"?

See https://github.com/travis-ci/travis-ci/issues/8703#issuecomment-392415754 ... if I don't `unset -f cd` before launching the script, the `cd` commands used in it don't work as expected.

BanzaiMan:
@sormuras Thanks for the comments. `$JAVA_HOME` would be rather huge, no? Caching all of it seems wasteful to me. Did I miss something?

BanzaiMan:
@sormuras Copy! So, I suppose we can set, for example, `$TRAVIS_BUILD_DIR/tmp` for such storage and advise users to cache that?

BanzaiMan:
@sormuras Affirmative. https://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

sormuras:
But now, if many users will use caching, the cache storage will "explode"... especially when `-ea` versions are cached.

BanzaiMan:
Seems to me that is a reasonable cost.

sormuras:
So. From my side: **"All systems: Go!"** 🚀

sormuras:
The `--verbose` option would be great to have!

And there's an error when running with `--target <dir>` on `osx`: https://staging.travis-ci.org/sormuras/sormuras.github.io/jobs/722374

The `<dir>` passed via `--target` is no longer equivalent to `JAVA_HOME`. You need to append `/Contents/Home` to `<dir>`

When sourcing the script, this block handles it:

```
    # Fix path to JDK Home...
    if [[ "$OSTYPE" == "darwin"* ]]; then
        target="${target}/Contents/Home"
    fi
```

Two solutions: 
- [ ] either you add a similar block to your setup script
- [ ] or will move all files from `/Contents/Home` to `<dir>`

lib/travis/build/script/shared/jdk.rb
```diff
@@ -44,6 +68,38 @@ def uses_java?
...
+            sh.cmd "curl -sf -O https://#{app_host}/files/install-jdk.sh"
+          end
+
+          def install_jdk_args(jdk)
+            args_for = {
+              # OpenJDK
+              'openjdk-ea'   => '-L GPL',
+              'openjdk9'     => '-F 9  -L GPL',
+              'openjdk10'    => '-F 10 -L GPL',
+              'openjdk11'    => '-F 11 -L GPL',
```
szpak: You may want to use `-C` for all openjdk builds - https://github.com/sormuras/bach/pull/12

sormuras: Thought about that too. But it would change the "default" setup.

Users just configuring `jdk:openjdkX` on Travis CI will be surprised if their local openjdkX installation behaves differently.

Documenting the `-C` switch when "openjdk" is selected is a good thing. Like was done with the cache configuration warning.

sormuras:
About to release version **2018-05-30** in a minute.  `/Contents/Home` is gone now.

```
Content of target directory:
bin		include		legal		libjli.dylib	release
conf		jmods		lib		man
```
`libjli.dylib` is an artifact from the archive. It doesn't prevent the extracted JDK from working with my test project. I guess it's native library ... used for?!

BanzaiMan:
Upon further reflection and testing, caching would not work, unfortunately. The reason here is that JDK is configured before cache is set up, and the cache directory will always be empty. Rearranging these steps is option, but I am a bit reluctant to take it, at least for now.

I am leaning towards releasing it and observing how the usage looks.

sormuras:
> I am leaning towards releasing it and observing how the usage looks.

Which is okay, too.

Now, I understand. Using `install-jdk.sh` in the `install`-phase works. But not when it is invoked by the default "Travis boot script".

sormuras:
🍰 🍻 🍸 and ☕️ !

sormuras:
Couldn't resist @BanzaiMan ... roll-out happend mid-flight, methinks!

https://travis-ci.org/sormuras/sormuras.github.io/builds/385764894

BanzaiMan:
@sormuras Probably. The macOS issue might need a deeper look; I have not done anything with the `cd()` you mentioned.

Also, we would want the download to be less chatty: https://travis-ci.org/BanzaiMan/travis_production_test/jobs/385765530#L421

sormuras:
Why does `wget` print all those lines? 

Here https://travis-ci.org/sormuras/sormuras.github.io/jobs/385764901 it is just a single line that is updated in-place.

sormuras:
Because of the `dist: trusty` or `sudo: false` worker?

sormuras:
https://travis-ci.org/sormuras/sormuras.github.io/builds/385773616

Linux builds are all fine!

2 issues on Mac OS X
- `openjdk8` is not supported on `osx`, right? I.e. not pre-installed (because not available, I guess).
- `oraclejdk-ea` is a `.dmg` archive, not a `.tar.gz` one. Have to fix it, if possible. Workaround: use the `openjdk-ea` build

sormuras:
Looks good: https://travis-ci.org/sormuras/sormuras.github.io/builds/385780432

> Also, we would want the download to be less chatty

I'll set the `wget` to `quite` by default. It only will be emitted when `--verbose` is set.

sormuras:
> The current process is to fetch the latest `install-jdk.sh` at the application deployment time. This process is currently manual also, but should generally happen when this repository gets an update.

When is the next update? :)
Is there a way to trigger/request an update?
Where can I find a log of `travis-build` update?

Applied a work-around to `install-jdk.sh` fix with https://github.com/sormuras/bach/commit/69abfbfcc365aa3c992d8a641b7e5244bc8ad77a for https://twitter.com/sormuras/status/1002226003647791105

sormuras:
Updated `install-jdk.sh 2018-07-13` with a fix for `jdk: openjdk11` ending up being `oraclejdk11` at the moment.

See issue at: https://travis-ci.org/filiphr/mapstruct-examples/jobs/403935366#L413-L414
Fixed version at: https://travis-ci.org/sormuras/sormuras.github.io/jobs/403949235#L457

sormuras:
Seems like `travis-build` picked-up the new version.

https://travis-ci.org/junit-team/junit5/jobs/404767934#L408

👍

sormuras:
Now, JDK **10.0.2** is released. Updated `install-jdk.sh` to `2018-07-17` -- can you please fetch the current version again? Thanks in advance.

PS: Is there a better way to ping or not to ping about such updates?

mmoayyed:
I'd second the move to JDK 10.0.2. Not sure if this is related, but Apereo CAS builds suddenly started to fail this afternoon MST, not able to download JDK 10.

Failing build:
https://travis-ci.org/apereo/cas/builds/405102222

Output:

```
...

$ export JAVA_HOME=$HOME/oraclejdk10
$ bash install-jdk.sh -F 10 -L BCL --target $JAVA_HOME --workspace $HOME/.cache/install-jdk
install-jdk.sh 2018-07-13
Variables:
  feature = 10
  license = BCL
       os = linux-x64
      url = http://download.oracle.com/otn-pub/java/jdk/10.0.1+10/fb4372174a714e6b8c52526dc134031e/jdk-10.0.1_linux-x64_bin.tar.gz
   status = 302
  archive = /home/travis/.cache/install-jdk/jdk-10.0.1_linux-x64_bin.tar.gz
Downloading JDK from http://download.oracle.com/otn-pub/java/jdk/10.0.1+10/fb4372174a714e6b8c52526dc134031e/jdk-10.0.1_linux-x64_bin.tar.gz...
The command "bash install-jdk.sh -F 10 -L BCL --target $JAVA_HOME --workspace $HOME/.cache/install-jdk" failed and exited with 8 during .
Your build has been stopped.
```

Travis CI issue? Oracle issue? Bad link?

If I can do anything to help resolve this, please let me know.

sormuras:
According to what @BanzaiMan said at https://github.com/travis-ci/travis-build/pull/1347#issuecomment-378322547 :

> The current process is to fetch the latest `install-jdk.sh` at the application deployment time. This process is currently manual also, but should generally happen when this repository gets an update.

Somebody needs to trigger an update for `travis-build`.

The fix was applied via: https://github.com/sormuras/bach/commit/2800b56ecfab66d640b3f0afbfaee194c368f883
You may resort to _raw_ mode which is described here https://sormuras.github.io/blog/2018-06-22-jdk-matrix.html#configuration -- or just wait until Travis CI fetched version `2018-07-17` of `install-jdk.sh`.

sormuras:
>  Dynamic Oracle JDK installations are currently unavailable due to a problem with our upstream provider. In the meantime, container-based builds come with JDKs preinstalled.

Having preinstalled JDKs would be really great! Especially when it comes to GA releases. Would result in faster builds and less noise.

`install-jdk.sh`'s focus always were EA builds, anyway.

PR #1348, opened 2018-04-03
Disable all non-core APT sources by default

meatballhat:
and remove various hacks/workarounds that accomplished a subset of this
functionality over the years ...


meatballhat:
Accompanying APT source additions: https://github.com/travis-ci/apt-source-whitelist/pull/380

lib/travis/build/appliances/disable_apt_sources.rb
```diff
@@ -0,0 +1,27 @@
+require 'travis/build/appliances/base'
+
+module Travis
+  module Build
+    module Appliances
+      class DisableAptSources < Base
+        def apply
+          sh.raw <<~BASH
```
BanzaiMan: I agree with the spirit of this PR, but I do have reservations about merging it right away.

This is going to break those builds that rely on these sources being available, so we should start with a warning, and give some time to adjust; e.g.,

    We will remove APT sources at YYYYMMDD. If your builds rely on them,
    we suggest adding them with `apt` addon. For more information, read DOC_LINK

lib/travis/build/appliances.rb
```diff
@@ -35,10 +32,10 @@
 require 'travis/build/appliances/validate'
 require 'travis/build/appliances/npm_registry'
 require 'travis/build/appliances/update_rubygems'
-require 'travis/build/appliances/update_mongo_arch'
 require 'travis/build/appliances/apt_get_update'
 require 'travis/build/appliances/no_world_writable_dirs'
 require 'travis/build/appliances/ensure_path_components'
+require 'travis/build/appliances/disable_apt_sources'
```
edmorley: Should this be before the `apt_get_update` step?

edmorley:
Hi!

I'm presuming this PR is to help avoid the "some APT source is having downtime and causes error spew at the top of Travis jobs" issue, plus presumably also will speed up the `apt-get update` invocation? If so, sounds like a great improvement :-)

Something that's quite frustrating at the moment, is the amount of job setup time that's spent in travis-build, often unnecessarily so for newer images (eg cleanup that only really applies to older images). For example try running a `language: bash \n build: true \n script: true` for both GCE and container, and see just how long it takes (even ignoring boot time). For say linter jobs that use the cache and are fast running, it's a significant amount of the job time.

One of the culprits is the `apt-get update` step - it would be great if it could be removed for newer images. Perhaps if `packer-templates` and/or `travis-cookbooks` could purge the `/etc/apt/sources.list.d/` directory and then run `apt-get update` after, `travis-build` could use the logic of "if /etc/apt/sources.list.d/ is empty, then this is a newer image that doesn't need an `rm -rf` and `apt-get update` fixup"?

Thoughts? :-)

joepvd:
I believe we can start to prune 3rd party repositories with Xenial. We have promised our customers a stable Trusty image, so I think it would be bad style to undermine its stability by means of travis-build. See [this post](https://blog.travis-ci.com/2017-12-01-new-update-schedule-for-linux-build-images) for the promises made.

I guess we will have to live with what we have decided before until we EOL Trusty, and occasionally react to broken PPAs in a hurried manner, and live with the extra time from the initial `apt-get` dance.

meatballhat:
Closing this (for now?) as it's more invasive than any solution we'd want.

PR #1350, opened 2018-04-05
Set strict-ssl to false for Node.js 0.8.x

dougwilson:
So this is the same as #188 , but extends to Node.js 0.8.x now. Old versions of npm included in these versions have a specific CA embedded in npm and now that it's changed no longer validates.

Ref: https://github.com/npm/npm/issues/20191


BanzaiMan:
Thank you!

PR #1351, opened 2018-04-06
Change R build URL to the cloudfront URL

jimhester:
Rather than being served directly from S3 we now serve these files from
a CloudFront distribution; which should make fetches faster and cheaper.


jimhester:
Not sure what is going on with the Travis results, but I don't believe they are due to issues with this change. Please let me know if this is incorrect.

jimhester:
I have fixed the conflicts and the build is now passing, it would be great to have this reviewed and merged, thanks!

jimhester:
I again fixed the conflicts for the PR, it would be good to have it merged!

PR #1352, opened 2018-04-08
update go 1.10.x to use go 1.10.1

jadr2ddude:
Go 1.10.1 is the latest 1.10 series go release. In the gopher slack, the issue was brought up that this does not default to 1.10.1 which has bugfixes. This is a pretty straightforward change.


shurcooL:
I suspect `"1"`, `"1.x"`, and `"1.x.x"` should also be updated from 1.10 to 1.10.1.

jadr2ddude:
Ok i will do that in a minute

jadr2ddude:
Thanks!

PR #1354, opened 2018-04-09
Invoke version alias update task during assets:precompile

BanzaiMan:


BanzaiMan:
😆 I removed that bit.

PR #1355, opened 2018-04-18
use npm ci if available

btkostner:
This should run `npm ci` instead of `npm install` when `npm` version is above v5.8.0. This is much faster and has added benifits of using exact versions from `package-lock.json`.

Changelog for the npm version:
https://github.com/npm/npm/releases/tag/v5.8.0

### DIFF
--- a/lib/travis/build/script/node_js.rb
+++ b/lib/travis/build/script/node_js.rb
@@ -218,7 +218,11 @@ module Travis
 
           def npm_install(args)
             sh.fold "install.npm" do
-              sh.cmd "npm install #{args}", retry: true
+              sh.if "$(vers2int `npm -v`) -gt $(vers2int 5.7.1)" do
+                sh.cmd "npm ci #{args}", retry: true
+              sh.else do
+                sh.cmd "npm install #{args}", retry: true
+              end
               sh.if "$(vers2int `npm -v`) -gt $(vers2int #{NPM_QUIET_TREE_VERSION})" do
                 sh.cmd "npm ls", echo: true, assert: false
               end


BanzaiMan:
Hello, there. Could you point to the documentation that discusses this subcommand?

And look at failing specs, please?

btkostner:
@BanzaiMan Hello!

Documentation can be found here:
https://docs.npmjs.com/cli/ci#description

The travis error seems to be unrelated, as it errored out before even running tests:
https://travis-ci.org/travis-ci/travis-build/jobs/367974287

joshk:
This is an awesome PR! Thank you!

lib/travis/build/script/node_js.rb
```diff
@@ -222,7 +222,12 @@ def iojs_3_plus?
 
           def npm_install(args)
             sh.fold "install.npm" do
-              sh.cmd "npm install #{args}", retry: true
+              sh.if "$(vers2int `npm -v`) -gt $(vers2int 5.8.0)" do
```
btkostner: It's worth noting that I'm not 100% sure how the `vers2int` command works here, and this _might_ be wrong

finnp:
This is awesome! However this should also check for a `package-lock.json`, because otherwise the `npm ci` command will fail.

finnp:
Also maybe @zkat has some further input here ☺️ ☝️

BanzaiMan:
Looking at the documentation, this needs to check for the presence of `package-lock.json` or `npm-shrinkwrap.json` (as @finnp points out).

Caching of `node_modules` may or may not be significant. Presumably `node_modules` will be removed when we move to `npm ci`, but the documentation is not clear on what `node_modules` looks like _after_ `npm ci`. Depending on the effects of `npm ci` on `node_modules`, the reasonable behavior can range from "just ignore it" to "ask users to drop `node_modules` from cache".

zkat:
You really shouldn't cache n_m if you're using npm ci

BanzaiMan:
@zkat Could you clarify what we can expect of `node_modules` after `npm ci`? Is it removed and not re-created?

zkat:
`npm ci` blows away n_m before installing so caching it would only make it slower.

A much better alternative is to cache ~/.npm by default imo

BanzaiMan:
@zkat Thanks for the clarification. Sounds like we should just educate users.

lib/travis/build/script/node_js.rb
BanzaiMan: You'd want `-ge`, not `-gt`.

BanzaiMan: You'd also need additional conditions, as discussed in this PR.

zkat:
@BanzaiMan having a cache: npm option would also help. It's a bit verbose to cache the directory right now?

Either way, `npm ci` will warn if there's an existing node_modules, so if anyone ends up bothering actually looking at it, they'll catch it :)

lib/travis/build/script/node_js.rb
```diff
@@ -208,6 +210,18 @@ def node_0_9?
           def use_npm_cache?
             Array(config[:cache]).include?('npm')
           end
+        
+          def has_package_lock_file?
+            sh.if "-f package-lock.json"
+          end
+        
+          def has_shrinkwrap_file?
+            sh.if "-f npm-shrinkwrap.json"
```
btkostner: Is this valid ruby?

finnp: Ruby only generates a bash script. So I don't think this will work. 

You could instead make nest the `sh.if`s here: https://github.com/btkostner/travis-build/blob/d6d0d6ec2b22cee1a6473369a0e701923775187d/lib/travis/build/script/node_js.rb#L239.

So maybe somethin like this would work:
``` ruby 
              sh.if "$(vers2int `npm -v`) -gt $(vers2int #{NPM_CI_CMD_VERSION})" do
                     sh.if "-f npm-shrinkwrap.json -o -f package-lock.json" do
                         sh.cmd "npm ci #{args}", retry: true
                     else
                         sh.cmd "npm install #{args}", retry: true
                    end
              end
```

Alternatively making the first `sh.if` more complex would probably be nicer.

lib/travis/build/script/node_js.rb
```diff
@@ -222,7 +224,16 @@ def iojs_3_plus?
 
           def npm_install(args)
             sh.fold "install.npm" do
-              sh.cmd "npm install #{args}", retry: true
+              sh.if "$(vers2int `npm -v`) -ge $(vers2int #{NPM_CI_CMD_VERSION})" do
+                sh.if "-f npm-shrinkwrap.json -o -f package-lock.json" do
+                  sh.cmd "npm ci #{args}", retry: true
+                else
```
finnp: This should have been `end <newline> sh.else do`. My bad 😬

btkostner: That's my bad. Ruby isn't my strong suit.

lib/travis/build/script/node_js.rb
```diff
@@ -227,7 +227,8 @@ def npm_install(args)
               sh.if "$(vers2int `npm -v`) -ge $(vers2int #{NPM_CI_CMD_VERSION})" do
                 sh.if "-f npm-shrinkwrap.json -o -f package-lock.json" do
```
finnp: Okayyy seems like it should be `-f npm-shrinkwrap.json || -f package-lock.json` instead..

finnp:
Nice, now the produced script looks valid ☺️ Looks like one of the tests errored though, maybe it just needs to be re-run?

btkostner:
@finnp yeah, looks like ruby errored. Unrelated to this PR though.

lib/travis/build/script/node_js.rb
```diff
@@ -222,7 +224,17 @@ def iojs_3_plus?
 
           def npm_install(args)
             sh.fold "install.npm" do
-              sh.cmd "npm install #{args}", retry: true
+              sh.if "$(vers2int `npm -v`) -ge $(vers2int #{NPM_CI_CMD_VERSION})" do
```
christophwitzko: I would suggest to merge the following `if` in this one with ```sh.if "$(vers2int `npm -v`) -ge $(vers2int #{NPM_CI_CMD_VERSION}) && (-f npm-shrinkwrap.json || -f package-lock.json)" do```.

BanzaiMan:
This has been deployed to https://staging.travis-ci.org/.

finnp:
👉 https://github.com/travis-ci/docs-travis-ci-com/pull/1866

BanzaiMan:
Thanks, everyone! This has been deployed.

PR #1373, opened 2018-05-06
R: Match RStudio v1.2 embedded pandoc version of 2.2

coatless:
Increments the pandoc version to 2.2 from 1.15.2, which is embedded in RStudio v1.2.

This change was brought about because pandoc 1.15.2 is too low to be used with `bookdown`.

Note: RStudio v1.1 has an embedded version of pandoc of 1.17.2


jimhester:
Could we add a test in the spec file to make sure the conditional works as expected for macOS versions < 1.19.2.2

```ruby
 data[:config][:os] = 'osx'
 data[:config][:pandoc_version] = '1.15.2' # or whatever before 1.19.2.2
 should include_sexp ...
```

coatless:
@jimhester add a download spec for pandoc 1.19.2.1, which was the last version that used the osx variant. 

However, the specs for macOS builds aren't being installed? Just downloaded? 

Any reason the packages aren't being installed via CLI like the ubuntu counterparts? 

```bash
sudo installer -pkg package.pkg -target /
```

jimhester:
They are being installed at https://github.com/travis-ci/travis-build/pull/1373/files#diff-5b7e1aa01f1c6f058df6e0b9e2486c62L482

coatless:
@jimhester no, within the spec file, e.g.

https://github.com/travis-ci/travis-build/blob/cbe49ea239ab37b9b38b5b44d287b7ec7a108c16/spec/build/script/r_spec.rb#L92-L98

vs.

https://github.com/travis-ci/travis-build/blob/cbe49ea239ab37b9b38b5b44d287b7ec7a108c16/spec/build/script/r_spec.rb#L118-L125

The latter line includes an actual `dpkg -i`. Maybe I am misunderstanding the test environment in `r_spec.rb`

jimhester:
We can add a test for the installation command in the spec if you would like, it can't hurt I suppose.

coatless:
@jimhester let's leave that for another PR as it is outside the focus of this one.

PR #1376, opened 2018-05-09
r: add feature to show logs on success

jeroen:
Currently we only get to see log files if there was an error. However I would like to be able to manually inspect the log files even if there was no error.  This PR adds that.

@jimhester

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -31,6 +31,7 @@ module Travis
           bioc_required: false,
           bioc_use_devel: false,
           disable_homebrew: false,
+          dump_logs: false,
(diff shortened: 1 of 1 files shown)


jimhester:
We actually had this, but it was causing builds to spuriously fail when there are too much output, see https://github.com/travis-ci/travis-build/pull/1283

jeroen:
Right, so I made it optional and disabled by default.

jimhester:
Ok, that makes sense, LGTM!

jeroen:
OK thanks all is good 🚀🚀

jimhester:
It would be great to get this reviewed and merged

PR #1379, opened 2018-05-11
Cache npm client directory with cache: npm

finnp:
This adds the ability to cache the npm directory/cache with `cache: npm`. Similar to [`cache: yarn`](https://docs.travis-ci.com/user/caching/#yarn-cache).

It looks like there was some undocumented option before, where you could also set an npm proxy via something like:
```yml
cache:
   - npm
hosts:
  npm_cache: https://myregistry.com
```

I removed that to avoid confusion. I also did a [quick search](https://github.com/search?utf8=%E2%9C%93&q=npm_cache+extension%3A.yml&type=Code) and it doesn't seem to be used.


joshk:
@BanzaiMan 👋 would you have some time to look into this?

finnp:
Let me know if I should update the PR to teh current master :)

PR #1382, opened 2018-05-16
Remove libmagick++-dev from dependencies

jimhester:
This had a conflict with libpng-dev which caused all R builds to fail

I also updated the dependencies to reflect the current state of

    apt-cache depends r-base-dev

Which added libpcre3-dev and libicu-dev

Fixes https://github.com/travis-ci/travis-ci/issues/9625


jimhester:
I am not sure why the tests are failing but it looks unrelated to these changes. @BanzaiMan, it would be great to get this reviewed and deployed sooner rather than later, as all R builds are currently failing due to this.

PR #1410, opened 2018-06-13
Fix postgresql addon for Xenial

joepvd:
- Adds convenience methods to `data` top determine what type of system this is running on
- Use systemd on Xenial, Upstart on previous Ubuntus
- Do not run addon on non-linux


joepvd:
Problem it tries to resolve: Get the postgres you specify, rather than always 9.6: https://travis-ci.org/joepvd/travis-experiment/jobs/388177347#L443

This is how it looks like on staging: https://staging.travis-ci.org/joepvd/travis-staging-test/builds/722684

PR #1414, opened 2018-06-22
Fix biocheck problems

joepvd:
Removes error: "Malformed version number string release"


LiNk-NY:
Greetings @joepvd @joecorcoran, 
Thanks for picking this up! I have a pending PR for review at travis-ci/travis-yml#56. Thanks!

PR #1424, opened 2018-06-29
Retry dart sdk curl requests on connection refused

jakemac53:
See https://travis-ci.org/dart-lang/build/jobs/398406858 for an example build with this failure, this is the primary error we actually see when this curl request is going to fail.

cc @kevmoo


kevmoo:
CC @BanzaiMan

joepvd:
Hi! The the `retry-connrefused` switch [was introduced](https://curl.haxx.se/changes.html#7_52_0) in `curl` 7.52.0, which is not present on the precise, trusty, and xenial images.

Would be happy to merge a solution based on `travis_retry` and `curl --fail`.

jakemac53:
@joepvd thanks, I updated it to use `travis_retry` and `--fail` (but do note I don't have a way of testing these changes).

Is the `--fail` part necessary? It seems like it will make it more difficult to diagnose issues since we won't be able to see why the curl request failed if/when it fails in the future.

jakemac53:
closing as this is quite stale now, we do still occasionally see some flakes when the download fails but it isn't a huge issue

PR #1426, opened 2018-07-03
Julia: define "JULIA_PROJECT" environment variable

simonbyrne:
Fixes https://github.com/travis-ci/travis-ci/issues/9837

Needed to run on latest nightlies.

CC: @travis-ci/julia-maintainers, @KristofferC

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -20,6 +20,7 @@ module Travis
 
           sh.export 'TRAVIS_JULIA_VERSION', config[:julia].to_s.shellescape,
             echo: false
+          sh.export 'JULIA_PROJECT', "@."
         end
 
         def setup


KristofferC:
LGTM

simonbyrne:
@joepvd where do tests go? as for docs, we're in the process of changing how packages work, and this is a bit of stopgap that will work across different versions. once completed we can remove this and standardise on something nicer.

joepvd:
Appreciate the extra context!

This would be the appropriate place for Julia specific tests: https://github.com/travis-ci/travis-build/blob/master/spec/build/script/julia_spec.rb

For the future, you might want to have a look how different default `install` and `script` steps are set up based on details in the repository for JVM based languages: [jvm.rb](https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/shared/jvm.rb).

simonbyrne:
Thanks. Unfortunately I don't know enough ruby to figure out how those tests work.

I think this is ready to go on our end.

PR #1428, opened 2018-07-03
C#: embed the pgp keys for mono and dotnet repos instead of using Ubuntu keyserver

akoeplinger:
Should help with reliability.


joepvd:
Thanks! This branch was deployed to staging. Could you run a smoke test in https://staging.travis-ci.org and link back the results?

I am happy to ignore the false test negatives for now, I'll see to it that the experience improves soon. Sorry about that!

akoeplinger:
@joepvd I've tested this on staging and it works as expected so I think this is ready :)

PR #1429, opened 2018-07-04
Inform user when no Gemfile is found

joecorcoran:
https://secure.helpscout.net/conversation/610609190/74157?folderId=30784

### DIFF
--- a/lib/travis/build/script.rb
+++ b/lib/travis/build/script.rb
@@ -57,11 +57,12 @@ module Travis
       include Module.new { Stages::STAGES.map(&:name).flatten.each { |stage| define_method(stage) {} } }
       include Appliances, DirectoryCache, Deprecation, Template
 
-      attr_reader :sh, :data, :options, :validator, :addons, :stages
+      attr_reader :sh, :raw_data, :data, :options, :validator, :addons, :stages
       attr_accessor :setup_cache_has_run_for
 
       def initialize(data)
-        @data = Data.new({ config: self.class.defaults }.deep_merge(data.deep_symbolize_keys))
+        @raw_data = data.deep_symbolize_keys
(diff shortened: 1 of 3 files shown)


grosser:
hmm that is a bit better, but ideally it would be:

```ruby
if config[:gemfile] || sh.if gemfile?
```

since if a user requests a gemfile and it does not exist, then it should fail (do what the user told it to do), since the user expected the file to exist/claimed the file exists

joecorcoran:
Thanks @grosser – I think the latest commit does the right thing now. We test for the presence of the user-provided Gemfile as early as possible and fail if it's not there.

lib/travis/build/script.rb
```diff
@@ -57,11 +57,12 @@ def defaults
...
       include Appliances, DirectoryCache, Deprecation, Template
 
-      attr_reader :sh, :data, :options, :validator, :addons, :stages
+      attr_reader :sh, :raw_data, :data, :options, :validator, :addons, :stages
       attr_accessor :setup_cache_has_run_for
 
       def initialize(data)
-        @data = Data.new({ config: self.class.defaults }.deep_merge(data.deep_symbolize_keys))
+        @raw_data = data.deep_symbolize_keys
+        @data = Data.new({ config: self.class.defaults }.deep_merge(self.raw_data))
```
grosser: should add rubocop so it tells you that the `self.` is not necessary here :)

lib/travis/build/script/shared/bundler.rb
```diff
@@ -71,6 +83,14 @@ def cache_slug
 
         private
 
+          def user_provided_gemfile
+            raw_data[:config] && raw_data[:config][:gemfile]
```
grosser: `raw_data.dig :config, :gemfile` if this is 2.3+

PR #1431, opened 2018-07-05
Use the right names for mongod(b)

joepvd:


cotsog:
@joepvd is this compatible with older images e.g. `group: deprecated-*` ?

joepvd:
@cotsog It should be. I am not aware that the name of the service has changed within an Ubuntu release.

grooverdan:
You've left the xenial name as 'mongod'.  The xenial has a mongodb@service name https://packages.ubuntu.com/xenial/amd64/mongodb-server/filelist

mongod is definitely wrong: https://travis-ci.org/grooverdan/travis-test/jobs/400759538#L399

joepvd:
Hmm... The amd64 xenial images use http://repo.mongodb.org/apt/ubuntu/dists/xenial/mongodb-org/4.0/multiverse/binary-amd64/, whereas the ppc64le image uses the default Ubuntu repository.

On the mongodb repository, the provided service-file looks like this:
`mongodb-org-server_4.0.0_amd64/lib/systemd/system/mongod.service`

Looks like an additional check for the architecture needs to be added.

grooverdan:
thanks for the rather large fix. I suspect architecture branching of service is something rather avoided so I'll try to see why there isn't ppc64le in the mongo repository. This will solve the moving beyond 2.6.10 too.

PR #1432, opened 2018-07-06
Tweaks for ppc64le

joepvd:
- Add extra aliases for postgres and mongodb, so these are not easily typo'd in the `services` list
- Start mongodb differently when the architecture is PPC64LE
- Detect architecture at the start of the build and store it as `$TRAVIS_ARCH`

### DIFF
--- a/lib/travis/build/appliances/services.rb
+++ b/lib/travis/build/appliances/services.rb
@@ -13,7 +13,10 @@ module Travis
           'memcache'     => 'memcached',
           'neo4j-server' => 'neo4j',
           'rabbitmq'     => 'rabbitmq-server',
-          'redis'        => 'redis-server'
+          'redis'        => 'redis-server',
+          'postgres'     => 'postgresql',
+          'mongo'        => 'mongodb',
+          'mongod'       => 'mongodb'
         }
         TEMPLATES_PATH = File.expand_path('../../templates', __FILE__)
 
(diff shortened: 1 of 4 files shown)


grooverdan:
looks good with the aliases.

`TRAVIS_ARCH` seems useful, I'd almost started to use `TRAVIS_OS_NAME` before this was linux normalised.

I'm looking to see what can be done to get a newer apt version for mongodb on ppc64le.

Note in x86_64 trusty that using apt to specify a later (4.0) mongodb version will still result in the base image 3.4.10 version running: https://travis-ci.org/grooverdan/travis-test/jobs/401564974

grooverdan:
Seems I can grab the ppc64le 3.6 version bionic easy enough and with this change it should start:
https://travis-ci.org/grooverdan/travis-test/jobs/402005365

joepvd:
:thinking: 

Maybe this would be a better solution?

```
sudo systemctl start mongodb || sudo systemctl start mongod
```

Would need some cosmetics, so it does not look so convoluted in the build log.

grooverdan:
To go fully compatible more is needed. Probably something like the postgresql addon detecting if packages where specified in the apt plugin and the determining which service from that.

As it stands a base image may have either a mongo or distro package, which can be upgraded by the user to a later mongo or distro package.

PR #1433, opened 2018-07-06
Decide on apt options depending on apt version

joepvd:
Decide the apt options based on the version of apt
Test build: https://staging.travis-ci.org/joepvd/travis-staging-test/builds/723980


joecorcoran:
Not familiar with the issue, but looks good to me. 👍

PR #1437, opened 2018-07-23
first prototype for tracing in build.sh

igorwwwwwwwwwwwwwwwwwwww:
This is still a work-in-progress.

Some things we may want to do:

* Introduce a [stack](https://gist.github.com/bmc/1323553) of spans
* Instrument more stuff, especially the appliances, and maybe stages on a high level

refs https://github.com/travis-ci/reliability/issues/142

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -13,7 +13,7 @@ GIT
 
 GIT
   remote: https://github.com/travis-ci/travis-github_apps
-  revision: b589c48935a020084347e6206e96044e0c0ce6ac
+  revision: b8f41e53ae51e7633a9e5cfa61525068064b3402
   specs:
     travis-github_apps (0.2.0)
       activesupport (>= 3.2)
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,5 +5,7 @@ services:
       context: .
       args:
       - GITHUB_OAUTH_TOKEN
+    environment:
+    - RACK_ENV=production
     ports:
     - 4000:4000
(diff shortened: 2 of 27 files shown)


igorwwwwwwwwwwwwwwwwwwww:
To test locally:

```
bundle install
bundle exec rackup -Ilib
```

Generate payload (write into `job_payload.json`):

```
heroku run -a travis-scheduler-staging ruby script/generate-job-payload.rb 725023
```

Generate build.sh:

```
curl -i -XPOST -d@job_payload.json localhost:9292/script
```

igorwwwwwwwwwwwwwwwwwwww:
I changed this to pre-compute most of the data going into the trace file at build time, to avoid complex bash logic for escaping and state. This means that even the span ids and the tree structure is pre-computed, and makes some assumptions about the order of operations. The only thing that happens in bash is interpolating the timestamp and the exit code, and appending the result to a file.

One tricky bit was that the individual steps `before_script`, `script`, etc. are defined as individual functions, and as such are not nested within the code that calls those functions. So I had to pre-allocate a root span and add a special case for that, so that those steps become children of the root.

igorwwwwwwwwwwwwwwwwwwww:
Looks like this still needs some work 🤦‍♂️ 

```
/home/travis/build.sh: line 1: travis_trace_span: command not found
/home/travis/build.sh: line 907: warning: here-document at line 15 delimited by end-of-file (wanted `EOFUNC')
```

igorwwwwwwwwwwwwwwwwwwww:
This is currently only producing a partial trace, not sure yet what is up. Will continue with this next week. :)

igorwwwwwwwwwwwwwwwwwwww:
This is working correctly now!

PR #1441, opened 2018-07-31
Do not update repository list in the background

MartinNowack:
Updating the repository list might take some time but it should finish before packages from the repository are installed.
This small but important patch does it in the foreground.

The following is an example to reproduce the issue.
```
        apt:
          update: true
          sources:
            - llvm-toolchain-precise
           packages:
             - clang-6
             - clang-format-6
```

The following is an example build where it happened: https://travis-ci.org/MartinNowack/klee/builds/410425177

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -185,7 +185,7 @@ module Travis
               end
 
               sh.export 'DEBIAN_FRONTEND', 'noninteractive', echo: true
-              sh.cmd "sudo -E apt-get -yq update &>> ~/apt-get-update.log", echo: true, timing: true
+              sh.cmd "sudo -E apt-get -yq update >> ~/apt-get-update.log", echo: true, timing: true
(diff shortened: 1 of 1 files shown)


joepvd:
Thanks! Not sure if that is what is going on here, though. `&>>` is a bash operator that was introduced with bash4:

> There is a new &>> redirection operator, which appends the standard output
>    and standard error to the named file.

For some reason, I could not get clang-6 installed, it seems to be [missing upstream](https://travis-ci.org/joepvd/travis-experiment/builds/410484430#L634-L671)?

If the clang-7 preview is acceptable for your purposes, this snippet might help:

```yml
addons:
    apt:
        sources:
             - ubuntu-toolchain-r-test
             - llvm-toolchain-trusty
        packages:
            - clang-7
            - clang-format-7
            - colordiff
```

MartinNowack:
@joepvd Woow, sorry - your absolutely right. Thanks a lot for your effort and time.

PR #1447, opened 2018-08-09
Disable `release` Julia version

simonbyrne:
cc: @travis-ci/julia-maintainers


lib/travis/build/script/julia.rb
```diff
@@ -11,10 +11,6 @@ module Travis
   module Build
     class Script
       class Julia < Script
-        DEFAULTS = {
-          julia: 'release',
```
tkelman: make this 1.0 ? I don't think all that many people do `language: julia` without any `julia:` specifiers, but may as well keep that possible?

lib/travis/build/script/julia.rb
```diff
@@ -108,8 +104,7 @@ def julia_url
             end
             case julia_version = Array(config[:julia]).first.to_s
             when 'release'
-              # CHANGEME on new minor releases (once or twice a year)
-              url = "julialang-s3.julialang.org/bin/#{osarch}/0.6/julia-0.6-latest-#{ext}"
+              sh.failure "'release' is no longer a valid julia version specifier in .travis.yml"
             when 'nightly'
               url = "julialangnightlies-s3.julialang.org/bin/#{osarch}/julia-latest-#{nightlyext}"
             when /^(\d+\.\d+)\.\d+$/
```
tkelman: oh I was wrong you'd need to also add a
```
when /^(\d+)$/
  url = "julialang-s3.julialang.org/bin/#{osarch}/#{$1}/julia-#{$1}-latest-#{ext}"
```
to the switch-case below to make just specifying the major ver possible, in addition to needing the corresponding file to exist on s3

simonbyrne: I'll wait until it is properly supported.

We will probably want 1.0 nightlies at some point as well.

lib/travis/build/script/julia.rb
```diff
@@ -108,14 +108,17 @@ def julia_url
             end
             case julia_version = Array(config[:julia]).first.to_s
             when 'release'
-              # CHANGEME on new minor releases (once or twice a year)
-              url = "julialang-s3.julialang.org/bin/#{osarch}/0.6/julia-0.6-latest-#{ext}"
+              sh.failure "'release' is no longer a valid julia version specifier in .travis.yml," \
+                "please use '1' instead."
```
tkelman: Duh, so just listing a major is going to have the exact same problem. If you test on "1" but say you support 1.0, you won't notice if you start depending on 1.1-only features.

simonbyrne: This is to solve the opposite problem: ensure that seldom-updated packages aren't just being tested on a 3 year old release. We should also encourage package authors to explicitly test their minimum supported version as well (probably would be good to write somewhere).

tkelman: That had been mentioned in the attobot tagging message, wasn't it? I don't think continuing to test against a constant release is a problem at all, it's beneficial and a better default than silently changing what's running. If a package hasn't been touched, travis doesn't run nightly (unless you've asked it to). That's a different issue and not really for a package's own CI to be responsible for when what's changing between test runs is the ecosystem around it, rather than the package itself.

simonbyrne: So what are you saying should be changed in this PR?

simonbyrne: Once this is merged, my plan is that the auto-generated .travis.yml will specify to run against versions 1 and the maj.min version used to generate the package. That way we should capture both. (I may also get a script to go through the registry and do this for existing packages as well).

tkelman:
That would be redundant for a lot of packages. I don't think "1" should be encouraged as it'll either be redundant or insufficient given the way most people will use it.
________________________________

simonbyrne:
Objection is noted, but I disagree.

I think this is ready to merge.

tkelman:
Disagree with which part? That it would be redundant or that it would be insufficient?

simonbyrne:
That it is redundant.

tkelman:
If someone lists
```
julia:
  - 1
  - 1.0
```
then you're wasting CI resources and duplicating jobs until a 1.1.0 exists.

simonbyrne:
Technically, yes. But they should be using `0.7` as the minimum version, not `1.0`. In either case, I don't think the extra few minutes of CI time is terribly onerous (and package authors are welcome to disable it if they don't like it). I just don't see you rationale for not allowing it at all.

tkelman:
It's less useful than its trivial replacement (adding the minor that you actually want to test against) in all cases. If packages aren't getting linted that min supported version is actually getting tested against, then people will tend to remove the minor in place of the major-only version since it won't need to be adjusted as often. But that is guaranteed to get out of sync and stop testing the actual min supported version as long as the pieces of information aren't derived or enforced to be equal.

simonbyrne:
We seem to be going in circles here.

tkelman:
Allowing people to do only
```
julia:
  - 1
```
or
```
julia:
  - 1
  - nightly
```
are both actively harmful and incorrect and don't serve any unique purpose that adding the specific minimum trailing minor digit can accomplish.

joepvd:
I'd be curious to what the rationale is for dropping support for the `release` version-alias.

Think I'd err on the side of not breaking backwards compatibility, but happy to get convinced otherwise :)

tkelman:
See discussion at https://github.com/travis-ci/docs-travis-ci-com/pull/1970#discussion_r209072844. It's not in active use in any non-abandoned packages as far as I'm aware. It drops testing over time of versions that shouldn't be dropped, and gets out of sync with the actual package manager's source of information about what's supported. As does the newly proposed `1` here, they're functionally equivalent, which is why I'm opposed to supporting a new name for the same discouraged and never-used legacy holdover.

joepvd:
Thanks for the pointers and elaborations! Your statement about the use of `release` is supported by our data about Julia versions specified per job that ran in the last 30 days on .org:

| name | percent |
|-----------|------------|
| nightly | 35.0232    |
| 0.6       | 31.7612    |
| 0.7       | 25.7991    |
| 1.0       | 5.91258    |
| 0.5       | 1.08419    |
| 0.4       | 0.13459    |
| release | 0.112158   |
| 0.6.0   |  0.0448632  |
| 0.6.3   | 0.037386   |
| [null]    | 0.0299088  |
| 0.6.4   | 0.00747719 |

KristofferC:
Having a `1` to mean the last stable Julia release would be useful. What needs to be done for this to get merged?

timholy:
Yes, we need a way to say "latest stable release." Many repos that I collaborate on plan to test 3 julia versions: 1.0 (LTS), latest released 1.x, and nightly. I don't care how this is spelled, but let's pick something and do it.

simonbyrne:
@stefankarpinski @ararslan @staticfloat are the Travis Julia maintainers, so I guess it's up to them?

staticfloat:
IMO `1` is useful and clear; we should support that.

PR #1448, opened 2018-08-10
Switch from xctool to xcodebuild+xcpretty for modern images

mjm:
This changes images from xcode8 and onward to use xcodebuild piped into xcpretty as the default script instead of xctool.

To make this actually usable, I also added support for a new `xcode_destination` key in the job config, which becomes the `-destination` argument. xcodebuild needs this in most cases when running tests.

We decided to keep older images using xctool, so the `xcode6.4` and `xcode7.3` images are explicitly checked for and fallback to xctool.

This has been tested in staging and works as expected in both cases.

### DIFF
--- a/lib/travis/build/script/objective_c.rb
+++ b/lib/travis/build/script/objective_c.rb
@@ -90,7 +90,11 @@ module Travis
 
           sh.else do
             if config[:xcode_scheme] && (config[:xcode_project] || config[:xcode_workspace])
-              sh.cmd "set -o pipefail && xcodebuild #{xcodebuild_args} build test | xcpretty"
+              if use_xctool?
+                sh.cmd "xctool #{xcodebuild_args} build test"
+              else
+                sh.cmd "set -o pipefail && xcodebuild #{xcodebuild_args} build test | xcpretty"
+              end
             else
               # deprecate DEPRECATED_MISSING_WORKSPACE_OR_PROJECT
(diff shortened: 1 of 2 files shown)


joshk:
First off, and most importantly, awesome work!

The one thing I am torn about is adding `xcode_destination` vs `xcode.destination` (or in yaml form)
```
xcode:
  destination:
```
but making this change for this one thing would be silly as none of the other configs use this newer syntax. Soooooo, I will write this up as a different change for later.

But again I state, awesome work, lets 🚢 it!

mjm:
Thanks Josh!

I'm going to write up another issue for the YAML thing, as I've mentioned internally some things I'd like us to do with a `xcode:` section in the YAML. I think for now keeping this consistent with what's already there is good, and we could improve this for everything at once as a separate piece of work, like you say!

I'd love to ship it now, but I think we're going to need documentation changes and to communicate this change to users before we do that.

cotsog:
Oh, just had a thought, how about we fold the `-destination` parameters under `xcode_destination:` e.g.

```yaml
xcode_destination:
  platform: iOS Simulator
  os: 11.3
  name: iPhone X
```

mjm:
@cotsog I think that's a great suggestion! I think rather than doing it here, though, I'd prefer to leave this as is (keep the change small and simple) and then rework all of the xcode YAML stuff at the same time, as we're discussing in https://github.com/travis-ci/product/issues/45

cotsog:
@mjm SGTM to do it at a later time. 👍

PR #1450, opened 2018-08-14
Add coverage upload support [julia]

simonbyrne:
This adds coveralls and codecov.io support to the Julia travis script.

Fixes https://github.com/travis-ci/travis-ci/issues/9982


simonbyrne:
cc: @travis-ci/julia-maintainers

lib/travis/build/script/julia.rb
```diff
@@ -90,6 +100,13 @@ def script
             sh.cmd 'julia --color=yes -e "VERSION >= v\"0.7.0-DEV.5183\" && using Pkg; Pkg.clone(pwd()); Pkg.build(\"${JL_PKG}\")"'
             # run tests
             sh.cmd 'julia --check-bounds=yes --color=yes -e "VERSION >= v\"0.7.0-DEV.5183\" && using Pkg; Pkg.test(\"${JL_PKG}\", coverage=true)"'
+            # coverage
+            if config[:codecov]
+              sh.cmd 'julia --color=yes -e "VERSION >= v\"0.7.0-DEV.5183\" && using Pkg; cd(Pkg.dir(\"${JL_PKG}\")); Pkg.add(\`"Coverage\`"); using Coverage; Codecov.submit(process_folder())"'
```
ararslan: You still have some stray backticks here and below

simonbyrne:
Any chance this could get merged into staging?

simonbyrne:
Bump?

BanzaiMan:
Hello. This has been deployed to https://staging.travis-ci.org.

simonbyrne:
Looks like it works on our test repo, I think this is good to go.

simonbyrne:
Can this be merged?

simonbyrne:
Thanks!

NHDaly:
Thanks Simon! :D I'd actually forgotten about this. :)

Can this change come with a documentation change somewhere?

I can send a PR to the Example.jl repo, but I'm not sure where the docs are that guide people on using Travis?

Thanks again!

simonbyrne:
@NHDaly There is an open pull request for the Travis documentation: https://github.com/travis-ci/docs-travis-ci-com/pull/2019

A PR to Example.jl would be great, thanks! (also, please try it out and let me know if there are any problems).

NHDaly:
Awesome, thanks Simon!

I've opened one for Example, here: https://github.com/JuliaLang/Example.jl/pull/33

ararslan:
This seemed to mask failures for us in https://github.com/invenia/CloudWatchLogs.jl/pull/6; the tests failed but the coverage step succeeded, so the build reported as having passed.

simonbyrne:
@ararslan Can you link to a commit & build log where it was a problem?

ararslan:
https://travis-ci.com/invenia/CloudWatchLogs.jl/jobs/147902828

simonbyrne:
@BanzaiMan do you know why this is happening? How do we get Travis to fail if any of the script `sh.cmd`s fail?

ararslan:
Would it not be possible to insert this into `after_success` rather than into `script`?

BanzaiMan:
`after_success` does not alter the build result.

To mark a command a failure with `sh.cmd`, pass `assert: true` as a part of the last arg option.

ararslan:
> `after_success` does not alter the build result.

Right. In this case, we should only submit coverage to the specified provider(s) if the build succeeded.

simonbyrne:
the problem is then that users defining their own `after_success` blocks would overwrite the coverage ones.

ararslan:
Hm. Well, if we add a `documenter: true` item as well, that would pretty much remove the need for custom `after_success` blocks in packages, I would think, in which case it should be safe to put all of this in a default `after_success`.

BanzaiMan:
I believe you can use `after_after_success` to execute after `after_success`. This is when the deployment code runs, and I think those will be run. (We'll have to test that, obviously.)

simonbyrne:
There seems to be a convention that default build scripts don't define an `after_success`.

simonbyrne:
@BanzaiMan So what determines a failure currently? Is it the return value of the last `sh.cmd`?

BanzaiMan:
@simonbyrne All of the commands that are set to `assert` do. If any of them returns non-zero, it will cause the job to be marked a failure (or an error, depending on when it happens).

BanzaiMan:
The default behavior for `sh.cmd` in the `script` section is `assert: true`.

simonbyrne:
Then why isn't it failing correctly here? The changes I made were in the `script` section.

BanzaiMan:
Are you sure that the command is exiting with non-zero?

simonbyrne:
Yes, because the test failures trigger travis failures correctly when the coverage options are disabled.

Is it because they're inside a `sh.if` block?

BanzaiMan:
Ah, sorry. I misspoke. The default for commands in `script` is *_not_* `assert`; we run all commands in `script`, no matter what the outcome.

simonbyrne:
So what determines whether the `script` block fails? Is it the result of the last output?

BanzaiMan:
@simonbyrne I've revisited the code again, and now I see the problem. Sorry about the confusion here.

1. When the `assert` option is `true`, the non-zero exit immediately halts execution and fails the build.
1. When it is not, the execution continues.
    1. When `sh.cmd` in a job phase (`script`, `install`, etc.) is used multiple times, the current behavior is such that we only check the last one (not inside an `sh.if` or some limited scope).

The last bit seems wrong to me. And this could be a problem with many languages. I'll have to check more thoroughly.

For the present problem for Julia, use of `after_after_success` might be best?

ararslan:
> For the present problem for Julia, use of `after_after_success` might be best?

+1 to that, that seems like a good solution to me.

simonbyrne:
> For the present problem for Julia, use of `after_after_success` might be best?

Will that conflict with deployment features?

BanzaiMan:
@simonbyrne I don't _think_ so, but it'll have to be tested.

NHDaly:
(when this is resolved, would you please ping https://github.com/JuliaLang/Example.jl/pull/33? thanks!)

PR #1454, opened 2018-08-17
Add deprecation warning for Xcode 6.4 image

mjm:
Please let me know what you think of this warning, especially the wording here. Also curious if this appliance is being applied in the right place for this sort of thing.

Here's a job that shows it in action on staging: https://staging.travis-ci.com/travis-infrastructure/Alamofire/jobs/219074


lib/travis/build/appliances/deprecate_xcode_64.rb
```diff
@@ -0,0 +1,16 @@
+require 'travis/build/appliances/base'
+
+module Travis
+  module Build
+    module Appliances
+      class DeprecateXcode64 < Base
+        def apply
+          if config[:osx_image] == 'xcode6.4'
+            sh.echo "Running builds with Xcode 6.4 in Travis CI is deprecated and will be removed in January 2019.", ansi: :yellow
```
joshk: could you add a new line above the first line, and below the last line, to help it stand out more?

mjm: Definitely!

lib/travis/build/appliances/deprecate_xcode_64.rb
```diff
@@ -0,0 +1,17 @@
...
+
+module Travis
+  module Build
+    module Appliances
+      class DeprecateXcode64 < Base
+        def apply
+          if config[:osx_image] == 'xcode6.4'
+            sh.echo ""
+            sh.echo "Running builds with Xcode 6.4 in Travis CI is deprecated and will be removed in January 2019.", ansi: :yellow
+            sh.echo "If Xcode 6.4 is critical to your builds, please contact our support team at support@travis-ci.com to discuss options.", ansi: :yellow
```
joshk: should we also put a newline after this line too?

mjm: Seems like there's already a blank line here in the log: https://staging.travis-ci.com/travis-infrastructure/Alamofire/jobs/219074

joshk:
true true, I should have checked. 

Ship It!

cotsog:
Hmmm, just looked at the rendering in the build log and I'm wondering if this should rather be a banner in travis-web instead?

cotsog:
Example banner for precise: https://travis-ci.org/cotsog/travis_ci_prod_regular/builds/417386088

I think it would standout better.

WDYT?

mjm:
Hmm...that would be nicer, for sure. I'm not sure how that banner is produced.

mjm:
I'll dig in to travis-web and see if this is easy to do.

Halle:
To clarify, LGTM is about the build message work, but I'm not against changing to a web banner. I vaguely recall that Matt and I discussed this as an option a while back (but possibly as a solution to the beta support message rather than this one?) but neither of us knew where it was done.

mjm:
I'm fine with both. Let's get this in there and I'll see about adding this to travis-web as well.

PR #1456, opened 2018-08-22
use directory cache for npm, too

zkat:
being able to do `cache: npm` would be pretty handy. idk what y'all have got going on with that network proxy, but the local disk cache on npm is pretty nice now and it would be nice if more people used it.


joshk:
Hi @zkat 

Are there some docs you can link to which provides recommendations on caching?

joshk:
should we also run `npm ci` if the npm cache is used?

zkat:
@joshk npm stores its network cache in `~/.npm`, and as of npm@5, having a warm cache is a massive speed boost. With `npm ci` and a present pkglock, this is even more so because it allows the cache to operate in pure content-addressable mode: that means a cache can be preserved across installs without affecting build reproducibility, since npm@5 and later do a full hash check before every extraction.

The best thing to do, imo, is for Travis to enable caching `~/.npm` by default, as it will just make things way faster (and cheaper?) for everyone involved. Is it possible we could do that or find other ways to maximize local cache usage for our users? That would be really amazing.

Historically, it seems like a common recommendation was to enable the cache for `node_modules`, but this can actually be slower in many cases, and is also entirely pointless with `npm ci` (because it blows away the `node_modules` before every run) -- it also exposes builds to having miscellaneous file corruption that is pretty much undetectable right now. As far as documents go, there's a [fairly good explainer in the `npm cache` docs](https://docs.npmjs.com/cli/cache#details), and I'm at your disposal if you have any questions or concerns (as the primary author of that section of our stack).

joshk:
@BanzaiMan and could you look into this one too?

BanzaiMan:
Thanks for pinging me. @zkat Do you mind opening a docs PR at https://github.com/travis-ci/docs-travis-ci-com/? Thanks!

BanzaiMan:
A duplicate of #1379.

PR #1460, opened 2018-08-28
update go version aliases to stable releases

jadr2ddude:
As of creating this commit, the stable releases of go are 1.11 and 1.10.4.

This updates the go 1.10.x alias to 1.10.4 and adds a 1.11.x alias.


meatballhat:
@jadr2ddude I'm sorry for leaving this file around and causing confusion!  This mapping is no longer used, as `gimme` now understands such aliases directly.

PR #1463, opened 2018-08-28
Use the full key id for the Matt Rutter PPA

jimhester:
The short ID is not secure, demonstrated by a collision with "Totally
Legit Signing Key <mallory@example.org>"

See https://keyserver.ubuntu.com/pks/lookup?search=0xE084DAB9&fingerprint=on&hash=on&op=vindex


joepvd:
https://github.com/travis-ci/travis-build/pull/1499

PR #947, opened 2017-01-31
Add `brew` addon

BanzaiMan:
For more declarative Homebrew package installation

PR #1467, opened 2018-08-29
Add an addon for installing Homebrew packages

mjm:
Linux has the APT addon for installing packages, but macOS users have to run `brew` commands manually to install packages. This change adds a new addon for Homebrew to set up taps and install packages and casks.

After I wrote this code, I did a search and found that @BanzaiMan had drafted a PR for this already: https://github.com/travis-ci/travis-build/pull/947. That's awesome: I'm glad there was already interest for this. I still want to open this one, though, as it's more full-featured.

A full example of what this looks like in config:

```
addons:
  homebrew:
    taps:           # pull in packages that aren't in homebrew/core
      - heroku/brew
      - homebrew/cask-versions
    casks: java8    # can specify a single string for any of these
    packages:
      - heroku      # this is pulled from the tap
      - imagemagick
      - python
    update: true    # optionally run 'brew update' before installing, which allows upgrading packages
```

Rather than run brew commands manually, this generates a Brewfile and then uses `brew bundle`, which handles figuring out if something needs to be installed or upgraded.


mjm:
I updated this to also support just using a Brewfile that already exists and running `brew bundle` on it:

```
addons:
  homebrew:
    brewfile: true   # Just run brew bundle in the repo. Can also use 'brewfile: MyBrewfile' for a custom path
    update: true
```

This can be used with the other keys as well, and it do both: bundle the generated Brewfile and the user one

lib/travis/build/addons/homebrew.rb
```diff
@@ -0,0 +1,97 @@
...
+        end
+
+        private
+
+        def config
+          @config ||= Hash(super)
+        end
+
+        def update_homebrew?
+          config[:update]
```
BanzaiMan: This simply tests truthiness, not `true`. I suggest normalizing and comparing. For example:

```ruby
config[:update].to_s.upcase == 'true'
```

lib/travis/build/addons/homebrew.rb
```diff
@@ -0,0 +1,97 @@
...
+          else
+            " --file=#{Shellwords.escape(config[:brewfile])}"
+          end
+        end
+
+        def install_homebrew_packages
+          sh.echo "Installing Homebrew Packages", ansi: :yellow
+
+          if user_brewfile?
+            sh.cmd "brew bundle#{bundle_arguments}", echo: true, timing: true
```
BanzaiMan: `bundle_arguments` seems a wrong name to me, since `bundle_*` sounds related to Bundler. Perhaps `brew_bundle_args`.

lib/travis/build/addons/homebrew.rb
```diff
@@ -0,0 +1,97 @@
...
+          end
+
+          if create_brewfile?
+            sh.file '~/.Brewfile', brewfile_contents
```
BanzaiMan: What happens if the user explicitly specifies `brewfile: ~/.Brewfile` (or equivalent)?

lib/travis/build/addons/homebrew.rb
```diff
@@ -0,0 +1,97 @@
...
+
+        def config_taps
+          @config_taps ||= Array(config[:taps]).flatten.compact
+        end
+
+        def create_brewfile?
+          !(config_taps.empty? && config_casks.empty? && config_packages.empty?)
+        end
+
+        def brewfile_contents
```
BanzaiMan: Does it make sense to cache this content for re-use? (There is currently only one place the value is used, so it is not a concern yet.)

lib/travis/build/addons/homebrew.rb
mjm: I think that would be a no-op, since I don't think their build will have an opportunity to put a file in that spot, and we will have already run ours. It would just run again and see nothing needed to happen.

lib/travis/build/addons/homebrew.rb
mjm: If we used it again, it would make sense, but I feel like that's not likely to happen. If we needed to use those contents in another way, we would probably be using them through the file.

joshk:
can we make `brewfile: true` a default? eg. if we detect a brewfile, we run `brew bundle`?

mjm:
Maybe, but I have some concerns depending on the answer to this question: do we run `brew bundle` when there is not an `addons.homebrew` section in the config file?

If yes, then my concern is that we are adding extra behavior users may not expect since they haven't opted in to it. You can tell me if we're okay with having this addon do stuff for a build even if they haven't referenced the addon at all in their config.

If no, then to get this default behavior, do you need an empty section for the addon in your config? Something like:

```
addons:
  homebrew:
```

It's weird but it's technically valid YAML.

I'm fine with implementing what you suggest in either of these ways depending on your feedback, @joshk.

joshk:
@mjm good point, lets have this be opt in for now, and we can make it a default later.

mjm:
Perfect! I’m thinking of landing this today/tomorrow then with an announcement when we release our next Xcode image

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

joshk:
@mjm love this! blog post? changelog annoucement?

mjm:
> @mjm love this! blog post? changelog annoucement?

@joshk Thanks! I’m planning to announce it in the same blog post announcing the Xcode 10 GM image that we expect to be coming soon

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

joshk:
@mjm awesome ❤️

stephengroat:
this is great, thanks for all the great work!

is there any way to use an existing `Brewfile` in the repo or does this just create a `Brewfile` from the `.travis.yml` and then run it?

adamnovak:
Is this actually deployed right now? [The docs](https://docs.travis-ci.com/user/installing-dependencies/#installing-packages-on-os-x) say I can use it, but [I tried it](https://github.com/vgteam/vg/blob/d6e0e95b8137835da838fff959a15e8fab91808b/.travis.yml#L64-L80) and [it failed when I tried to force-link Bison](https://travis-ci.org/vgteam/vg/jobs/430157755#L394), apparently saying that Bison was not installed. I see no evidence that any brew things were run; maybe [this weird error](https://travis-ci.org/vgteam/vg/jobs/430157755#L66) is where it was supposed to happen?

mjm:
It is deployed, but that error does seem like it could be related to why it's not running. We're looking into it!

slodki:
https://travis-ci.org/slodki/moneymanagerex/jobs/430258662#L70
Brewfile cannot ee found as `brew bundle` is run before `git clone`...
:-1:

mjm:
@slodki, you're right! I'll see about fixing that up, thanks for pointing that out.

slodki:
`brew` errors are ignored - the running job should be stopped on errors IMHO or separate toggle added for this.

mjm:
PR to move `brew bundle` to after the git clone: https://github.com/travis-ci/travis-build/pull/1491

mjm:
Fix for using `Brewfile` from repo is deployed now, so that should be working as expected.

slodki:
Thx.

**Next problem:**
https://travis-ci.org/slodki/moneymanagerex/jobs/430672853#L154
`No output has been received in the last 10m0s...`
This error can occur when the binary bottle is not available and brew must rebuild formulae from sources...

I used ` --verbose` to see compile details in such case.
You can also prefix `brew bundle` command with `travis_wait`
Could you fix this too?

mjm:
I see. I think we probably want to just use the `--verbose` flag when we run `brew bundle`, as the output is helpful and consistent with what people expect to see when they would run `brew install`. I'll make a PR.

PR #1471, opened 2018-08-30
Use full key ids for updating apt-keys

joepvd:


lib/travis/build/appliances/update_apt_keys.rb
```diff
@@ -5,12 +5,24 @@ module Build
...
+            apt-key adv --list-public-keys --with-fingerprint --with-colons \\
+              | awk -F: '
+                  $1=="pub" && $2=="e" { state="expired"}
+                  $1=="fpr" && state == "expired" {
+                    out = sprintf("%s %s", out, $(NF -1))
+                    state="valid"
+                  }
+                  END {
+                    if (length(out)>0)
+                      printf "sudo apt-key adv --recv-keys --keyserver keys.gnupg.net %s", out
```
meatballhat: I think we should be using ha.pool.sks-keyservers.net instead of keys.gnupg.net, assuming we can't switch to our own keyserver :grin:

lib/travis/build/appliances/update_apt_keys.rb
```diff
@@ -5,12 +5,24 @@ module Build
...
-          if command -v apt-get &>/dev/null && [[ -d /var/lib/apt/lists ]]; then
-            LANG=C apt-key list | awk -F'[ /]+' '/expired:/{printf "apt-key adv --recv-keys --keyserver keys.gnupg.net %s\\n", $3}' | sudo sh &>/dev/null
-          fi
-          EOF
-          sh.cmd command, echo: false
+          sh.if '"$TRAVIS_OS_NAME" == linux' do
+            command = <<~KEYUPDATE
+            apt-key adv --list-public-keys --with-fingerprint --with-colons \\
+              | awk -F: '
+                  $1=="pub" && $2=="e" { state="expired"}
```
meatballhat: add a space before final `}` &lt;/nitpick&gt;

lib/travis/build/appliances/update_apt_keys.rb
```diff
@@ -5,12 +5,24 @@ module Build
...
         def apply
-          command = <<-EOF
-          if command -v apt-get &>/dev/null && [[ -d /var/lib/apt/lists ]]; then
-            LANG=C apt-key list | awk -F'[ /]+' '/expired:/{printf "apt-key adv --recv-keys --keyserver keys.gnupg.net %s\\n", $3}' | sudo sh &>/dev/null
-          fi
-          EOF
-          sh.cmd command, echo: false
+          sh.if '"$TRAVIS_OS_NAME" == linux' do
+            command = <<~KEYUPDATE
+            apt-key adv --list-public-keys --with-fingerprint --with-colons \\
```
meatballhat: pipe at the end of this line and no newline escape, per shfmt/shellcheck :grin: &lt;/nitpick&gt;

lib/travis/build/appliances/update_apt_keys.rb
```diff
@@ -5,12 +5,24 @@ module Build
...
+                  $1=="fpr" && state == "expired" {
+                    out = sprintf("%s %s", out, $(NF -1))
+                    state="valid"
+                  }
+                  END {
+                    if (length(out)>0)
+                      printf "sudo apt-key adv --recv-keys --keyserver keys.gnupg.net %s", out
+                  }
+                ' \\
+              | sh &>/dev/null
```
meatballhat: Given that we're using `bash` most (every?) other places, can this be `bash` instead of `sh`?

lib/travis/build/appliances/update_apt_keys.rb
```diff
@@ -5,12 +5,24 @@ module Build
...
+                  $1=="pub" && $2=="e" { state="expired"}
+                  $1=="fpr" && state == "expired" {
+                    out = sprintf("%s %s", out, $(NF -1))
+                    state="valid"
+                  }
+                  END {
+                    if (length(out)>0)
+                      printf "sudo apt-key adv --recv-keys --keyserver keys.gnupg.net %s", out
+                  }
+                ' \\
```
meatballhat: Same here with pipe at end of line and no newline escape

meatballhat:
Given the swamp-like nature of wild SKS servers, I would also like to propose that instead of pulling from any keyserver at all that these are downloaded by fingerprint, e.g. `curl -sSL "https://%{app_host}/files/gpg/${fingerprint}.asc" | sudo apt-key add -`, which would also require tracking these keys ourselves (a good thing, imho)

PR #1472, opened 2018-08-31
support dotnet:latest for csharp build

hanabi1224:


lib/travis/build/script/shared/dotnet.rb
```diff
@@ -0,0 +1,59 @@
...
+        def get_dotnet_core_sdk_version_latest()
+            return extract_latest_sdk_version(get_dotnet_core_download_page_link_latest())
+        end
+
+        def get_dotnet_core_sdk_version_preview()
+            return extract_latest_sdk_version(get_dotnet_core_download_page_link_preview())
+        end
+
+        def get_dotnet_core_version_anchors()
+            dotnet_download_archive_page_html = Nokogiri::HTML(open("https://www.microsoft.com/net/download/archives"))
```
akoeplinger: There's an easier way to do this, you can get `https://dotnetcli.azureedge.net/dotnet/Sdk/Current/latest.version`, the second line indicates the latest version.

In general we should do this as part of the sh install commands in csharp.rb and not in ruby so it properly runs as part of the build.

Something like this:

```ruby
...
sh.cmd 'sudo apt-get update -qq', timing: true, assert: true
if config_dotnet == "latest"
  sh.cmd "latest_dotnet_version=$(curl -sSL https://dotnetcli.azureedge.net/dotnet/Sdk/Current/latest.version | tail -n1)" , timing: true, assert: true
  sh.cmd "sudo apt-get install -qq dotnet-sdk-$latest_dotnet_version", timing: true, assert: true
else
  sh.cmd "sudo apt-get install -qq dotnet-#{dotnet_package_prefix}-#{dotnet_package_version}", timing: true, assert: true
end
...
```

Additionally you'll also need to tweak a bunch of logic that looks at `config_dotnet` to make sure it accepts `latest` as a valid value.

hanabi1224: @akoeplinger  Regarding your second point, I agree if latest is valid alias from package provider, but obviously it's not in dotnet case,  I think I will be much easier to resolve latest to a valid alias before hand in such case, which takes a single path instead of hack linux and osx separately, does that make sense?

hanabi1224: @akoeplinger  that blob link is awesome, BTW , what is the link for preview.version?

akoeplinger: you can use the test suite (https://github.com/travis-ci/travis-build#running-test-suites) for trying your changes locally. it should be pretty easy to verify the correct commands get emitted in the install script.

re: second point: encapsulating everything in the install script makes it easier to verify what is going on, i.e. I'm not sure Travis wants to have the language scripts make arbitrary HTTP requests as part of the ruby app. I'm not entirely sure how everything fits together but afaik the way it works is that the .travis.yml is processed by the app and spits out a .sh script which is what actually gets run in the build VM/container. That's why we need to keep the separation.

re: preview link: it seems they're not very consistent in offering one, e.g. there is `https://dotnetcli.azureedge.net/dotnet/Sdk/release/2.0.0-preview2/latest.version` but not one for the latest 2.2 preview. I'd try opening an issue on https://github.com/dotnet/core-setup to see if it can be fixed.

PR #1476, opened 2018-09-04
Add shfmt & shellcheck integration

meatballhat:
This is a fairly disruptive change that touches a lot of files and changes the development workflow, so I'm expecting some discussion and debate :sweat_smile:   Opinions wanted!

**Update**: This patch now includes what was previously in #1477 
**Update**: :arrow_up: same, but #1480


meatballhat:
This is working well (enough) in production :tada:

lib/travis/build/script/julia.rb
```diff
@@ -51,7 +51,7 @@ def setup
             else
               sh.failure "Operating system not supported: #{config[:os]}"
             end
-            sh.cmd 'export PATH="${PATH}:${HOME}/julia/bin"'
+            sh.cmd 'export PATH="${PATH}:${TRAVIS_HOME}/julia/bin"'
```
andreasnoack: It doesn't look like `TRAVIS_HOME` is [an official environment variable](https://docs.travis-ci.com/user/environment-variables/) and it doesn't seem to be defined when I run the container locally.

PR #1477, opened 2018-09-06
Break out (most) bash sources

meatballhat:
and replace all ERB rendering with either variable definition or function params.  This PR is intentionally targeting the branch for #1476, which began the rework of bash things.


igorwwwwwwwwwwwwwwwwwwww:
It's quite a large patch, so I didn't review it in depth. The overall idea lgtm!

Only concern I would have is to make sure it is working with the recently added build tracing stuff, since that depends on functions being defined upfront, and I'm not 100% if this patch changes the order of things in the generated `build.sh`.

Build tracing is enabled for all jobs on staging, so if this works on staging then that's good enough for me! :)

PR #1478, opened 2018-09-06
Do not append and replace the --no-manual

jimhester:
Replacing in-place was causing the --no-manual to be repeatedly
inserted. Assigning the configuration variable explicitly should
hopefully fix the issue.

Fixes https://github.com/travis-ci/travis-ci/issues/10089


jeroen:
Great! This has bothered me for a while but I was never able to figure out the problem. Hope this fixes it.

lgtm @joepvd!

PR #1480, opened 2018-09-07
Support infrastructure-specific APT mirror substitution

meatballhat:
which also introduces infrastructure detection via a `travis_whereami` function.  Please note that this PR targets the branch for #1476.

### DIFF
--- a/lib/travis/build/bash/travis_munge_apt_sources.bash
+++ b/lib/travis/build/bash/travis_munge_apt_sources.bash
@@ -16,6 +16,7 @@ travis_munge_apt_sources() {
   fi
 
   if [[ ! "${mirror}" ]]; then
+    echo -e "${ANSI_YELLOW}No APT mirror found; skipping source munging.${ANSI_RESET}"
     return
   fi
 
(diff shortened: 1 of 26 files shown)


meatballhat:
Example in staging: https://staging.travis-ci.org/meatballhat/yolo-octo-adventure/builds/727400

lib/travis/build/bash/travis_munge_apt_sources.bash
```diff
@@ -0,0 +1,26 @@
...
+    return
+  fi
+
+  local mirror="${TRAVIS_APT_MIRRORS_BY_INFRASTRUCTURE[${TRAVIS_INFRA}]}"
+  if [[ ! "${mirror}" ]]; then
+    mirror="${TRAVIS_APT_MIRRORS_BY_INFRASTRUCTURE[unknown]}"
+  fi
+
+  if [[ ! "${mirror}" ]]; then
+    return
```
cotsog: Could we print some message if we ever fall into this case?

lib/travis/build/bash/travis_setup_env.bash
```diff
@@ -53,6 +55,12 @@ travis_setup_env() {
   TRAVIS_TMPDIR="$(mktemp -d 2>/dev/null || mktemp -d -t 'travis_tmp')"
   export TRAVIS_TMPDIR
 
+  TRAVIS_INFRA=unknown
+  if [[ ! "${TRAVIS_DISABLE_INFRA_DETECTION}" ]]; then
```
cotsog: Nit: could we reverse this to use `TRAVIS_ENABLE_INFRA_DETECTION` instead?

PR #1482, opened 2018-09-12
Use remote docker server

meatballhat:
This is <del>an experimental</del> a change that connects to a remote docker server during the run of `./scripts/assert-examples`, which means that the docker image does not need to be pulled every job.  This reduced the job time by ~17m, which is back in the range of where it was before the addition of integration checks via `./scripts/assert-examples`.  <del>The infrastructure to allow for this was manually provisioned, so it's not in a production-ready state</del>.  Without further changes, such infrastructure will cost ~$25USD/month.

**Update**: travis-ci/terraform-config#545 fills in automation of the remote docker host


meatballhat:
@BanzaiMan I'm expecting the docker host maintenance to be similar to that of the pull-through registries running in EC2, which required some initial setup + testing, but have been running (and re-created) without intervention for over a year.  That being said, I would pessimistically estimate weekly/monthly involvement of a few human hours for a fleet of docker server hosts.

acnagy:
@meatballhat Oohhhhh this is cool! Might I ask it's portability to Enterprise (🔔 @danishkhan)? That said, it looks to me like it should be easy, assuming we set all those env vars correctly, which is totally ~doable~ documentable. I can think of a few folks enterprise and not-enterprise who might be into this :)

meatballhat:
@acnagy I can't think of a reason offhand why this wouldn't be possible on Enterprise :grin:   This implementation is fairly manual, though.  If deemed appropriate for productization, I would expect most/all of the implementation and configuration to be transparent.

PR #1484, opened 2018-09-14
Hide "Disabling Gradle daemon" from log output

igorwwwwwwwwwwwwwwwwwwww:
As far as I can tell, we display that line on almost every job. Yet it is not really that useful imo. This change would clean up the job output a bit.

FWIW, if there are good reasons to keep it, I'm totally fine with that too!


BanzaiMan:
The issue that prompted this was https://github.com/travis-ci/travis-ci/issues/8205. I don't think there is a good reason to display this.

joshk:
🎉

PR #1490, opened 2018-09-18
Don't use bash associative arrays

meatballhat:
since we still have bash 3 in the wild :sweat_smile:

### DIFF
--- a/lib/travis/build/appliances/apt_get_update.rb
+++ b/lib/travis/build/appliances/apt_get_update.rb
@@ -36,9 +36,11 @@ module Travis
           end
 
           def use_mirror
-            define_mirrors_by_infrastructure
-            sh.raw bash('travis_munge_apt_sources')
-            sh.cmd 'travis_munge_apt_sources'
+            sh.if '${TRAVIS_OS_NAME} == linux' do
+              define_mirrors_by_infrastructure
+              sh.raw bash('travis_munge_apt_sources')
+              sh.cmd 'travis_munge_apt_sources'
+            end
           end
 
           def define_mirrors_by_infrastructure
--- a/lib/travis/build/appliances/apt_get_update.rb
+++ b/lib/travis/build/appliances/apt_get_update.rb
@@ -42,9 +42,9 @@ module Travis
           end
 
           def define_mirrors_by_infrastructure
-            sh.raw 'declare -A TRAVIS_APT_MIRRORS_BY_INFRASTRUCTURE'
+            sh.raw 'declare -a _TRAVIS_APT_MIRRORS_BY_INFRASTRUCTURE'
             mirrors.each do |infra, url|
-              sh.raw %{TRAVIS_APT_MIRRORS_BY_INFRASTRUCTURE[#{infra}]="#{url}"}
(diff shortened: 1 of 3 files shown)


mjm:
Seems good to me!

However: why do we run this code on macOS anyway? We don't have APT there. You could get away with using Bash 4 stuff if you have it in all the Linux environments if this code wasn't emitted when platform = osx.

meatballhat:
@mjm I agree it's a bit wonky that this is being evaluated on all platforms.  I'll additionally wrap it in a platform check :+1:

lib/travis/build/appliances/apt_get_update.rb
```diff
@@ -36,9 +36,11 @@ def used?
           end
 
           def use_mirror
-            define_mirrors_by_infrastructure
-            sh.raw bash('travis_munge_apt_sources')
-            sh.cmd 'travis_munge_apt_sources'
+            sh.if '${TRAVIS_OS_NAME} == linux' do
```
mjm: I had actually anticipated this being a `config[:os]` check, where this code wouldn't even be in the script for macOS builds. Since it's interpreted, it probably doesn't matter too much.

meatballhat: I would like to get away from depending on config-level values in general :grin:  I realize this makes for a larger script, but I don't like trusting the job config for runtime values.

mjm: Gotcha.

I just noticed though that `== linux` may be too restrictive. I've seen references in travis-build to another value here, something like `linux-ppc64`? If this should run for that OS value as well, then maybe this should be `!= osx`.

meatballhat: Gads!  How about the latest commit, which should address `linux-ppc64`?

mjm: I'm not sure, but I was thinking about the values `config[:os]` has (which I thought would match TRAVIS_OS_NAME), not `uname` values. This is where I saw the reference to it: https://github.com/travis-ci/travis-build/blob/4f580b238530108cdd08719c326cd571d4e7b99f/spec/build/addons/apt_spec.rb#L38

I don't know what this is in the `uname` value, but I feel like changing how TRAVIS_OS_NAME is defined is a more far-reaching change than I know how to access here.

It looks like TRAVIS_OS_NAME can only have those two values, so this may have been fine how it was. The relationship between uname, config[:os], and TRAVIS_OS_NAME is currently making my head spin.

meatballhat: If it helps, `TRAVIS_OS_NAME` is a fairly recent addition, and (IIUC) intended to replace the inconsistent uses of `config[:os]` and `$(uname ...)` variations.

mjm: O rly? I had assumed it had been around awhile, and was there to let people with build matrices in both OSes conditionalize their setup.

meatballhat: Er, sorry, I mean that having `TRAVIS_OS_NAME` defined at runtime in this way is fairly recent.

mjm: Oh that makes more sense. Alright, I'm just gonna trust you on this because I don't know what's actually going on here! 😄

PR #1491, opened 2018-09-18
Run Homebrew addon in before_install

mjm:
The custom Brewfile support needs the sources to be checked out in order to be useful.


mjm:
Tested this in staging and it runs in the before_install step as expected and it can find the Brewfile in the repo.

PR #1494, opened 2018-09-19
Show verbose output for brew bundle

mjm:
Fix for https://github.com/travis-ci/travis-build/pull/1467#issuecomment-422913083.

By default, `brew bundle` doesn't show the normal Homebrew output for installing a package, but that means that long compiles of a single package can cause us to timeout the build.

I think the verbose output here is desirable anyway, since it gives more visibility into what the build is doing, and it's consistent with what a `brew install` command would show you.


stephengroat:
this can have some unintended consequences, i.e. a build process that is long enough that it overflows logs. just fyi

mjm:
That's a good point! This doesn't produce output that is any more verbose than an unadorned `brew install` command, so I'm leaning towards going forward with this. If folks report that they are having issues with log size due to this, though, we could add a flag to the config to control this, but I'd prefer to keep it simple until that's shown to be necessary.

stephengroat:
@mjm understood. from experience, @slodki will have the issue on `slodki/moneymanagerex` has had the issue with `wxmac` in the past

mjm:
Hmm...that may be a case to run your own `brew bundle` command in `before_install`, possibly using `travis_wait` if needed. I'm not sure if the addon can support all of these cases (timeouts, long log output) correctly without being cumbersome to work with.

PR #1498, opened 2018-09-21
Set CFLAGS='-w', CCFLAGS='-w'

BanzaiMan:
To suppress compiler warnings.

This PR introduces a new environment variable type, `Defaults`, that are not strictly Travis built-ins, but are common environment variables that users may define and subsequently override with their configurations (be in Settings panel, or in `.travis.yml`).

### DIFF
--- a/lib/travis/build/env.rb
+++ b/lib/travis/build/env.rb
@@ -1,4 +1,5 @@
 require 'travis/build/env/builtin'
+require 'travis/build/env/defaults'
 require 'travis/build/env/config'
 require 'travis/build/env/settings'
 require 'travis/build/env/var'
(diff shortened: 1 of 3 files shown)


BanzaiMan:
Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/727771 (See, in particular, https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/727771#L249-L250)

This shows how `.travis.yml` overrides the default value: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/727775#L26

cotsog:
Hmmm could we declare this environment variable only for `nvm` ? Maybe it would be less disruptive this way?

BanzaiMan:
@cotsog The wider change was suggested by @joshk.

PR #1514, opened 2018-10-02
Suppress message during Heroku CLI update

BanzaiMan:
It is a bit noisy; alert only when something
goes wrong.


BanzaiMan:
Tested https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/728897#L467-L468

PR #1519, opened 2018-10-05
Update Gemfile dependencies

meatballhat:
and pin `jwt` to `~> 1.5` given specificity of spec expectations.

This PR is related to #1517 but I'm trying to isolate the source of the spec failures which I can't reproduce locally...


meatballhat:
Found it!  Further bisecting `travis-config` versions over in #1517

PR #1523, opened 2018-10-08
Improvments for Windows support

BanzaiMan:
This PR adds a couple of improvements for the upcoming Windows support.

1. It extracts Node.js version managers to separate classes, so that different ones can be used based on some criteria, in particular, the OS. For the present purposes, we use `nvs` for Windows, `nvm` for everything else (Linux and macOS).
1. Disables caching when it the job runs on Windows. (This needs a separate investigation and fix.)

### DIFF
--- a/lib/travis/build/script/node_js.rb
+++ b/lib/travis/build/script/node_js.rb
@@ -111,18 +111,6 @@ module Travis
           end
         end
 
-        def default_version
-          DEFAULT_VERSION
-        end
-
-        def yarn_required_node_version
-          YARN_REQUIRED_NODE_VERSION
-        end
-
-        def npm_ci_cmd_version
-          NPM_CI_CMD_VERSION
-        end
-
         def node_js_given_in_config?
           !!config[:node_js]
         end
@@ -210,7 +198,7 @@ module Travis
             sh.if "-f yarn.lock" do
               sh.if yarn_req_not_met do
                 sh.echo "Node.js version $(node --version) does not meet requirement for yarn." \
-                  " Please use Node.js #{yarn_required_node_version} or later.", ansi: :red
+                  " Please use Node.js #{YARN_REQUIRED_NODE_VERSION} or later.", ansi: :red
                 npm_install config[:npm_args]
               end
               sh.else do
@@ -236,7 +224,7 @@ module Travis
           end
 
           def yarn_req_not_met
-            "$(travis_vers2int $(echo `node --version` | tr -d 'v')) -lt $(travis_vers2int #{yarn_required_node_version})"
+            "$(travis_vers2int $(echo `node --version` | tr -d 'v')) -lt $(travis_vers2int #{YARN_REQUIRED_NODE_VERSION})"
           end
 
           def is_win?
--- a/lib/travis/build/script/node_js.rb
+++ b/lib/travis/build/script/node_js.rb
@@ -36,7 +36,7 @@ module Travis
 
         def announce
           super
-          if iojs_3_plus?
+          if iojs_3_plus? && !is_win?
             sh.cmd 'echo -e "#include <array>\nstd::array<int, 1> arr = {0}; int main() {return 0;}" > /tmp/foo-$$.cpp', echo: false
             sh.raw "if ! ($CXX -std=c++11 -o /dev/null /tmp/foo-$$.cpp >&/dev/null || g++ -std=c++11 -o /dev/null /tmp/foo-$$.cpp >&/dev/null); then"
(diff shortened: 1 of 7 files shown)


BanzaiMan:
Tested on staging on Trusty: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/729092

lib/travis/build/script/node_js.rb
```diff
@@ -263,6 +226,18 @@ def prepend_path(path)
           def yarn_req_not_met
             "$(travis_vers2int $(echo `node --version` | tr -d 'v')) -lt $(travis_vers2int #{YARN_REQUIRED_NODE_VERSION})"
           end
+
+          def is_win?
+            config[:os].downcase.strip == 'windows'
```
joepvd: Can we rely on `config[:os]` to be sure of the operating system where the script will run?

BanzaiMan: I'm open to suggestions.

lib/travis/build/script/node_js/manager/nvm.rb
```diff
@@ -0,0 +1,73 @@
...
+              sh.if '$? -ne 0' do
+                sh.echo "Failed to install #{ver}. Remote repository may not be reachable.", ansi: :red
+                sh.echo "Using locally available version #{ver}, if applicable."
+                sh.cmd "nvm use #{ver}", assert: false, timing: false
+                sh.if '$? -ne 0' do
+                  sh.echo "Unable to use #{ver}", ansi: :red
+                  sh.cmd "false", assert: true, echo: false, timing: false
+                end
+              end
+              sh.export 'TRAVIS_NODE_VERSION', ver, echo: false
```
joshk: shouldn't we export this in NVS too?

BanzaiMan: Yes.

BanzaiMan: Actually, I take this back. This is executed when the config does not have `node_js` value defined, and the repository relies on `.nvmrc` to supply the Node.js version. So this should happen only with NVM.

lib/travis/build/script/node_js.rb
meatballhat: @BanzaiMan Why does a runtime `${TRAVIS_OS_NAME}` check mean that nvm and nvs need to be present on all images? :confused:

lib/travis/build/script/node_js.rb
```diff
@@ -263,6 +231,19 @@ def prepend_path(path)
...
+          end
+
+          def setup_os
+            if is_win?
+              sh.echo "Using NVS for managing Node.js versions on Windows (BETA)", ansi: :yellow
+              sh.export 'NVS_HOME', '$ProgramData/nvs', echo: false
+              sh.cmd 'git clone --single-branch -b joshk-msys_nt https://github.com/joshk/nvs $NVS_HOME'
```
meatballhat: Can whatever changes are in the `joshk/nvs` fork be moved/integrated into a `travis-ci/nvs` fork, assuming they can't be accepted upstream yet?

BanzaiMan: https://github.com/jasongin/nvs/pull/106 has been merged, so we can change this to the upstream.

lib/travis/build/script/node_js/manager/base.rb
```diff
@@ -0,0 +1,41 @@
...
+
+          end
+
+          def install
+
+          end
+
+          def show_version
+
+          end
```
meatballhat: Stylistic nitpick: why are these empty methods left with gaps like this?

BanzaiMan: Artifacts of editor styling.

lib/travis/build/script/node_js/manager/nvs.rb
```diff
@@ -0,0 +1,40 @@
...
+
+          end
+
+          def setup
+
+          end
+
+          def update
+
+          end
```
meatballhat: More gappy empty methods? :confused: &lt;/nits&gt;

lib/travis/build/script/node_js.rb
BanzaiMan: @meatballhat It appears my thinking was wrong.

BanzaiMan: I think we should decide which version manager to use at the script compile time. Otherwise we need to check `${TRAVIS_OS_NAME}` value everywhere.

BanzaiMan:
We'll punt the `config[:os]` concern to another day.

PR #1527, opened 2018-10-10
Exit early when running on Windows + unsupported language

BanzaiMan:
This introduces env var `TRAVIS_WINDOWS_LANGS`, which is a comma-separated list of language names that are allowed to run on Windows.


BanzaiMan:
Tested: https://staging.travis-ci.com/BanzaiMan/com_testing/builds/232020

PR #1530, opened 2018-10-11
Use tar for copying files to go_import_path

BanzaiMan:
`rsync` appears to have problems on Windows.
job__7_2_-_tianon_gosleep_-_travis_ci


BanzaiMan:
Apparently ☝️ is not just Windows. So fun.

PR #1533, opened 2018-10-12
Java adjustments

joepvd:
Staging: https://staging.travis-ci.org/joepvd/travis-staging-test/builds/729471


lib/travis/build/bash/travis_remove_from_path.bash
```diff
@@ -0,0 +1,10 @@
+travis_remove_from_path() {
+  local target="$1"
+  echo "path: $PATH"
+  echo "java home: $JAVA_HOME"
```
meatballhat: Are these `echo`s intentionally left here?

PR #1539, opened 2018-10-14
Skip running `npm install` when yarn requirement is not met

ai:
I have a case of:

1. I have Node.js project with Yarn
2. I must test Node 10, 8, 6, 4, 0.12, 0.8.
3. Since most of the test tools don’t work on Node.js 0.12-0.8 I have custom [install script](https://github.com/ai/nanoid/blob/783ba0ca2ed7d45fb3dc3db8645a7306d4c49a93/test/ci-install) which install dependencies only for Node 10-6.

Unfortunately, right now this case [will not work with Travis](https://travis-ci.org/ai/nanoid/builds/441120983) since if Yarn doesn’t meet the requirement, Travis CI ignore custom `install` script and run `npm install`.

This PR should fix it and make this logic more predictable.

### DIFF
--- a/lib/travis/build/script/node_js.rb
+++ b/lib/travis/build/script/node_js.rb
@@ -64,7 +64,7 @@ module Travis
                 npm_install config[:npm_args]
               end
               sh.else do
-                sh.cmd "yarn", retry: true, fold: 'install'
+                npm_install config[:npm_args]
               end
             end
             sh.else do
@@ -204,7 +204,6 @@ module Travis
               sh.if yarn_req_not_met do
                 sh.echo "Node.js version $(node --version) does not meet requirement for yarn." \
                   " Please use Node.js #{YARN_REQUIRED_NODE_VERSION} or later.", ansi: :red
-                npm_install config[:npm_args]
               end
               sh.else do
                 sh.fold "install.yarn" do
--- a/lib/travis/build/script/node_js.rb
+++ b/lib/travis/build/script/node_js.rb
@@ -64,7 +64,7 @@ module Travis
                 npm_install config[:npm_args]
               end
               sh.else do
-                npm_install config[:npm_args]
+                sh.cmd "yarn", retry: true, fold: 'install'
               end
             end
             sh.else do


lib/travis/build/script/node_js.rb
```diff
@@ -204,7 +204,6 @@ def install_yarn
               sh.if yarn_req_not_met do
                 sh.echo "Node.js version $(node --version) does not meet requirement for yarn." \
                   " Please use Node.js #{YARN_REQUIRED_NODE_VERSION} or later.", ansi: :red
-                npm_install config[:npm_args]
```
BanzaiMan: This is a more salient question to ask. Is desirable to run `npm install` when the repository clearly indicates the use of `yarn` but the version of Node.js does not meet the requirement.

The current assumes "yes". I chose this behavior when `yarn` was relatively new to the scene and falling back to `npm` made sense.

A couple of years later, this assumption may no longer holds. However, dropping this could break existing builds that rely on this behavior.

BanzaiMan: Hmm. That's a good point. To think, this does seem incongruous. The config option seems heavy-handed here, though. 🤔

ai: Other suggestion. Maybe we can suggest config to disable `yarn`? And I will use it in `matrix` to disable it for old Node.js versions?

ai: I made brunch to show current problem with `yarn`: https://travis-ci.org/ai/nanoid/builds/441188572

BanzaiMan: Right. So, dropping this line seems like a good first step towards cleaning up the logic. Would you like to have this PR reopened, or would you like to open a new one?

ai: I think reopening will be the best option to keep our good conversation :).

Do you want me to update PR and remove using this line to `install`?

BanzaiMan: I was thinking just dropping this line.

ai:
PR was fixed (hope you will squash all commits)

BanzaiMan:
Tested: https://staging.travis-ci.com/BanzaiMan/nanoid/builds/233110
Compare with: https://travis-ci.com/BanzaiMan/nanoid/builds/87954675

PR #1543, opened 2018-10-16
Use `rvm` for deployment only when available

BanzaiMan:
Our Windows image does not have `rvm` installed, so the deployment
will fail. In such a case, we will try installing `dpl` with the one ruby we have.

### DIFF
--- a/lib/travis/build/addons/deploy/script.rb
+++ b/lib/travis/build/addons/deploy/script.rb
@@ -147,11 +147,21 @@ module Travis
               if edge_changed?(last_deploy, config)
                 cmd "gem uninstall -aIx dpl", echo: true
               end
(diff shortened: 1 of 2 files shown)


BanzaiMan:
Here's a test from yesterday: https://staging.travis-ci.com/BanzaiMan/com_testing/builds/233647

On Windows, I/O can happen out of order, and it is a little difficult to see that it is working here. https://staging.travis-ci.com/BanzaiMan/com_testing/jobs/233648#L77 shows the `script` deployment's output "`yes!`" (among other outputs from deployment) before deployment's folds happen.

Today, the deployment is failing because Sauce Labs is [having a problem](https://status.saucelabs.com/incidents/7hg8v9sbsn1l) and our deployment currently depends on it.

PR #1544, opened 2018-10-17
Set filter strategy to :redirect_io on Windows

BanzaiMan:
On Windows, PTY module is unavailable, so the filter will fail with the
:pty strategy.

This PR will set the filter strategy to `redirect_io`, which does not have this limitation.

### DIFF
--- a/lib/travis/build/appliances/setup_filter.rb
+++ b/lib/travis/build/appliances/setup_filter.rb
@@ -85,7 +85,11 @@ module Travis
           end
 
           def strategy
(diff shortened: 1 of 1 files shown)


BanzaiMan:
The `:redirect_Io` [goes further](https://staging.travis-ci.com/BanzaiMan/com_testing/jobs/233742) than `:pty` on Windows, but it still fails.

BanzaiMan:
Superseded by https://github.com/travis-ci/travis-build/pull/1719

PR #1545, opened 2018-10-18
Die early if deprecated JDK is requested

BanzaiMan:


BanzaiMan:
See https://travis-ci.community/t/help-with-oraclejdk10-build-on-linux/398

sormuras:
lgtm

sergeivisotsky:
Is it right that oraclejdk10 needs to be removed from the build file by now?

PR #1552, opened 2018-10-26
Accept brewfile: true passed as a string

mjm:
For now, we don't have a way to pass this as a boolean when using travis-yml (see https://github.com/travis-ci/travis-yml/issues/77), so we need to handle both true and "true".

### DIFF
--- a/lib/travis/build/addons/homebrew.rb
+++ b/lib/travis/build/addons/homebrew.rb
@@ -72,7 +72,7 @@ module Travis
         end
 
         def brew_bundle_args
-          if config[:brewfile] == true
+          if config[:brewfile].to_s.downcase == 'true'
             ''
           else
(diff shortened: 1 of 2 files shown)


lib/travis/build/addons/homebrew.rb
```diff
@@ -72,7 +72,7 @@ def user_brewfile?
         end
 
         def brew_bundle_args
-          if config[:brewfile] == true
+          if config[:brewfile].to_s.downcase == 'true'
             ''
           else
```
BanzaiMan: So, the unlikely scenario where user specifies the brewfile named `true` is ignored. Is it worth documenting?

mjm: I'd be surprised if someone actually did this in practice, so I don't think it's worth documenting.

PR #1556, opened 2018-10-29
Suppress "Network availability" message

BanzaiMan:
We should expect this to be the case, and not make a fuss when this
basic expectation is met.


spec/build/script/shared/bash_script.rb
```diff
@@ -121,11 +121,6 @@
 
       expect($?.exitstatus).to be_zero,
         "expected script exec in container #{cid}, got #{$?.exitstatus}"
-
-      log_minus_header = logdest.read.sub(
-        /.*Network availability confirmed\./m, ''
-      )
-      expect(log_minus_header).to match(/^Done\. Your build exited with 0\./)
```
joshk: don't you want to keep the expect, but remove the `log_minus_header` bit?

PR #1558, opened 2018-10-30
Skip specs for travis_internal_ruby()

BanzaiMan:
These specs depend on actual execution of `travis_internal_ruby`,
which may be influenced by factors outside of this code base,
which makes them very brittle.

These need to be re-framed.


BanzaiMan:
I tried fixing the specs by defining `travis_internal_ruby` in the specs to use configurable Regexp to determine the return value, but so far I have been unsuccessful.

PR #1559, opened 2018-10-31
Capture TEENY when looking for first 2.3+ Ruby

BanzaiMan:
Otherwise, we would try to install '2.3', which may be newer than
the one pre-installed on the image. This is kind of OK on Ruby jobs,
because we install whatever the alias 2.3 resolves to, but on non-Ruby
jobs, this causes the addon to fail entirely.


BanzaiMan:
TESTED https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/730688#L92

PR #1561, opened 2018-10-31
Install 2.3.5 when no Ruby 2.3+ is available

BanzaiMan:
`2.3` does not work. https://travis-ci.org/BanzaiMan/travis_production_test/builds/449001602#L63-L74

2.3.5 has been around, and is known to work.


BanzaiMan:
Low-priority, since this is present for only xcode6.4 and non-Ruby lang.

PR #1568, opened 2018-11-04
[julia] Assertions to avoid masked failures

simonbyrne:
Fixes the problems identified in https://github.com/travis-ci/travis-build/pull/1450#issuecomment-424808189

cc: @travis-ci/julia-maintainers @ararslan @NHDaly


NHDaly:
Thanks Simon!

simonbyrne:
@ararslan any chance you could try this out on the repo you were having the problem with?

ararslan:
I've opened a PR on the repo updating it to use this, but I'm not sure what the cause of the earlier failure was that got masked.

simonbyrne:
Well, you could checking that `@test 1==2` still causes a test failure?

ararslan:
https://github.com/JuliaLang/Example.jl/pull/37

simonbyrne:
Looks good then.

simonbyrne:
From @ararslan  
> My only gripe with this now is that it shows up as build errors (designated with ❗️) rather than build failures (designated with ❌).

https://travis-ci.org/JuliaLang/Example.jl/builds/451103142?utm_source=github_status&utm_medium=notification

@BanzaiMan: I'm guessing to fix that we would need to move it to `after_after_success`?

iamed2:
Yes that is a big annoyance. Also, will failures in coverage submission now cause the build to fail?

simonbyrne:
Yes, that is actually a very good argument for moving it.

ararslan:
Unless that's fixed soon, can we revert this in the meantime?

simonbyrne:
Well, the previous behaviour was worse IMHO

simonbyrne:
just don't use coverage for now

ararslan:
Does this not affect builds that don't have coverage submission at the top level? It seems like it would because of the `assert`s.

ararslan:
It does, see https://travis-ci.org/JuliaLang/Example.jl/builds/451118887?utm_source=github_status&utm_medium=notification

PR #1574, opened 2018-11-06
[community: r] clean up quotes

LiNk-NY:


jimhester:
Thanks, looks good to me, the three levels of escaping you have to work through (ruby, bash, R) is always a good time!

wlandau:
Come to think of it, I have had to click "Restart Build" more often these days due to network issues. Could that be related?

LiNk-NY:
The `r.rb` script does not do anything with network modules AFAICT...

wlandau:
Thanks for your work on repairing this. The quoting issue appears fixed, but it appears that Travis is no longer finding a CRAN mirror. For [this build](https://travis-ci.org/ropensci/drake/builds/452470577?utm_source=github_status&utm_medium=notification), I tried to set a mirror in [`.travis.yml`](https://github.com/ropensci/drake/blob/master/.travis.yml#L7), but this appears to have had no effect.

jimhester:
@wlandau I observed this as well, it stems from the order in which the export and configure blocks run. https://github.com/travis-ci/travis-build/pull/1580 should fix it.

wlandau:
Thanks, Jim!

PR #1575, opened 2018-11-07
Support Snap, similar to how we support Apt

joshk:


joshk:
@BanzaiMan I need your help with my final failing spec. I'm at a loss.

PR #1576, opened 2018-11-07
Replace `travis` CLI addon invocation with stand-alone script

BanzaiMan:


BanzaiMan:
@meatballhat My intention is to replace `script/compile` entirely. I didn't do it with this PR, because I thought that the diff would be too large to properly comprehend.

meatballhat:
@BanzaiMan Replacing `./script/compile` makes sense to me, but I don't agree with removing the ability to use the script to test local changes. I would prefer to see this replacement happen within the same PR and I promise I won't complain about the diff size :heart: I would also like the script location to be `./script/compile`, or even better imho would be `./bin/travis-build-compile`.

BanzaiMan:
@meatballhat Good points. I'll make some changes.

README.md
```diff
@@ -117,9 +53,65 @@ docker-compose run -e RACK_ENV=development -p 4000:4000 web
...
+it is executed (e.g., edit `/etc/resolv.conf`, install software).  Some require
+`sudo` privileges and they are not easily undone.
+
+It is highly recommended that you run this in a container or other virtualized
+environment.
+
+The generated script can be used in a container or virtualized environment that
+closely mimics Travis CI's build environment to aid you in debugging the build
+failures.  Instructions for running such a container are available
+[in the Travis CI docs](https://docs.travis-ci.com/user/common-build-problems/#running-a-container-based-docker-image-locally).
```
BanzaiMan: This reference is removed at this removed.

PR #1579, opened 2018-11-08
Uninstall oclint when running on Mac

BanzaiMan:
This package is interfering with other more important packages such
as gcc. So we remove it on the fly.


BanzaiMan:
Test: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/731393

BanzaiMan:
and https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/731398

BanzaiMan:
Xcode6.4 does not seem to work; I tend to think it is a general problem with that image.

mjm:
Looks like Homebrew is broken, it doesn't seem to update successfully. Homebrew doesn't officially support an OS that old anymore. I'm not sure what we can do there :(

BanzaiMan:
Right. The image doesn't have Ruby 2.3, which is required for the current Homebrew, but installing Ruby 2.3 itself will require dependencies to be installed via—you guessed it—Homebrew. We've announced deprecation of Xcode6.4 for January, 2019, but it might be beyond salvation now.

mjm:
I'm inclined to agree. I think it'll just have to remain in its current state, and anyone who is still able to use it can until January.

PR #1580, opened 2018-11-08
Export R_PROFILE during the configure step, rather than export

jimhester:
export is run after configure, and R_PROFILE needs to be set for
`setup_bioc`.

It looks to me like the export stage is always run after configure, which is why the repos were not being set https://github.com/travis-ci/travis-build/blob/4f580b238530108cdd08719c326cd571d4e7b99f/lib/travis/build/stages.rb#L15-L19

An example error this should fix is https://travis-ci.org/ropenscilabs/available/jobs/452418452#L913-L916

cc @LiNk-NY


LiNk-NY:
Thanks @BanzaiMan ! :tada:

PR #1583, opened 2018-11-09
Build with scala 2.12.7

raboof:
Which is the currently common version for Scala development


BanzaiMan:
Sorry for the delay. Is 2.12.7 still the preferred version? Or is there a new one?

raboof:
No problem! 2.12.8 has come out since.

PR #1584, opened 2018-11-09
Assert Rust's `version` commands

BanzaiMan:
If Rust's installation has succeeded but the tools are not properly,
the execution reaches user-defined phases, which would most likely
fail.

In order to prevent this, we assert that Rust's various `--version`
commands all pass, so that we can fail the job as early as possible.


BanzaiMan:
See https://travis-ci.community/t/macos-rust-beta-build-failures/803/3 for more information.

PR #1585, opened 2018-11-10
nix: also print channel version

Mic92:
Using the hash from lib.version allows to locally
reproduce the exact same nixpkgs version,
which makes reproducing build failures a lot easier.

### DIFF
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -65,6 +65,7 @@ module Travis
           sh.echo 'Please open any issues at https://github.com/travis-ci/travis-ci/issues/new and cc @domenkozar @garbas @matthewbauer @grahamc', ansi: :green
 
           sh.cmd "nix-env --version"
+          sh.cmd "nix-instantiate --eval -E 'with import <nixpkgs> {}; lib.version or lib.nixpkgsVersion'"
         end
 
         def script


Mic92:
cc @domenkozar @garbas @matthewbauer @grahamc

lib/travis/build/script/nix.rb
```diff
@@ -65,6 +65,7 @@ def announce
           sh.echo 'Please open any issues at https://github.com/travis-ci/travis-ci/issues/new and cc @domenkozar @garbas @matthewbauer @grahamc', ansi: :green
 
           sh.cmd "nix-env --version"
+          sh.cmd "nix-instantiate --eval -E 'with import <nixpkgs> {}; lib.version or lib.nixpkgsVersion'"
```
Mic92: Example output:

"19.03.git.a7f4a860d0c"

At some point `nixpkgsVersion` was renamed to `version` since someone may use an old channel we fallback to `nixpkgsVersion` for backwards compatibility

domenkozar:
I'm not 100% sure it's worth wasting a few seconds, can't someone just add that when they need to debug instead? :)

grahamc:
Seems totally worth the few seconds if it means saving minutes of effort per dev when they want to know this?

Mic92:
It is around 100ms on my system because I have a nixpkgs git checkout. 
But with channels it only needs to read a `.version` file, which is quite fast.
Speaking of speed: Is it possible for community supported languages to run on the container infrastructure? I thought about having nix packaged in a deb file.

BanzaiMan:
I'll assume this is good to go. I'll merge this in the next couple of days.

PR #1587, opened 2018-11-12
Improve the MVN settings fixes

joshk:


joshk:
https://www.mkyong.com/mac/sed-command-hits-undefined-label-error-on-mac-os-x/

fixes the following in Mac builds:
image

PR #1588, opened 2018-11-12
Mac log improvements

joshk:


joshk:
@mjm do you think we need the certificate appliance?

Looking at https://travis-ci.com/travis-infrastructure/chirp-com-production/jobs/158129670#L48-L49
image

I suspect the deleting of the cert isn't working, and thus adding the same cert back is also failing.

But when you look here: https://travis-ci.org/joepvd/travis-experiment/jobs/453927132#L13 (which is an older Mac image), it seems to work.

In this PR I pipe the output to `/dev/null` for now, to at least remove the noise from the log.

mjm:
This is definitely before my time. I do believe we need to do it for those older images. It looks like it is 7.3 and 6.4 that are affected. My understanding is that we fixed the problem there in newer images, but older ones needed a workaround in travis-build.

We could put an explicit check around it for which image we are running on and skip it for most of them. Or we could just suppress the output. I'm not particular about which it is.

joshk:
@mjm are you fine with me suppressing the output and not asserting the exit code?

mjm:
Sure, that works.

PR #1598, opened 2018-11-15
Add bash+ruby func for ipv4 DNS lookups

meatballhat:
so that we can get addresses for names without having a tool like `host` or `dig` installed.

### DIFF
--- a/lib/travis/build/bash/travis_getaddrinfo.bash
+++ b/lib/travis/build/bash/travis_getaddrinfo.bash
@@ -1,0 +1,13 @@
+travis_getaddrinfo() {
+  local nodename="${1}"
+
+  if [[ ! "${nodename}" ]]; then
+    return
+  fi
+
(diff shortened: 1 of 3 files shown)


lib/travis/build/bash/travis_getaddrinfo.bash
```diff
@@ -0,0 +1,14 @@
+travis_getaddrinfo() {
+  local nodename="${1}"
+
+  if [[ ! "${nodename}" ]]; then
+    return
+  fi
+
+  # shellcheck disable=SC1010
+  rvm "$(travis_internal_ruby)" --fuzzy do ruby -rsocket <<EORUBY
```
BanzaiMan: Do we need rvm here? Seems straight-up `ruby` should work.

meatballhat: I'm happy to use `ruby` alone if that's an option!

BanzaiMan: Let's do that.

PR #1615, opened 2018-12-07
Julia: Use verbose build in default scripts

fredrikekre:
With Julia 1.1 `Pkg.build` has an option for verbose builds, see https://github.com/JuliaLang/Pkg.jl/pull/809. This should be the default on Travis, since otherwise Travis builds for packages with long build times might get killed due to no output, see e.g. https://github.com/JuliaLang/Pkg.jl/issues/665.

cc @ararslan , @simonbyrne

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -80,7 +80,7 @@ module Travis
               sh.cmd 'git fetch --unshallow'
             end
             # build
-            sh.cmd 'julia --color=yes -e "if VERSION < v\"0.7.0-DEV.5183\"; Pkg.clone(pwd()); Pkg.build(\"${JL_PKG}\"); else using Pkg; Pkg.build(); end"', assert: true
+            sh.cmd 'julia --color=yes -e "if VERSION < v\"0.7.0-DEV.5183\"; Pkg.clone(pwd()); Pkg.build(\"${JL_PKG}\"); else using Pkg; if VERSION >= v\"1.1.0-rc1\"; Pkg.build(verbose=true); else Pkg.build(); end; end"', assert: true
             # run tests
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/julia.rb
```diff
@@ -90,14 +90,14 @@ def script
...
-            
+
           end
           sh.else do
             sh.if '-a .git/shallow' do
               sh.cmd 'git fetch --unshallow'
             end
             # build
-            sh.cmd 'julia --color=yes -e "VERSION >= v\"0.7.0-DEV.5183\" && using Pkg; Pkg.clone(pwd()); Pkg.build(\"${JL_PKG}\")"', assert: true
+            sh.cmd 'julia --color=yes -e "if VERSION >= v\"0.7.0-DEV.5183\"; using Pkg; Pkg.clone(pwd()); if VERSION >= v\"1.1.0\"; Pkg.build(\"${JL_PKG}\"; verbose=true); else Pkg.build(\"${JL_PKG}\"); end"', assert: true
```
ararslan: You have two `if`s here with only one `end`.

fredrikekre: Thanks, same error on L83!

BanzaiMan:
@ararslan @fredrikekre What's the 1.1.0 release date?

fredrikekre:
No fixed date, but the the release is beeing finalized ATM. I can comment here when the release is final.

fredrikekre:
1.1-rc1 is now live, and  `julia: 1.1` works on travis (right @ararslan ?) so this should be good to merge.

ararslan:
Yes, 1.1 should work on Travis. To set this active right now for 1.1 builds, you could change the `VERSION` restriction to `VERSION >= v"1.1.0-rc1"`, since this otherwise won't be active until the full release.

fredrikekre:
> To set this active right now for 1.1 builds, you could change the VERSION restriction to VERSION >= v"1.1.0-rc1", since this otherwise won't be active until the full release.

Good idea, updated.

fredrikekre:
bump @BanzaiMan

PR #1618, opened 2018-12-12
Embed the marutter PGP key

jimhester:
Querying the keyserver tends to fail sporadically, so instead we can
embed the key directly into the script. The csharp build script does
this as well for a few of their keys.

Fixes https://travis-ci.community/t/ubuntu-keyserver-is-still-failing-builds/648

The implementation is largely borrowed from the csharp script

https://github.com/travis-ci/travis-build/blob/abc8a7aee2c667f1ceaacfb6b8db068f00be1ebf/lib/travis/build/script/csharp.rb#L207-L221


jeroen:
Maybe it's easier to read this from a file on a (reliable) server than hardcoding the literal key?

PR #1620, opened 2018-12-13
Ensure /etc/hosts ends with newline

joepvd:
Querying /etc/hosts from a job does not contain a line containing localhost and 127.0.0.1.

See e.g. https://staging.travis-ci.org/joepvd/travis-staging-test/builds/736810#L153

Ensuring that `/etc/hosts` ends with a newline magically makes the last line not disappear:

https://staging.travis-ci.org/joepvd/travis-staging-test/jobs/736823#L158


BanzaiMan:
This is "neat", for some definition of "neat."

PR #1621, opened 2018-12-13
Export TRAVIS_DEBUG_MODE for debug sessions

iriberri:
This came from a user request and I found it a quite interesting one!

Having `TRAVIS_DEBUG_MODE` env var, or similar, that can be checked during build time could allow users to have a way to change the behavior of their scripts/tests to immediately be more verbose as soon as someone restarts the job in debug mode, as an example.

If we decide to go ahead with this, it'd require some small docs updates as well - I'll take care of these.


BanzaiMan:
Waiting for docs PR.

PR #1625, opened 2018-12-18
Update URL for docker image language mapping

meatballhat:


BanzaiMan:
Is there a way to discover this URL (so it won't break again)?

PR #1626, opened 2018-12-19
Enable IPv4 forwarding on precise, trusty and xenial

soulshake:
Temporary patch to address https://github.com/travis-ci/travis-ci/issues/9696.

To be reverted once precise and trusty are gone.


soulshake:
The test I added, in its current form, _should_ be failing but is not. :thinking:

PR #1628, opened 2018-12-30
Homebrew: discard `brew update` output

mhinz:
`brew update` adds 700+ lines to the log.

Make it possible to silence that output via following config:

```
addons:
  homebrew:
    update: silent
```

(If you're in favor of this change, I'll make a PR for the docs as well.)

### DIFF
--- a/lib/travis/build/addons/homebrew.rb
+++ b/lib/travis/build/addons/homebrew.rb
@@ -44,7 +44,7 @@ module Travis
 
         def update_homebrew
           sh.echo "Updating Homebrew", ansi: :yellow
-          sh.cmd "rvm $brew_ruby do brew update", echo: true, timing: true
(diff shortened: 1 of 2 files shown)


mjm:
I think I would be fine with swallowing the output always. It's hard to think of when it would be particularly useful. A more complex approach would be to capture it and print it if the command fails, but that may be more trouble than it's worth.

mhinz:
> The echo option is for re-printing the command (here, brew update …) to the logs. Not whether or not STDOUT is seen.

Oops!

If you're fine with simply discarding stdout, should I change the commit to do this?

```diff
-          sh.cmd "rvm $brew_ruby do brew update", echo: true, timing: true
+          sh.cmd "rvm $brew_ruby do brew update 1>/dev/null", echo: true, timing: true
```

BanzaiMan:
@mhinz Yes. Something like that would work. 👍

PR #1629, opened 2018-12-30
Disable pip progress bar

adamjstewart:
As discussed in https://github.com/pypa/pip/issues/6101, Travis's logs don't respect the terminal control characters that pip's download progress bar uses. This results in thousands of lines of cruft being added to the logs. In the extreme case of large downloads, like `torch`, this results in [job termination](https://travis-ci.community/t/large-pip-installs-exceed-maximum-log-length/1599). 

This PR disables the use of the progress bar on all pip downloads, at least when `language: python` is set.

@BanzaiMan @native-api

### DIFF
--- a/lib/travis/build/script/python.rb
+++ b/lib/travis/build/script/python.rb
@@ -35,6 +35,7 @@ module Travis
           sh.cmd 'python --version'
           sh.cmd 'pip --version'
           sh.export 'PIP_DISABLE_PIP_VERSION_CHECK', '1', echo: false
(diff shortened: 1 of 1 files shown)


adamjstewart:
I'm not too worried. A 10 minute download would have exceeded the maximum log length hundreds of times over anyway.

adamjstewart:
Thanks @BanzaiMan! When do these changes go live? Is it instantaneous or do you have a release schedule?

PR #1630, opened 2019-01-04
Update R release version to 3.5.2

jimhester:


BanzaiMan:
Is there a way to discover this and keep it up to date?

We can run certain tasks at travis-build deploy time and set this value.

jimhester:
We have an API setup (https://rversions.r-pkg.org/r-release), so it would be possible.

However the bigger problem is I have to manually kickoff the binary build after each release and if we auto-update when a build doesn't exist they will fail.

But I could automate that process pretty easily as well I guess, so sure it is possible.

BanzaiMan:
Is there an API endpoint for the binary package instead of the release?

PR #1631, opened 2019-01-04
Add nixpkgs configuration option

cprussin:
This PR adds an option to specify the version of nixpkgs to use with a nix build.

### DIFF
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -57,7 +57,7 @@ module Travis
               sh.cmd 'source /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
             end
 
-            unless config[:nixpkgs] == 'unstable'
+            unless config[:nixpkgs].to_s == 'unstable'
               sh.cmd "nix-channel --add https://nixos.org/channels/nixos-#{config[:nixpkgs]} nixpkgs"
               sh.cmd 'nix-channel --update'
             end
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -58,7 +58,7 @@ module Travis
             end
 
             config[:channels].map do |channel, url|
-              sh.cmd "nix-channel --add #{expand_channel_url(channel, url)} #{channel}"
+              sh.cmd "nix-channel --add #{url} #{channel}"
             end
 
             sh.cmd 'nix-channel --update'
@@ -77,25 +77,6 @@ module Travis
 
         def script
           sh.cmd 'nix-build'
-        end
-
-        private
-
-        def is_number?(str)
-          true if Float(str) rescue false
-        end
-
-        def expand_channel_url(channel, url)
-          if is_number? url
-            case channel.to_s
-            when "nixpkgs"
-              "https://nixos.org/channels/nixos-#{url}"
-            else
-              url
-            end
-          else
-            url
-          end
         end
       end
     end
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -58,7 +58,7 @@ module Travis
               sh.cmd 'source /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
             end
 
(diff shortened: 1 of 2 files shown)


cprussin:
FYI @domenkozar @garbas @matthewbauer @grahamc

lib/travis/build/script/nix.rb
```diff
@@ -9,7 +9,8 @@ module Build
     class Script
       class Nix < Script
         DEFAULTS = {
-          nix: '2.0.4'
+          nix: '2.0.4',
+          nixpkgs: 'unstable'
```
matthewbauer: Could we make this a map so that you could:

```
channels:
  nixpkgs: 'https://nixos.org/channels/nixos-unstable'
```

and it runs:

```
sh.cmd "nix-channel --add https://nixos.org/channels/nixpkgs-unstable nixpkgs"
```

?

cprussin: Sure thing.  What do you think about allowing the following as a shorthand:

```rb
channels:
  nixpkgs: '18.09'
```

Which would be equivalent to:

```rb
channels:
  nixpkgs: 'https://nixos.org/channels/nixos-18.09'
```

?

cprussin: I went ahead and added this change in along with the shorthand and a few more tests.  Let me know if you have any feedabck and I'm happy to add more changes.

grahamc:
-1 on the shorthand. Will reply further soon.

lib/travis/build/script/nix.rb
```diff
@@ -71,6 +78,25 @@ def announce
...
+
+        def is_number?(str)
+          true if Float(str) rescue false
+        end
+
+        def expand_channel_url(channel, url)
+          if is_number? url
+            case channel.to_s
+            when "nixpkgs"
+              "https://nixos.org/channels/nixos-#{url}"
```
grahamc: I think the channel shouldn't be "expanded", it doesn't need to be for the user to get a nice experience: Users can specify `channel:nixpkgs-unstable` or `channel:nixos-18.09-darwin` (this is new as of Nix 2.0) so they don't need to specify a full URL anyway. This can be added to the Nix on Travis documentation (https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/languages/nix.md).

Beyond this nit, this is great work.

cprussin: Very cool, I was not aware that you could do that.  That will significantly simplify this Pr.

I'll go ahead and put a PR into the docs-travis-ci-com repo after this one lands to reflect the change.

Thanks for the feedback!

lib/travis/build/script/nix.rb
cprussin: As per @grahamc's comment below, I've gone ahead and removed the shorthand, since it sounds like `nix-channel` has a built-in shorthand that we can use here anyways.

cprussin:
@grahamc @matthewbauer Just a friendly nudge to remind you guys about this, I'd love to merge it in but can't until I get reviews.  Thanks!

cprussin:
@grahamc @matthewbauer any chance on getting this reviewed nad merged in?  Thanks!

BanzaiMan:
How is this supposed to be used? It appears that it adds a new "channel" configuration parameter in the top level. Is that true?

cprussin:
@BanzaiMan that's correct, it would be used similar to the `nix` option.  The intent would be for the build matrix to run the cross product of nix versions and channels.

cprussin:
I'm going to close this PR as it just doesn't seem like there's any interest in moving forward on it.  If anyone does want to revisit and has interest reviewing and merging I'll be more than happy to re-open in the future.

mwillsey:
Are there plans to implement this feature? For me, it's critical to have CI not be on unstable, it just moves too fast.

It seems like this was approved, but not merged?

cprussin:
I never received approval (positive messages, but no actual approval) so I could not ever merge this in. Still happy to revisit if any mods can give me a review, but it sat open for some time and I pinged folks a few times and I couldn't manage to get a review, so I consider this effort DOA

PR #1632, opened 2019-01-07
[R] Add --no-manual to r_build_args when latex: false

yutannihilation:
`R CMD build` usually doesn't build PDF manuals. But, if the manual contains `\Sexprs{}` macro, it does, and it will end up an error when `latex: false` (e.g. https://github.com/tidyverse/dplyr/pull/4085#issuecomment-451810489).

This PR adds `--no-manual` to `r_build_args` as well to avoid the error. (But, sorry, I don't know how to test this...)

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -163,6 +163,7 @@ module Travis
                 setup_latex
               else
                 config[:r_check_args] << " --no-manual"
+                config[:r_build_args] << " --no-manual"
               end
 
               setup_bioc if needs_bioc?


BanzaiMan:
@jimhester Review, please!

jimhester:
I think it looks ok, but https://github.com/travis-ci/travis-build/pull/1478/files should probably be merged first, and I think this PR should also use `+` rather than `<<` to avoid the repeated `--no-manual` problem.

yutannihilation:
Thanks, I'll fix after it's merged.

jimhester:
LGTM

yutannihilation:
Thanks!

PR #1633, opened 2019-01-07
Hardcode RubyGems version (< 3.0.0) to update to

BanzaiMan:
RubyGems 3.0.0 now requires Ruby 2.3.0, breaking many on-demand
installations of Rubies older than that.
So if our Ruby is older than 2.3.0, we stop at the last 2.7.x.


BanzaiMan:
Tested here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/736899#L151

PR #1635, opened 2019-01-08
Update RVM for JRuby jobs

BanzaiMan:


BanzaiMan:
Tested https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/736916

sfgeorge:
@BanzaiMan I'm wondering whether perhaps we should now revert this PR.  Here are a few thoughts on why this might be a good idea, interested in your thoughts as well.
* The minimum version of RVM needed for the switch to new maven-based JRuby download URLs (1.29.5) seems to already be installed on current Travis xenial images (which currently have RVM 1.29.7 pre-installed).
* [As you already noted](https://travis-ci.community/t/rvm-tries-to-download-blank-version-of-itself/3972/5), the additional `rvm get ...` calls result in more unauthenticated GitHub API calls by RVM on Travis nodes;  since only 60 such requests per i.p. address and Travis only has a [finite number of public i.p. addresses](https://docs.travis-ci.com/user/ip-addresses/), these can easily get rate-limited, leading to the 403 Forbidden that many have noticed. 
* While getting the latest stable makes sense and is a good endeavor, it's a deviation from what is done for other rubies.  And pragmatically for now, avoiding it if not needed would reduce API hits.

Just starting a convo on this.  What are your thoughts?

BanzaiMan:
I'll review the community thread and respond there. This is not a place for further discussion.

PR #1636, opened 2019-01-10
Add support for golang 1.11+ modules

meatballhat:


lib/travis/build/bash/travis_export_go.bash
```diff
@@ -0,0 +1,10 @@
+travis_export_go() {
+  if [[ ! "${1}" ]]; then
+    echo 'Missing go version positional argument' >&2
+    return 86
+  fi
+  export TRAVIS_GO_VERSION="${1}"
+  export GIMME_GO_VERSION="${TRAVIS_GO_VERSION}"
+  : "${GOMAXPROCS:=2}"
```
joepvd: Would it be worth it to default to `nproc` rather than a hardcoded 2, at least on `$TRAVIS_OS_NAME == linux`?

lib/travis/build/bash/travis_install_go_dependencies.bash
```diff
@@ -0,0 +1,30 @@
...
+          travis_cmd godep\ restore --retry --timing --assert --echo
+        fi
+      fi
+    fi
+  fi
+
+  if travis_has_makefile; then
+    echo 'Makefile detected'
+  else
+    travis_cmd "go get ${gobuild_args} ./..." --retry
```
nono: Shouldn't we have `-t` here, to also get the test dependencies?

meatballhat: @nono I'm working on a follow-up PR that should address this :+1:

PR #1638, opened 2019-01-11
Match only [0-9] and '.' in travis_vers2int()

BanzaiMan:
Some tools such as `npm` may return unusual values such as 
`6.5.0-next.0`.


BanzaiMan:
See https://travis-ci.community/t/node-project-generating-error-from-rvm-after-use-hook/1671

BanzaiMan:
1. `--only` is a valid flag for `grep` on both Linux and Mac.
1. We may also consider not matching with the start of the line with `^`, in case strings such as `v1.2.3` are passed.

PR #1639, opened 2019-01-14
Install Bundler if not found

BanzaiMan:
We add `bundler` to RVM's "global" gemset: https://github.com/travis-ci/travis-build/blob/5caec04f7f6a411c55093c9ffafbff705824a6e6/lib/travis/build/script/shared/rvm.rb#L169 This means that during the on-demand installation of Ruby, we attempt to install the latest Bundler.

As of this writing, the latest version of Bundler is [2.0.1](https://rubygems.org/gems/bundler/versions/2.0.1), and it requires Ruby 2.3.0. On older Rubies which do not have Bundler _within_ the archives, the installation fails (though there is a warning), and the users are left with a Ruby installation without Bundler, even though we do invoke `bundle` by default.

This is not ideal.

This PR attempts to install an appropriate version of Bundler when the command is not found.

### DIFF
--- a/lib/travis/build/script/shared/rvm.rb
+++ b/lib/travis/build/script/shared/rvm.rb
@@ -25,6 +25,8 @@ module Travis
           pkuczynski
         )
 
+        BUNDLER2_RUBY = '2.3.0'
+
         def export
           super
           sh.export 'TRAVIS_RUBY_VERSION', version, echo: false if rvm?
@@ -77,6 +79,8 @@ module Travis
             sh.cmd('type rvm &>/dev/null || source ~/.rvm/scripts/rvm', echo: false, assert: false, timing: false)
             sh.file '$rvm_path/user/db', CONFIG.join("\n")
             send rvm_strategy
+
+            install_bundler
           end
 
           def rvm_strategy
@@ -185,6 +189,18 @@ module Travis
 
           def without_teeny?(version)
             version =~ /\A(\d+)(\.\d+)\z/
+          end
+
+          def install_bundler
+            sh.if "! $(command -v bundler)" do
+              sh.echo "Installing Bundler", ansi: :yellow
+              sh.if "$(travis_vers2int \"$(ruby -e 'puts RUBY_VERSION')\") -lt $(travis_vers2int #{BUNDLER2_RUBY})" do
+                sh.cmd "gem install bundler -v '< 2'"
+              end
+              sh.else do
+                sh.cmd "gem install bundler"
+              end
+            end
           end
       end
     end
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Tested https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/736963#L129 (2.2.10)

https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/736962 skips Bundler installation, since 2.3.7 on-demand installation succeeds.

lib/travis/build/script/shared/rvm.rb
```diff
@@ -186,6 +190,18 @@ def setup_rvm_aliases
           def without_teeny?(version)
             version =~ /\A(\d+)(\.\d+)\z/
           end
+
+          def install_bundler
+            sh.if "! $(command -v bundler)" do
```
joepvd: Think this leaks the path to bundler if it is found.

lib/travis/build/script/shared/rvm.rb
```diff
@@ -186,6 +190,18 @@ def setup_rvm_aliases
...
+              sh.echo "Installing Bundler", ansi: :yellow
```
joepvd: Could this be wrapped in a fold? And maybe also a newline before the fold?
I'd say the name of the fold should just be the command, the header `Installing Bundler` seems unnecessary when compared to `rvm use` line.

BanzaiMan:
Re-ran the build. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/736963#L128

joepvd:
:trophy:

khiav223577:
@BanzaiMan The installation is broken in Ruby 1.8.7. Could you take it a look?

See the [broken job](https://travis-ci.com/github/khiav223577/case_register/jobs/469389394) and the raw job log:
```
e[0Ktravis_fold:end:rvm
e[0Ktravis_fold:start:install_bundler
e[0Ktravis_time:start:072c452a
e[0K$ gem install bundler -v '< 2'
ERROR:  Could not find a valid gem 'bundler' (< 2), here is why:
          Unable to download data from https://rubygems.org/ - hostname was not match with the server certificate (https://rubygems.org/specs.4.8.gz)
ERROR:  While executing gem ... (TypeError)
    can't modify frozen object
travis_time:end:072c452a:start=1609997846431058838,finish=1609997846557437923,duration=126379085,event=
e[0Ke[31;1mThe command "gem install bundler -v '< 2'" failed and exited with 1 during .e[0m

Your build has been stopped.
```

---

I guess `rubygems` needs to be updated before installing `bundler` gem.

The following is the workaround I use now. Unfortunately, `travis_fold:start:install_bundler` runs before `before_install`. I have no chance to run my patch.
```rb
before_install:
  - if `ruby -e 'exit(RUBY_VERSION.to_f < 2.7)'`; then
      gem i rubygems-update -v '< 3' && update_rubygems;
      gem install bundler -v '< 2';
    fi
```

PR #1645, opened 2019-01-24
Add ability to specify missing GPG keys at build time

BanzaiMan:
The keys to receive are specified by the environment variable
`TRAVIS_BUILD_APT_GPG_KEYS`, which contains a comma-separated list of
GPG key IDs.


BanzaiMan:
It has also been suggested that the keys should be selectively (depending on the Ubuntu release) fetched.

BanzaiMan:
Another approach is to set up a mechanism similar to other `apt` addon directives; maintain a list of GPG keys for each Ubuntu release and apply them depending on the release at build script compile time.

PR #1646, opened 2019-01-27
Add "-d" to "go get"

Al2Klimov:


Al2Klimov:
Rationale: https://travis-ci.org/Al2Klimov/check_linux_sensors/jobs/485137101

BanzaiMan:
Sorry, but I don't see why this is desirable in general. In https://golang.org/cmd/go/, I see:

> The -d flag instructs get to stop after downloading the packages; that is, it instructs get not to install the packages. 

Since this command is run by default for all Go builds in the `install` phase, it is desirable that the phase installs the packages.

It seems to me that if your build is failing because of the lack of the `-d` flag, then I suggest you override the `install` phase.

If you disagree, please include more information.

I'm closing this for now.

PR #1648, opened 2019-02-04
[WIP]Add flag to enable/disable network checks

ggcampinho-travis:
The flag was added for `enterprise-2.2` branch and we are porting it to `master`, so we can stay in sync (https://github.com/travis-ci/travis-build/commit/491197656370cf77a1fe709aa6e4480efd1aabe5)

### DIFF
--- a/lib/travis/build/config.rb
+++ b/lib/travis/build/config.rb
@@ -96,7 +96,7 @@ module Travis
           ),
         },
         network: {
-          wait?: ENV.fetch('TRAVIS_BUILD_NETWORK_WAIT', 'true').to_bool,
+          wait: ENV.fetch('TRAVIS_BUILD_NETWORK_WAIT', 'true').to_bool,
           wait_retries: Integer(ENV.fetch(
             'TRAVIS_BUILD_NETWORK_WAIT_RETRIES',
             ENV.fetch('NETWORK_WAIT_RETRIES', '20')
--- a/lib/travis/build/config.rb
+++ b/lib/travis/build/config.rb
@@ -96,6 +96,7 @@ module Travis
           ),
         },
         network: {
+          wait?: ENV.fetch('TRAVIS_BUILD_NETWORK_WAIT', 'true').to_bool,
           wait_retries: Integer(ENV.fetch(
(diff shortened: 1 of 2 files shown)


ggcampinho-travis:
I just decided to follow the patterns that exist in the files, adding the configuration below network and respecting the naming conventions, besides that everything would be the same, except for a problem that is what I mention on my comment before.

The `TRAVIS_WAIT_FOR_NETWORK_CHECK` variable actually skips the network check instead of waiting for it, so I tried to resolve the confusion. Since we are bumping a major version on Enterprise, I think we should communicate the changes and remove the old one to avoid confusion.

lib/travis/build/config.rb
```diff
@@ -96,6 +96,7 @@ def sc_data
           ),
         },
         network: {
+          wait?: ENV.fetch('TRAVIS_BUILD_NETWORK_WAIT', 'true').to_bool,
```
philpeble: Are we sure that naming this with a `?` won't cause issues? I think it is much more clear but I don't see any examples of this usage anywhere.

Something to monitor during testing!

ggcampinho-travis:
Note also that all the config names start with `TRAVIS_BUILD_`

lib/travis/build/config.rb
ggcampinho-travis: I checked the gem, and it actually does this for us (https://github.com/svenfuchs/hashr)

ggcampinho-travis:
I also think we should wait and also I want to test this somehow because it doesn't contain any tests

philpeble:
@ggcampinho-travis If you check out [this comment from Danish](https://github.com/travis-pro/travis-enterprise/issues/359#issuecomment-460380645) in the issue surrounding this in the `travis-enterprise` repo it sounds like maybe this commit is actually not required. We could accomplish the same thing by setting an existing env var to an empty string.

Whether we want to actually make this available for configuration for customers is, to me, a moot point for this PR. All we wanted to accomplish is to say "we can use the master version of `travis-build` in enterprise" and this seems to say to me that the answer is "Yes". 

What do you think?

PR #1650, opened 2019-02-06
Add retries for language archive downloads

BanzaiMan:
To reduce the effects of flaky network conditions

### DIFF
--- a/lib/travis/build/script/node_js/manager/nvm.rb
+++ b/lib/travis/build/script/node_js/manager/nvm.rb
@@ -51,7 +51,7 @@ module Travis
 
           def install_version(ver)
             sh.fold "nvm.install" do
-              sh.cmd "nvm install #{ver}", assert: false, timing: true
+              sh.cmd "nvm install #{ver}#{devnull}", assert: false, timing: true
               sh.if '$? -ne 0' do
                 sh.echo "Failed to install #{ver}. Remote repository may not be reachable.", ansi: :red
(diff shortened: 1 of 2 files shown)


BanzaiMan:
We could also add a config value for the number of retries, instead of hard coding 2.

BanzaiMan:
https://travis-ci.community/t/travis-python-archive-download-is-flaky/2147

webknjaz:
Hi @BanzaiMan, 
In distributed systems it's usually worth having a bit more complicated logic. It's usually a good practice to use exponential backoff for calculating delays.

Here's a pure bash implementation of an exponential backoff, feel free to reuse it:
https://github.com/webknjaz/ansible-gentoo-laptop/blob/master/vault/cat-keybase.sh

BanzaiMan:
@webknjaz My understanding is that `curl`'s `--retry` does exponential backoff by itself.

https://curl.haxx.se/docs/manpage.html#--retry

> When curl is about to retry a transfer, it will first wait one second and then for all forthcoming retries it will double the waiting time until it reaches 10 minutes which then will be the delay between the rest of the retries. By using --retry-delay you disable this exponential backoff algorithm. See also --retry-max-time to limit the total time allowed for retries.

BanzaiMan:
2 is probably insufficient for the issues we're seeing.

webknjaz:
Oh, fair enough! How about 5 retries?

PR #1653, opened 2019-02-07
Add GPG based externally maintained list

BanzaiMan:
The list's URL is indicated by the environment variable
`$TRAVIS_BUILD_GPG_KEYS_URL`.

The URL should point to a raw JSON file, whose content is of the form:

    [
        {
            "name": "descriptive-name",
            "id": "GPG_KEY_ID"
        }
    ]


BanzaiMan:
Like #1645, but with a list of GPG keys which is separately maintained. Tentatively, that list is https://github.com/travis-ci/apt-gpg-keys.

BanzaiMan:
Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/737302#L82-L88

BanzaiMan:
Perhaps the keys should be definable for each Ubuntu release?

piem:
Hello @BanzaiMan !

what is the status of this PR?

aubio builds are now [failing](https://travis-ci.org/aubio/aubio/builds/513866910) since adding two more packages (libvorbis-dev and libflac0-dev), . The error message mentions `git-lfs` which is not required by this build and seems unrelated. any clue?

thanks for your work on travis,

cheers, piem

native-api:
Does this command really need to be run every time in a build? I reckon just updating Linux images once in a while will be much simpler.

BanzaiMan:
I believe that we've removed many (all?) third-party APT repositories from the Xenial image, so the issue should be less pronounced. We can therefore narrow the scope of this code to older images, which will not be updated.

PR #1656, opened 2019-02-14
Use CDN for language archives

BanzaiMan:
This PR alters where we look for language archives from the S3 buckets to CDN.

By using CDN, we accomplish two things:

1. Speed up language archive downloads
1. Allow Enterprise users to sidestep the S3 requriement


BanzaiMan:
At the moment, only Python archives are set up for delivery via CDN, but it is expected that all other languages (Ruby, Perl, PHP, Erlang) will be set up as well.

PR #1659, opened 2019-02-14
Adding Git Wire Protocol v2

cotsog:


lib/travis/build/appliances/git_wire_protocol_v2.rb
```diff
@@ -0,0 +1,22 @@
...
+
+        GIT_VERSION = /(\d)\.(\d+)\.(\d+)/
+        GIT_WIRE_PROTOCOL_V2_MAJOR = 2
+        GIT_WIRE_PROTOCOL_V2_MINOR = 18
+        
+        def apply
+          git_version = `git version`
+          major, minor, build = git_version.match(GIT_VERSION).captures
+          if major.to_i => GIT_WIRE_PROTOCOL_V2_MAJOR && minor.to_i => GIT_WIRE_PROTOCOL_V2_MINOR
+            sh.cmd "git config --global protocol.version 2", assert: false, echo: true
```
joshk: I think you can always apply this as it won't do anything for versions under 2.18

cotsog: Nice! Updated!

lib/travis/build/appliances/git_wire_protocol_v2.rb
```diff
@@ -0,0 +1,13 @@
+require 'travis/build/appliances/base'
+
+module Travis
+  module Build
+    module Appliances
+      class GitWireProtocolV2 < Base
+        def apply
+            sh.cmd "git config --global protocol.version 2", assert: false, echo: false
```
joepvd: indentation is off

lib/travis/build/appliances/git_wire_protocol_v2.rb
```diff
@@ -0,0 +1,13 @@
+require 'travis/build/appliances/base'
+
+module Travis
+  module Build
+    module Appliances
+      class GitWireProtocolV2 < Base
+        def apply
```
joepvd: Could there be a clause specifying the images where this should apply?
Reason is mostly to have it documented where this should apply, so that this workaround can get removed when the images have been fixed/removed.

joshk: Could that be something we add after?

lib/travis/build/appliances/git_wire_protocol_v2.rb
joshk:
```suggestion
          sh.cmd "git config --global protocol.version 2", assert: false, echo: false
```

jeroen:
@cotsog does this make homebrew git a hard dependency? We use a feature that explicitly tests on a setup on which we uninstall homebrew: https://github.com/travis-ci/travis-build/blob/bd9714b5a920630b147a91f0a170583f5e02890c/lib/travis/build/script/r.rb#L542-L547

These builds are now failing [like this](https://travis-ci.org/jeroen/curl/jobs/494186788):

```
/Users/travis/.travis/functions: line 315: /usr/local/bin/git: No such file or directory
```

How do we opt-out of this to use native `/usr/bin/git` instead of homebrew `/usr/local/bin/git`?

cotsog:
@jeroen This change shouldn't make homebrew git a hard dependency.

[Looking at your build history](https://travis-ci.org/jeroen/curl/builds), it seems your last succeeding build was one month ago hence it's possible that it's another change that triggered this error.

I'll look further into your issue this week.

Thank you for your patience!

cotsog:
Changes since last succeeding build: https://github.com/travis-ci/travis-build/compare/5a1fdb7c4...7713e6a71

jeroen:
@catsog this log was just one example, the problem started appearing for all of my repositories since 2 days ago. Here is another example that proves that the problem started happening exactly when the PR above got merged: https://travis-ci.org/ropensci/git2r/builds

It looks like this PR somehow sets the default git to `/usr/local/bin/git` instead of just `git`...

BanzaiMan:
@jeroen That file is generated dynamically, so it is hard to guess what's in it. However, I have a theory.

At the beginning of the R build, the bash process running the build script caches the path to `git` as `/usr/local/bin/git`, which is just a symbolic link to the Homebrew-installed `git` executable, something to the effect of:

```sh-session
$ ls -l `which git`
lrwxr-xr-x  1 travis  staff  28 Oct 27 19:29 /usr/local/bin/git -> ../Cellar/git/2.19.1/bin/git
```

When the R script uninstalls Homebrew, this symbolic link is lost, and it causes the error.

You might be wondering, "Why did this not happen to the older build? I have been using this configuration for a while and it never failed!" Well, my guess for this is that, we changed the way we define and execute `bash` function in the recent months, and this caused this problem to surface.

jeroen:
@BanzaiMan perhaps we can use [this solution](https://unix.stackexchange.com/a/5610) to clear the cash in the `uninstall_homebrew` script then. I'll send a PR for this.

PR #1668, opened 2019-02-22
Add configurable tracing to git commands

cotsog:


lib/travis/build/git/clone.rb
```diff
@@ -17,6 +17,53 @@ def apply
...
+            if ENV["TRACE_COMMAND"].to_s == TRACE_COMMAND_STRACE
+              TRACE_COMMAND_STRACE
+            else
+              DEFAULT_TRACE_COMMAND
+            end
+          end
+
+          def git_clone
+            if trace_git_commands?
+              sh.cmd "#{trace_command} git clone #{clone_args} #{data.source_url} #{dir}", assert: false, retry: true
```
BanzaiMan: I think you can simply this logic by combining the `git` command into a method:

```ruby
def git_cmd
  trace_git_commands? ? "#{trace_command} git" : "git"
end

def git_clone
  sh.cmd "#{git_command} clone …" , …
end

def git_fetch
  sh.cmd "#{git_command} -C #{dir} …", …
end

def fetch_ref # maybe we want to rename this method as `git_fetch_ref` for consistency
  sh.cmd "#{git_command} fetch …", …
end
```

lib/travis/build/git/clone.rb
```diff
@@ -17,6 +17,53 @@ def apply
...
+          def repo_slug
+            data.repository[:slug].to_s
+          end
+
+          def owner_login
+            repo_slug.split('/').first
+          end
+
+          def trace_git_commands_owners
+            ENV["TRACE_GIT_COMMANDS_OWNERS"].to_s.split(',')
```
BanzaiMan: I think it is better to put this logic in `Travis::Build.config` (via `lib/travis/build/cofig.rb`).

lib/travis/build/git/clone.rb
```diff
@@ -17,6 +17,53 @@ def apply
...
+          end
+
+          def trace_git_commands_slugs
+            ENV["TRACE_GIT_COMMANDS_SLUGS"].to_s.split(',')
```
BanzaiMan: Ditto.

lib/travis/build/git/clone.rb
```diff
@@ -11,12 +11,59 @@ def apply
             clone_or_fetch
             sh.cd dir
             fetch_ref if fetch_ref?
-            checkout
+            sh.cmd "git fsck", assert: false, retry: true if trace_git_commands?
```
BanzaiMan: What happened to `#checkout`?

cotsog: Oops, good catch! Fixed.

lib/travis/build/git/clone.rb
cotsog: Nice! Fixed.

PR #1671, opened 2019-03-04
Add retrospective message for default change

BanzaiMan:
So users who hadn't built during the warning period would see the message
after the fact.


BanzaiMan:
We will rely on the web banner to relay this change post date. https://github.com/travis-ci/travis-web/pull/1968

PR #1672, opened 2019-03-07
Add Hack language support

BanzaiMan:
This PR adds Hack language support, superseding #1318.


BanzaiMan:
A test build is here: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/737688

This PR currently:
1. Accepts `version` key with values listed in https://docs.hhvm.com/hhvm/installation/linux.
1. Has no default `install` or `script`. As mentioned in https://github.com/travis-ci/travis-build/pull/1318#issuecomment-463319842, Hack does not support many common PHP tools such as `composer`. (And I've also tested `phpunit`.) So there is no clear tools to use for these.
1. Inherits from `Script::Php`, because HHVM-related methods are useful. However, if we think HHVM is not useful for PHP, we should simply move those methods to the new class. (This will simplify the `#cache_slug` method.)

BanzaiMan:
@fredemmott Do you perhaps have a sample Hack repo with which I can test this PR?

fredemmott:
https://github.com/hhvm/hsl - though it's currently Docker-based. You might find the parent of https://github.com/hhvm/hsl/commit/61f5e5ebe1956862d7f80917e360c09fd8263068 (PHP-based) a better starting point

fredemmott:
Also, for clarity, both real PHP and HHVM are required right now: HHVM is required to execute Hack code, PHP is required to execute the package manager (we still use PHP's one, and HHVM can't execute it any more)

BanzaiMan:
@fredemmott Thanks for the info. Does the PHP component involve any sort of requirement (for example, only PHP 7.0+ is supported), or will any functioning `composer` do?

fredemmott:
As long as the PHP is real PHP, and able to execute an up-to-date composer, it'll work. That said, I have a preference for using a version of PHP that's [not reached EOL](http://php.net/supported-versions.php) - so 7.1+.

lib/travis/build/script/hack.rb
```diff
@@ -0,0 +1,68 @@
+module Travis
+  module Build
+    class Script
+      class Hack < Php
+        DEFAULTS = {
+          hhvm: 'hhvm', # lts
```
fredemmott: Is this a change or incorrect comment? Currently, 'hhvm' gets you the latest release, which might not be an LTS

lib/travis/build/script/hack.rb
```diff
@@ -0,0 +1,68 @@
...
+          php: '7.2',
+        }
+
+        VALID_HHVM = %w(
+          hhvm
+          hhvm-dbb
```
fredemmott: - I'm guessing this a typo of debug
- HHVM now follows standard-ish debian packaging: -dbg packages are just debug symbols for the main package - they're no longer include debug executables. It doesn't really make sense to select them /instead of/ a non-debug package.
- it was possible to specify arbitrary versions, e.g. '3.21-lts' - is it possible to bring this back rather than use the whitelist?

lib/travis/build/script/hack.rb
```diff
@@ -0,0 +1,68 @@
...
+          hhvm-nightly
+          hhvm-nightly-dbg
+          hhvm-dev-nightly
```
fredemmott: This doesn't exist

lib/travis/build/script/hack.rb
BanzaiMan: I see this now. I'll look into LTS further.

lib/travis/build/script/hack.rb
BanzaiMan: These values are taken from the official installation guide at https://docs.hhvm.com/hhvm/installation/linux#other-packages. (Though, as you correctly point out, `dbb` is a typo of `dbg`.)

lib/travis/build/script/hack.rb
BanzaiMan: Then it should be removed from the documentation (see above).

lib/travis/build/script/hack.rb
BanzaiMan: I'll look into LTS. However, I'll point out that it is rather unwieldy to validate the user input; https://docs.hhvm.com/hhvm/installation/release-schedule#Lifecycle says 3.27 and 3.30 are LTS, but 3.27 is not, and we have no idea how 3.28 behaves. So we will have to rely on users to supply sensible values.

fredemmott: 3.27 is an LTS and is supported
3.28 and 3.29 are unsupported, I’ll update that

BanzaiMan:
Hello, @fredemmott! Are there recent releases of HHVM? https://docs.hhvm.com/hhvm/installation/release-schedule#Lifecycle says 4.0 was the last supported one, and its supported ended on April 8.

BanzaiMan:
Oh, I see that 4.12.0 is the latest.

BanzaiMan:
See also: https://github.com/travis-ci/travis-web/pull/2083

BanzaiMan:
An example build with various HHVM versions: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/742961

fredemmott:
Thanks - it's a bit unclear what's going on with those specific versions:
- 4.12 is not an LTS
- the nightly is failing
- `hhvm-dev` is a package containing C development headers for the `hhvm` package, and doesn't make sense as an hhvm version to target

Sticking with your naming scheme, I'd expect something like:

```
hhvm:
 - hhvm
 - hhvm-nightly
 - hhvm-4.13
```

... to work. Or, alternatively:

```
hhvm:
 - latest
 - nightly
 - 4.13
```

I'll update the docs; since we switched to weekly releases with overlapping support, every new .x release gets a new apt 'dist' - the primary ones, e.g. `xenial` are just used for the latest and nightly packages

BanzaiMan:
* It makes sense to drop the `hhvm-` prefix for HHVM versions. In this case, though, introducing a new keyword like `latest` seems dubious, in case of a future name collision. It's better to allow the empty string or the `nil`/`false` value to indicate the default `hhvm`.
    ```yaml
    hhvm:
      - # default
      - nightly
      - "4.13"
    ```
* The use of `-lts` presents a problem, admittedly. A given version may or may not be an LTS, and it is not clear what should happen if it's not. Should it fail? Should it install the version without the LTS? What if it's not available at all? Or perhaps it should install the "last" LTS before the specified one (assuming that such information is _easily_ obtainable).

  Furthermore, it is not clear how we should maintain the list of LTS versions. I would like to avoid manual maintenance if at all possible.

fredemmott:
> `hhvm-` prefix

Would `stable` be less worrying/clearer about intent?

> LTS prefix and maintenance

Since we moved to weekly releases (starting with 4.3), we now have much more automation and LTSs aren't special - I don't think there's anything that needs to be done here. For every release since 4.3, both LTS (4.8) and non-LTS you can always use `xenial-4.x` apt distribution - `lts-` is no longer part of the apt distribution. I don't think there's any need for Travis to be aware that 4.8 is an LTS and 4.7/4.9 aren't, so no maintenance should be needed

3.30 is the only still-supported release where this isn't true, as it's before the switch to weekly releases

And sorry, yes, https://docs.hhvm.com/hhvm/installation/linux needs a lot of updating since then.

fredemmott:
Filed https://github.com/hhvm/user-documentation/issues/678 for the various documentation issues - I'd appreciate it if you comment with anything else I didn't notice

BanzaiMan:
> Would `stable` be less worrying/clearer about intent?

Any name is acceptable, as long as we are guaranteed that the name will not collide with another. (I also suspect that `nil`/`false` will fall back to the default name anyway (though I have not written any code to do any of this).

BanzaiMan:
At the moment, I'm inclined to accept any of the following:

1. empty string (`nil` or `false`) or `hhvm` — installs `hhvm` from the latest release
1. `hhvm-x.y.z` and `x.y.z` — installs hhvm version x.y.z, with the optional `-lts` suffix
1. `hhvm-dbg` and friends — installs the package with that name

It might be difficult to check the veracity of the strings of the second form. I am inclined to accept them as given, and do the best; if it fails, the user is expected to know how to find valid strings, or we can detect the failure and direct the user to the documentation (preferably on https://docs.hhvm.com) that shows the valid one.

We may also add a special name that can be accepted as an alternative to `hhvm`, but this requires a guarantee that the name will not collide (however unlikely) with a future package name.

BanzaiMan:
No one should be using HHVM for PHP at the moment, and Hack is completely a new feature. Therefore, I feel that this PR is reasonably safe to deploy to production, after a small amount of testing in staging.

BanzaiMan:
In production: https://travis-ci.org/BanzaiMan/travis_production_test/builds/559964585

BanzaiMan:
Needs to reject non-Linux builds. (Either when we handle the requests, or when the request comes to travis-build`.)

BanzaiMan:
See also #1746

fredemmott:
So, if I'm reading https://github.com/travis-ci/travis-build/pull/1750/files#diff-155b33766f0c43dc659b636f5e9d7897R154-R160 correctly, 3.30 is the only supported specific release that can be tested; no 4.x release have the '-lts' thing, so we're back to Docker being the best way to test Hack code; is this still a work in progress?

PR #1676, opened 2019-03-22
Make ffmpeg-dev install dependent on distribution

stephlocke:
When running `dist: xenial` the travis build was failing because the ppa for ffmpeg-dev doesn't have a xenial version. This PR adds a conditional check so it will only run that code if the distribution is trusty.

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -82,7 +82,7 @@ module Travis
 
                 # Both c2d4u and c2d4u3.5 depend on this ppa for ffmpeg
                 if config[:os] == 'trusty'
-                sh.cmd 'sudo add-apt-repository -y "ppa:kirillshkrogalev/ffmpeg-next"'
+                  sh.cmd 'sudo add-apt-repository -y "ppa:kirillshkrogalev/ffmpeg-next"'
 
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/r.rb
```diff
@@ -81,7 +81,9 @@ def configure
                 end
 
                 # Both c2d4u and c2d4u3.5 depend on this ppa for ffmpeg
-                sh.cmd 'sudo add-apt-repository -y "ppa:kirillshkrogalev/ffmpeg-next"'
+                if config[:os] == 'trusty'
```
BanzaiMan: I don't think `config[:os]` will ever be set to `trusty`. I don't think anything will, either.

At run time, you can test with `$(lsb_release -cs) = 'trusty'`:

```ruby
sh.if "$(lsb_release -cs) = 'trusty'" do
  sh.cmd 'sudo add-apt-repository -y "ppa:kirillshkrogalev/ffmpeg-next"'
end
```

stephlocke: Change incorporated! Thanks for the suggestion

stephlocke:
Thank you @BanzaiMan! 🙏

PR #1678, opened 2019-04-06
With 270 days left until Python 2 EOL, it is time to upgrade default Python

cclauss:
…Python

Upgrade from Python 2.7 --> 3.6.  Python 3.7.3 is current but __dist: xenial__ is required for Python >= 3.7.

PR #1680, opened 2019-04-06
Upgrade default Python from 2.7 --> 3.7

cclauss:
Like #1678 but with Python 3.7 instead of 3.6.  Python >= 3.7 _requires_ __dist: xenial__.


cclauss:
Closing in favor of #1678

PR #1681, opened 2019-04-08
Add warning for upcoming Python default change

BanzaiMan:
To correspond to #1678.

### DIFF
--- a/lib/travis/build/script/python.rb
+++ b/lib/travis/build/script/python.rb
@@ -10,7 +10,7 @@ module Travis
         DEPRECATIONS = [
           {
             name: 'Python',
-            current_default: DEFAULTS[:phython],
+            current_default: DEFAULTS[:python],
             new_default: '3.6',
             cutoff_date: '2019-04-16',
           }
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/python.rb
```diff
@@ -7,6 +7,15 @@ class Python < Script
           virtualenv: { system_site_packages: false }
         }
 
+        DEPRECATIONS = [
+          {
+            name: 'Python',
+            current_default: DEFAULTS[:phython],
```
cotsog: Is this a typo?

```suggestion
            current_default: DEFAULTS[:python],
```

emdantrim: `phython`

BanzaiMan:
> Quick one, where's the code using the `DEPRECATIONS` array?

It's here: https://github.com/travis-ci/travis-build/blob/a8d57ff7ca73f2d5dcef93f2ac61bbece4010465/lib/travis/build/script.rb#L445-L456

cotsog:
Thanks! 🐑:it:

lib/travis/build/script/python.rb
BanzaiMan: 🙈

PR #1682, opened 2019-04-09
Use remotes package by default.

jimhester:


lib/travis/build/script/r.rb
```diff
@@ -322,10 +324,13 @@ def r_install(packages)
         def r_github_install(packages)
           return if packages.empty?
           packages = Array(packages)
-          setup_devtools
+
+          setup_remotes
+          setup_devtools if config[:use_devtools]
```
BanzaiMan: What is the purpose of this new configuration parameter?

jimhester: To allow people to fall back to the previous behavior if they desire.

We lose a little functionality using only the remote package. No printing of the installed packages and reprinting of the checks that failed.

lib/travis/build/script/r.rb
```diff
@@ -322,10 +324,13 @@ def r_install(packages)
...
+
           sh.echo "Installing R packages from GitHub: #{packages.join(', ')}"
           pkg_arg = packages_as_arg(packages)
-          install_script = "devtools::install_github(#{pkg_arg}, build_vignettes = FALSE)"
+          install_script = "remotes::install_github(#{pkg_arg}, build_vignettes = FALSE)"
```
bastistician: This PR has already been merged so I'm not sure if I should comment here... Anyway, I just came here through https://github.com/r-lib/devtools/issues/1979#issuecomment-481281487. When looking at the code changes I noticed that there actually is no `build_vignettes` argument in `remotes::install_github()` as opposed to `devtools::install_github()`. The extra argument doesn't hurt as it gets passed through to `install.packages()` and then to `download.file()` - unused - but it should probably be removed. @jimhester

jimhester: You are right, want to open a PR to fix it?

PR #1685, opened 2019-04-11
Change default arguments of pip install in python.rb 

2fasc:
Set default of pip install arguments to '''pip install --progress-bar off -r''' to mitigate issues raised in https://travis-ci.community/t/large-pip-installs-exceed-maximum-log-length/1599/8

### DIFF
--- a/lib/travis/build/script/python.rb
+++ b/lib/travis/build/script/python.rb
@@ -58,10 +58,10 @@ module Travis
 
         def install
           sh.if '-f Requirements.txt' do
-            sh.cmd 'pip install -r Requirements.txt', fold: 'install', retry: true
+            sh.cmd 'pip install --progress-bar off -r Requirements.txt', fold: 'install', retry: true
           end
           sh.elif '-f requirements.txt' do
-            sh.cmd 'pip install -r requirements.txt', fold: 'install', retry: true
+            sh.cmd 'pip install --progress-bar off -r requirements.txt', fold: 'install', retry: true
           end
           sh.else do
             sh.echo REQUIREMENTS_MISSING # , ansi: :red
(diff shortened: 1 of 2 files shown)


2fasc:
I dont't understand this Failure. Could somebody help me out here?

Failures:
  1) Travis::Build::Script::Python behaves like a bash script full docker container execution runs and passes
     Failure/Error: expect(log_minus_header).to match(/^Done\. Your build exited with 0\./)
     
     
system_info
       Diff:
       @@ -1,2 +1,456 @@
       -/^Done\. Your build exited with 0\./

adamjstewart:
This issue should have been solved by #1629

BanzaiMan:
The error is coming from `pip`'s failure to execute, e.g., https://travis-ci.org/travis-ci/travis-build/jobs/518833458#L1183, and the consequential failure to look for the matching line to ensure that the script we generate exits properly.

It is not clear to me if #1629 properly fixed the issue with the progress bar; at least, there is an indication that it didn't.

2fasc:
> It is not clear to me if #1629 properly fixed the issue with the progress bar; at least, there is an indication that it didn't.

At least not for me - I looked at the raw log file and the progress bar is still present and floods the log.

joepvd:
I think the pips bundled with most of the python's on Trusty are too old to listen to the environment variable (and to react to the option on the command line).

I think on the Xenial images, pip should be fresh enough so the environment variable should be adhered to. Could you run a build with `dist: xenial` and see if the progress bars disappear?

2fasc:
> I think on the Xenial images, pip should be fresh enough so the environment variable should be adhered to. Could you run a build with `dist: xenial` and see if the progress bars disappear?

I tried it and it works! Thank you!

adamjstewart:
Ah, that explains why my PR worked for me.

PR #1693, opened 2019-04-21
Fix perl6 build

patzim:
`rakudobrew` just had a major upgrade which breaks the way rakudobrew is used in travis-build. This change makes sure the old, still working, version of rakudobrew is used.
I did not test this, as this is a bit complex (docker-compose). This change is trivial enough that I'd be willing to just give it a try, especially as the perl6 build is broken at the moment.


patzim:
The referenced PR tadzik/rakudobrew#135 did not work out. This PR is thus still valid and needed.

PR #1695, opened 2019-04-24
 Add double quotes to return tag

DamianSzymanski:
Fix issue with git checkout when tag contains special characters

### DIFF
--- a/lib/travis/build/git/clone.rb
+++ b/lib/travis/build/git/clone.rb
@@ -102,7 +102,7 @@ module Travis
 
           def checkout_ref
             return 'FETCH_HEAD' if data.pull_request
-            return data.tag     if data.tag
+            return tag if data.tag
             data.commit
           end
 


BanzaiMan:
What are problematic tags that this PR is supposed to solve?

DamianSzymanski:
@BanzaiMan https://github.com/travis-ci/bugs-questions-comments-ideas/issues/129

BanzaiMan:
Ah, I see. Could you add a spec for that problematic value, then? I tend to think, though, that we want to shell escape the tags properly rather than quoting with `%Q{""}`. We need to test it, to be sure.

lib/travis/build/git/clone.rb
```diff
@@ -101,8 +101,8 @@ def checkout
           end
 
           def checkout_ref
-            return 'FETCH_HEAD' if data.pull_request
-            return data.tag     if data.tag
+            return 'FETCH_HEAD'      if data.pull_request
+            return %Q{"#{data.tag}"} if data.tag
```
cotsog: Could we call the `tag` method instead i.e. https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/git/clone.rb#L140-L142 ?

DamianSzymanski:
@BanzaiMan I updated to shellescape, please check
@cotsog i am not sure about def tag we will have double if
Here is the build:
https://staging.travis-ci.org/DamianSzymanski/ddd/builds/739212#L180

lib/travis/build/git/clone.rb
```diff
@@ -101,8 +101,8 @@ def checkout
           end
 
           def checkout_ref
-            return 'FETCH_HEAD' if data.pull_request
-            return data.tag     if data.tag
+            return 'FETCH_HEAD'         if data.pull_request
+            return data.tag.shellescape if data.tag
```
BanzaiMan: @DamianSzymanski I think what @cotsog is saying is that this line can be rewritten to:

```ruby
return tag if data.tag
```

DamianSzymanski:
@BanzaiMan makes sense, please check
@cotsog Thanks

PR #1703, opened 2019-05-07
Set Docker daemon's Maximum Transfer Unit to 1460

BanzaiMan:
The default value, 1500, on the docker0 network interface causes a
substantial performance issue with Google's, which is set to 1460.


BanzaiMan:
Additional considerations for Enterprise is still required.

PR #1704, opened 2019-05-10
Only use Git V2 protocol on xcode10.2 image

mjm:
Older images have a new enough Git in `/usr/local/bin/git`, but not in `/usr/bin/git` (which is provided by Apple).

Some tools, particularly Homebrew, need to use the system Git in `/usr/bin/git`, so that one needs to be new enough to use the new protocol as well.

### DIFF
--- a/lib/travis/build/appliances/git_v2.rb
+++ b/lib/travis/build/appliances/git_v2.rb
@@ -9,7 +9,7 @@ module Travis
         end
 
         def apply?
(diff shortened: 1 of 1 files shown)


jeroen:
I'm still getting these errors on 10.1 errors with system git (not the homebrew git):

    fatal: unknown value for config 'protocol.version': 2

What problem does this protocol.version': 2 solve?

PR #1706, opened 2019-05-10
Invoke correct method to turn off apt safelist checking

BanzaiMan:


BanzaiMan:
Unfortunately, I don't remember what this PR was supposed to fix.

BanzaiMan:
This is superfluous, as Hashr does it automatically.

PR #1707, opened 2019-05-15
R: update R version to Bioc-devel 3.10

LiNk-NY:
Problem: 'bioc-devel' configuration should be using R release version `3.6.0`. 
Solution: Change hard-coded values in the `r.rb` file.

Potential alternative solution: Use version mapping at https://bioconductor.org/config.yaml

cc: @jimhester @BanzaiMan

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -601,7 +601,8 @@ module Travis
           when 'bioc-devel'
             config[:bioc_required] = true
             config[:bioc_use_devel] = true
-            normalized_r_version('devel')
+            config[:r] = 'release'
+            normalized_r_version
           when 'bioc-release'
             config[:bioc_required] = true
             config[:bioc_use_devel] = false


lib/travis/build/script/r.rb
```diff
@@ -601,7 +601,8 @@ def normalized_r_version(v=Array(config[:r]).first.to_s)
           when 'bioc-devel'
             config[:bioc_required] = true
             config[:bioc_use_devel] = true
-            normalized_r_version('devel')
```
jimhester: Is there a reason you aren't using `normalized_r_version('release')`?

jimhester:
While I agree using the config to figure this out automatically is appealing, having it in the script as we do now avoids spurious failures do to intermittent networks. This is also the reason I don't use the R versions API, e.g. https://rversions.r-pkg.org/r-release

lib/travis/build/script/r.rb
LiNk-NY: I was just using the code in `when 'bioc-release'` as an example. I can change lines 605 & 610 to `normalized_r_version('release')`

LiNk-NY: it could also be `r_latest`

jimhester:
LGTM

PR #1708, opened 2019-05-21
Add the ability to ignore APT source aliases defined in remote JSON file

BanzaiMan:
This PR does the following:

1. Rename apt source safelist to apt source aliases; we are running in GCE environment nowadays, where there are no restrictions on the additional APT repositories. It is no longer a list of safe APT repositories, just aliases.
1. Add the ability to turn off alias list loading via the environment variable `TRAVIS_BUILD_APT_LOAD_SOURCE_ALIAS_LIST`. By setting this value to `false`, the alias list will be ignored. This is convenient for Enterprise installations.

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -15,6 +15,8 @@ module Travis
           xenial
         ).freeze
 
+        attr_reader :safelisted, :disallowed_while_sudo
+
         class << self
           def package_safelists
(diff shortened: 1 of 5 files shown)


BanzaiMan:
Test with skipping alias list loading: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/739589#L160-L161

Jamiemcasa:
H

PR #1710, opened 2019-05-26
Install Bioconductor in accordance with https://bioconductor.org/inst…

native-api:
Should fix https://travis-ci.community/t/bioconductor-version-error/3332
In any case, this is the correct way to install according to https://bioconductor.org/install/

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -429,6 +429,8 @@ module Travis
                   '  install.packages("BiocManager");'\
                   "if (#{as_r_boolean(config[:bioc_use_devel])})"\
                   ' BiocManager::install(version = "devel");'\
+                  'else'\
+                  ' BiocManager::install();'\
                   'cat(append = TRUE, file = "~/.Rprofile.site", "options(repos = BiocManager::repositories());")'
                 end
                 sh.cmd "Rscript -e '#{bioc_install_script}'", retry: true
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -430,7 +430,7 @@ module Travis
                   "if (#{as_r_boolean(config[:bioc_use_devel])})"\
                   ' BiocManager::install(version = "devel", ask = FALSE);'\
                   'else'\
-                  ' BiocManager::install();'\
+                  ' BiocManager::install(ask = FALSE);'\
                   'cat(append = TRUE, file = "~/.Rprofile.site", "options(repos = BiocManager::repositories());")'
                 end
                 sh.cmd "Rscript -e '#{bioc_install_script}'", retry: true


BanzaiMan:
@jimhester @jeroen Review, please.

LiNk-NY:
Actually, I don't think this PR needed as there is a subsequent `install` call in:

https://github.com/travis-ci/travis-build/blob/e60caafb45f5a5d585fa81f30411dabce1dcf684/lib/travis/build/script/r.rb#L435-L440

That should run `BiocManager::install` the same for release. It is only for devel that we need the version change with `ask = FALSE`. 
@native-api can you close this one? You don't need to merge them anymore.

native-api:
> https://github.com/travis-ci/travis-build/blob/e60caafb45f5a5d585fa81f30411dabce1dcf684/lib/travis/build/script/r.rb#L435-L440
> 
> That should run `BiocManager::install` the same for release. It is only for devel that we need the version change with `ask = FALSE`.

https://travis-ci.community/t/bioconductor-version-error/3332 testifies that just `install.packages("BiocManager")` is insufficient to ensure that we have the latest version of BiocManager itself. Apparently, Bioconductor maintainers do not guarantee that the version at CRAN is the latest release.

`BiocManager::install("<package>")` does not fix that because as per https://cran.r-project.org/web/packages/BiocManager/vignettes/BiocManager.html#basic-use, it doesn't update any installed packages, only `BiocManager::install()` without parameters does.

Finally, if we use anything other than the prescribed installation procedure, we are strictly on unsupported territory -- Bioconductor maintainers do not guarantee that we'll have a working setup. You need a really good reason to try to be smarter than them and willfully violate it -- even if that happens to work now, it can break at any moment in the future.

LiNk-NY:
> https://travis-ci.community/t/bioconductor-version-error/3332 testifies that just `install.packages("BiocManager")` is insufficient to ensure that we have the latest version of BiocManager itself. Apparently, Bioconductor maintainers do not guarantee that the version at CRAN is always the latest release.

Actually, we've been having some difficulties with getting `BiocManager` updated on CRAN. It's **not** that we 'don't' guarantee an updated `CRAN` version, it's that updating the `CRAN` version takes time and we have to work with the `CRAN` maintainers.

> `BiocManager::install("<package>")` does not fix that because as per https://cran.r-project.org/web/packages/BiocManager/vignettes/BiocManager.html#basic-use, it doesn't update any installed packages, only `BiocManager::install()` without parameters does.

The vignette does not say that it doesn't update installed packages. The `update = TRUE` argument is the default. In interactive mode, it asks the user to update packages. As long as `BiocManager::install` is called it should download the appropriate version of `BiocVersion` which will dictate the right Bioconductor version. The example that you posted a link to (1) is a bit old and the code base has changed. I say that because I see no section that installs Bioconductor. 

> Finally, if we use anything other than the prescribed installation procedure, we are strictly on unsupported territory -- Bioconductor maintainers do not guarantee that we'll have a working setup. You need a really good reason to try to be smarter than them and willfully violate it -- even if that happens to work now, it can break at any moment in the future.

We are the maintainers of BiocManager.

native-api:
> We are the maintainers of BiocManager.

Okay, that IS a good reason!

> Actually, we've been having some difficulties with getting BiocManager updated on CRAN.

And I suspect that calling `BiocManager:install()` would download the update from `bioconductor.org` instead and mitigate that.
However, if you say that the above situation is not normal and shouldn't repeat upon the next release, I'm okay with that.

native-api:
> The example that you posted a link to ([1](https://travis-ci.org/SydneyBioX/scMerge/jobs/529095216)) is a bit old and the code base has changed. I say that because I see no section that installs Bioconductor.

I think I know what's happening here! They have `/Users/travis/R/Library` in Travis cache! (Which is a reasonable thing to do to drastically reduce build times, especially if some modules are compiled.)

Should we support such a scenario?
The user can't support it themselves because package installation happens before any user-supplied scripts get control.

LiNk-NY:
The immediate solution is to have the user should clear their cache.

native-api:
It turns out that [caching `/Users/travis/R/Library` is an advertized Travis feature](https://docs.travis-ci.com/user/languages/r/#packrat). So we have to support such a scenario.

native-api:
It turns out that [caching `/Users/travis/R/Library` is an advertized Travis feature](https://docs.travis-ci.com/user/languages/r/#packrat). So we have to support a scenario where they have older package versions in cache.

PR #1712, opened 2019-05-31
Retry more commands in apt addon

k0kubun:
At TravisCI of ruby/ruby repository, we often see random failures on `apt` plugin, but we do not want to see CI failure on that part. 

I'd like to suggest introducing retries for the following failures:

#### sudo -E apt-add-repository -y ... 
* May 26: https://travis-ci.org/ruby/ruby/jobs/537363191
* May 31: https://travis-ci.org/ruby/ruby/jobs/539694628

#### sudo -E apt-get -yq --no-install-suggests --no-install-recommends $(travis_apt_get_options) install ...
* May 26: https://travis-ci.org/ruby/ruby/jobs/537354169
* May 26: https://travis-ci.org/ruby/ruby/jobs/537354958
* May 31: https://travis-ci.org/ruby/ruby/jobs/539259851

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -163,7 +163,7 @@ module Travis
               safelisted.each do |source|
                 sourceline = source['sourceline'].output_safe
                 if sourceline.start_with?('ppa:')
-                  sh.cmd "sudo -E apt-add-repository -y #{sourceline.inspect}", echo: true, assert: true, timing: true
+                  sh.cmd "sudo -E apt-add-repository -y #{sourceline.inspect}", retry: true, echo: true, assert: true, timing: true
                 else
                   sh.cmd "curl -sSL \"#{safelisted_source_key_url(source)}\" | sudo -E apt-key add -", echo: true, assert: true, timing: true
                   # Avoid adding deb-src lines to work around https://bugs.launchpad.net/ubuntu/+source/software-properties/+bug/987264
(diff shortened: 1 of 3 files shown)


BanzaiMan:
Hello there.

We've tried retry logic earlier with #600. This turned out to be rather problematic; the failures with the `apt` command tends to leave the system in an inconsistent state, from which the identical command cannot continue. There are subsequent attempts such as #605 and #1343, but the issue remains.

joshk:
FYI we recently (last weekend) enabled an Apt Caching Proxy for all of .org and .com, this should improve Apt performance and reliability.

k0kubun:
Hello, thank you for your responses!

> We've tried retry logic earlier with #600. This turned out to be rather problematic; the failures with the apt command tends to leave the system in an inconsistent state, from which the identical command cannot continue. There are subsequent attempts such as #605 and #1343, but the issue remains.

I see. Sorry for duplicating the history. So retrying `apt-get ... install ...` may not be a good idea.

How about `apt-add-repository ...` one? Have you ever also retried that and experienced the similar issue, or do you think the underlying cause is the same and it's likely to trigger a similar problem?

> FYI we recently (last weekend) enabled an Apt Caching Proxy for all of .org and .com, this should improve Apt performance and reliability.

That's a good news :)  I'll monitor the reliability for a while.

k0kubun:
We hit a failure on `apt-get install` again https://travis-ci.org/ruby/ruby/jobs/547497336. Is the "Apt Caching Proxy" stale now? Is there anything we can do to make the CI stable?

joshk:
Hi @k0kubun , this is really odd and really interesting. It doesn't look like network connectivity but instead the APT server reporting that no item exists.

Our APT caching will drop items after 3 hours, but then download them again and stream them to the client, so I don't think it is the issue.

I do wonder if we should add the retries logic just to the APT addon.

@BanzaiMan thoughts?

PR #1713, opened 2019-06-01
Fix "Error: Bioconductor version not changed" on changing existing ve…

native-api:
…rsion

Fixes https://travis-ci.community/t/error-bioconductor-version-not-changed/3491

The same change will need to be done for https://github.com/travis-ci/travis-build/pull/1710 since current images have 3.8 preinstalled apparently.

Tested locally.

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -428,7 +428,7 @@ module Travis
                   'if (!requireNamespace("BiocManager", quietly=TRUE))'\
                   '  install.packages("BiocManager");'\
                   "if (#{as_r_boolean(config[:bioc_use_devel])})"\
-                  ' BiocManager::install(version = "devel");'\
+                  ' BiocManager::install(version = "devel", ask = FALSE);'\
                   'cat(append = TRUE, file = "~/.Rprofile.site", "options(repos = BiocManager::repositories());")'
                 end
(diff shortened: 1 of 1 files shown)


BanzaiMan:
@jimhester @jeroen Review, please. If this PR provides similar/related fix to #1710, perhaps combining them would show a better overall picture of the fixes.

jimhester:
I think it look ok, but agree the PRs should probably be merged. @LiNk-NY does this look correct to you?

native-api:
The two PRs technically solve different problems. Yes, this one will need to be altered after #1710 is merged as I commented in the initial post (or #1710 will need to be altered is this one is merged first). I can do this at an appropriate moment if you wish, or do this yourself, fine by me either way.

LiNk-NY:
I agree both PRs should be merged. It looks good to me. 
This is related to https://support.bioconductor.org/p/120679/#120701 and changes https://github.com/Bioconductor/BiocManager/issues/47
Thanks @native-api

LiNk-NY:
**Update:** On second look, we don't really need #1710 because `BiocManager::install` gets called right after which also installs `BiocVersion` for the release version of Bioconductor. We only need the version change to work for Bioc devel which this PR takes care of. 
Thanks!

LiNk-NY:
@BanzaiMan please merge, thanks!

PR #1714, opened 2019-06-03
Build message update.

piotr-travisci:
Texts updates according
https://github.com/travis-ci/bugs-questions-comments-ideas/issues/106

### DIFF
--- a/lib/travis/build/bash/travis_cmd.bash
+++ b/lib/travis/build/bash/travis_cmd.bash
@@ -55,8 +55,8 @@ travis_cmd() {
     result="${?}"
     if [[ -n "${secure}" && "${result}" -ne 0 ]]; then
       echo -e "${ANSI_RED}We were unable to parse one of your secure environment variables.${ANSI_CLEAR}
-${ANSI_RED}Please make sure to escape special characters such as ' ' and $ with .${ANSI_CLEAR}
-${ANSI_RED}For example, thi$isanexample would be typed as thi$isanexample. See https://docs.travis-ci.com/user/encryption-keys.${ANSI_CLEAR}"
+${ANSI_RED}Please make sure to escape special characters such as ' ' (white space) and $ (dollar symbol) with \ (backslash) .${ANSI_CLEAR}
+${ANSI_RED}For example, thi$isanexample would be typed as thi\$isanexample. See https://docs.travis-ci.com/user/encryption-keys.${ANSI_CLEAR}"
     fi
   fi
 
--- a/lib/travis/build/bash/travis_cmd.bash
+++ b/lib/travis/build/bash/travis_cmd.bash
@@ -54,9 +54,9 @@ travis_cmd() {
     fi
     result="${?}"
     if [[ -n "${secure}" && "${result}" -ne 0 ]]; then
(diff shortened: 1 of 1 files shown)


lib/travis/build/bash/travis_cmd.bash
```diff
@@ -54,9 +54,9 @@ travis_cmd() {
     fi
     result="${?}"
     if [[ -n "${secure}" && "${result}" -ne 0 ]]; then
-      echo -e "${ANSI_RED}The previous command failed, possibly due to a malformed secure environment variable.${ANSI_CLEAR}
-${ANSI_RED}Please be sure to escape special characters such as ' ' and '$'.${ANSI_CLEAR}
-${ANSI_RED}For more information, see https://docs.travis-ci.com/user/encryption-keys.${ANSI_CLEAR}"
+      echo -e "${ANSI_RED}We were unable to parse one of your secure environment variables.${ANSI_CLEAR}
+${ANSI_RED}Please make sure to escape special characters such as ' ' and $ with .${ANSI_CLEAR}
```
BanzaiMan: Is there something missing after "with "?

lib/travis/build/bash/travis_cmd.bash
```diff
@@ -54,9 +54,9 @@ travis_cmd() {
...
+${ANSI_RED}For example, thi$isanexample would be typed as thi$isanexample. See https://docs.travis-ci.com/user/encryption-keys.${ANSI_CLEAR}"
```
BanzaiMan: I am not sure if I understand this sentence.

lib/travis/build/bash/travis_cmd.bash
piotr-travisci: "with ." (dot sign). This it how it was described in task.
Maybe it should be explained more precisely with "(dot sign)" description after . sign.
What do you think @MariadeAnton ?

lib/travis/build/bash/travis_cmd.bash
MariadeAnton: Which sentence?

lib/travis/build/bash/travis_cmd.bash
BanzaiMan: Perhaps we can be more explicit in each of these cases, so that the text printed on the logs is:

    Please make sure to escape special characters such as ' ' (white space) and $ (dollar symbol) with \ (backslash) .

lib/travis/build/bash/travis_cmd.bash
BanzaiMan: > For example, thi$isanexample would be typed as thi$isanexample.

`thi$isanexample` appears exactly the same, and does not seem like a good example.

lib/travis/build/bash/travis_cmd.bash
MariadeAnton: that'd be super helpful, yes

lib/travis/build/bash/travis_cmd.bash
MariadeAnton: Oh, again the formatting in markdown escaping haha

The second one should be `thi\$isanexample`

PR #1715, opened 2019-06-06
Add possibility to disable git clone

anarosas:


anarosas:
@joshk yes, sounds like a good idea. 
What should it say? Something like:
`Skipping "git clone" due to configuration` ?

joshk:
Instead of 'due to' maybe 'based on' is better

PR #1721, opened 2019-06-12
Use crystal snap for nightly version

anarosas:


BanzaiMan:
We'd probably want to extract the non-nightly logic to a method, so that we don't have a huge if-else branch.

BanzaiMan:
🤔 This requires use of Xenial (and later) with Crystal nightly. Automatic routing is probably possible, but it also feels like an overkill. Maybe we want to add a warning?

anarosas:
@joshk @MariadeAnton  any thoughts on this? should we just show a warning that crystal nightly will only be available on Xenial?

joshk:
:+1:

BanzaiMan:
On a second thought, I believe it would suffice to install `snapd` on Trusty. It would be a noop on systems releases that has it already.

joshk:
with Trusty being EOL, I think this is a good way to push people to Xenial

BanzaiMan:
Well, it doesn't work anyway. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/740801#L605

```sh-session
$ sudo snap install crystal --classic --edge
error: too early for operation, device not yet seeded or device model not acknowledged
```

PR #1718, opened 2019-06-12
Fix failure when setting Elm version(s)

zaneli:
```yaml
- language: elm
  elm:
    - 0.19.0
```

```sh
$ npm install -g elm@["0.19.0"]
npm ERR! code EINVALIDTAGNAME
npm ERR! Invalid tag name "[0.19.0]": Tags may not have any characters that encodeURIComponent encodes.
npm ERR! A complete log of this run can be found in:
npm ERR!     /home/travis/.npm/_logs/2019-06-12T06_14_44_617Z-debug.log
The command "npm install -g elm@["0.19.0"]" failed and exited with 1 during .
```

PR #1723, opened 2019-06-13
Fix failure when setting Rust version(s)

zaneli:
same as #1718

```yaml
- language: rust
  rust:
    - stable
```

```sh
0.39s$ curl -sSf https://build.travis-ci.org/files/rustup-init.sh | sh -s -- --default-toolchain=$TRAVIS_RUST_VERSION -y
info: downloading installer
Warning: Not forcing TLS v1.2, this is potentially less secure
error: Pre-checks for host and toolchain failed: invalid toolchain name: '["stable"]'
I
```

### DIFF
--- a/lib/travis/build/script/rust.rb
+++ b/lib/travis/build/script/rust.rb
@@ -12,7 +12,7 @@ module Travis
         def export
           super
 
-          sh.export 'TRAVIS_RUST_VERSION', config[:rust].to_s.shellescape, echo: false
+          sh.export 'TRAVIS_RUST_VERSION', version.shellescape, echo: false
         end
 
         def setup_cache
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/rust.rb
```diff
@@ -12,7 +12,7 @@ class Rust < Script
         def export
           super
 
-          sh.export 'TRAVIS_RUST_VERSION', config[:rust].to_s.shellescape, echo: false
+          sh.export 'TRAVIS_RUST_VERSION', version.shellescape, echo: false
```
zaneli: `version` method is here.
https://github.com/travis-ci/travis-build/blob/c5dab5e13794e1c597e810dfa5db40c2eb70bcfa/lib/travis/build/script/rust.rb#L63-L65

PR #1727, opened 2019-06-19
Respect `update: true` in apt addon

BanzaiMan:
Previously, it ran only when we install packages

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -82,7 +82,7 @@ module Travis
         def before_prepare
           sh.fold('apt') do
             add_apt_sources unless config_sources.empty?
-            if config[:update] || !config_packages.empty?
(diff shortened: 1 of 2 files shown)


lib/travis/build/addons/apt.rb
```diff
@@ -206,7 +210,10 @@ def add_apt_packages
                 stop_postgresql
               end
 
-              sh.cmd 'travis_apt_get_update', retry: true, echo: true, timing: true
+              unless config[:update]
+                # this is run elsewhere when sources are added
+                sh.cmd 'travis_apt_get_update', retry: true, echo: true, timing: true
+              end
```
native-api: This will make `update:` work backwards if `packages:` is specified and `sources:` is not.

BanzaiMan:
Test https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/747886

PR #1728, opened 2019-06-20
Handle arch specification in Julia script

ararslan:
Currently the Julia script assumes that the architecture for Linux is x86-64. This adjusts the logic such that `config[:arch]` is handled appropriately for arm64, which Julia calls AArch64.

Note that not all versions of Julia have AArch64 binaries available, as it's not considered a tier 1 platform (so binaries are usually but not always available for releases). Thus requesting a particular Julia version with `arch: arm64` may fail, probably with some kind of weird, cryptic message from curl. The case of nightly binaries with ARM is handled explicitly here though, as we do not now nor have we ever had ARM nightly binaries.

cc @joshk

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -40,6 +40,9 @@ module Travis
             sh.cmd 'CURL_USER_AGENT="Travis-CI $(curl --version | head -n 1)"'
             case config[:os]
             when 'linux'
+              if config[:julia] == 'nightly' && config[:arch] == 'arm64'
+                sh.failure 'Nightly Julia binaries are not available for AArch64'
+              end
               sh.cmd 'mkdir -p ~/julia'
               sh.cmd %Q{curl -A "$CURL_USER_AGENT" -s -L --retry 7 '#{julia_url}' } \
                        '| tar -C ~/julia -x -z --strip-components=1 -f -'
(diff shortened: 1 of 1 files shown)


ararslan:
The push build failed because it can't find my branch, but my branch is on my fork, so a push build shouldn't have been triggered here at all. 🤔

lib/travis/build/script/julia.rb
```diff
@@ -40,6 +40,9 @@ def setup
             sh.cmd 'CURL_USER_AGENT="Travis-CI $(curl --version | head -n 1)"'
             case config[:os]
             when 'linux'
+              if config[:julia] == 'nightly' && config[:arch] == 'arm64'
+                sh.failure 'Nightly Julia binaries are not available for AArch64'
```
BanzaiMan: Is this a typo?

ararslan: Not as far as I can tell. Which part seems like a typo?

BanzaiMan: Oh, I see. I was ignorant of the "AArch64". :-)

ararslan: Oh, okay. Yeah, Julia's names for things aren't the same as like the Debian names, for example, so we call 64-bit ARM "AArch64." If you think it would be better or less confusing for users, I can change it to match the declared `arch` in the YAML. I just figured I'd what Julia calls it.

BanzaiMan:
Let's test this tomorrow (2019-07-09).

ararslan:
@BanzaiMan, did you end up getting to test this yesterday?

BanzaiMan:
@ararslan So sorry! I didn't hear from you, so I did not. Are you available for testing Friday or Monday next week?

ararslan:
Either should be fine.

BanzaiMan:
Super. Let's do Friday! 🎉

ararslan:
Sounds good!

BanzaiMan:
@ararslan This has been deployed to https://staging.travis-ci.org/.

PR #1729, opened 2019-06-21
Restrict envs to specific branches

piotr-travisci:


spec/build/env_spec.rb
```diff
@@ -88,6 +90,15 @@
         expect(keys).to_not include('BAZ')
       end
     end
+    
+    describe 'for env jobs (pull requests) restricted to branch' do
+      it 'includes vars restricted to foo-(dev) branch' do
```
BanzaiMan: Can we try include a case where the request is on a non-`master` branch? The string `master` _could_ be hiding in unexpected places because it is the default value.

PR #1731, opened 2019-06-21
Rust caching improvements

BanzaiMan:


lib/travis/build/script/rust.rb
```diff
@@ -18,7 +18,7 @@ def export
         def setup_cache
           if data.cache?(:cargo)
             sh.fold 'cache.cargo' do
-              directory_cache.add "${TRAVIS_HOME}/.cargo", "target"
+              directory_cache.add "${TRAVIS_HOME}/.cargo", "target", "${TRAVIS_HOME}/.rustup", "${TRAVIS_HOME}/.cache/sccache"
```
BanzaiMan: Note to self: We probably want a separate constant or a method to refer to these directories.

lib/travis/build/script/rust.rb
```diff
@@ -58,6 +58,10 @@ def use_directory_cache?
           super || data.cache?(:cargo)
         end
 
+        def before_cache
+          sh.cmd 'rm -rf "$HOME/.cargo/registry/src"', timing: false, echo: false
```
kinnison: You may want a note that this is to remove the unpacked crate sources because there's no need to cache those.

BanzaiMan: @kinnison Thanks. The documentation PR is found in https://github.com/travis-ci/docs-travis-ci-com/pull/2391. I appreciate your feedback there as well.

PR #1732, opened 2019-06-21
Set Julia release version to 1.0

ararslan:
Currently `julia: release` uses Julia 0.6, which is long past end of life. This change sets `release` to point to 1.0, the current long term support version, and slightly simplifies the definition for `release`
rather than duplicating logic.

Do note that this will break builds that are currently assuming that `release` is 0.6, but then again, that is the entire design of `release`, which shouldn't be really be used anyway.

cc @staticfloat @StefanKarpinski

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -121,10 +121,12 @@ module Travis
               ext = 'mac64.dmg'
               nightlyext = ext
             end
-            case julia_version = Array(config[:julia]).first.to_s
-            when 'release'
+            julia_version = Array(config[:julia]).first.to_s
+            if julia_version == 'release'
               # CHANGEME on new minor releases (once or twice a year)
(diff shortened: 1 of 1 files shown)


staticfloat:
I thought we had agreed to stop using `release` completely (which is why we left it at 0.6) and only encourage people to use `1`, `1.2`, etc...  The reasoning being that there's no reason why someone would want their code to be tested against Julia 1.X, and then suddenly have it being tested at 2.X when Julia 2 comes out; we should only allow people to select a single "compatibility run" at a time, as it were. (with an obvious exception for `nightly`)

ararslan:
> I thought we had agreed to stop using `release` completely (which is why we left it at 0.6)

I knew there had to be some reason why it hadn't been updated from 0.6. Then can we remove `release` entirely and just set the default version if none is specified to 1.0?

ararslan:
Or even better, remove the default entirely

ararslan:
Oh, derp: #1447.

PR #1733, opened 2019-06-24
Set CC, CC_FOR_BUILD, CXX, CXX_FOR_BUILD only if user does not

BanzaiMan:


BanzaiMan:
See https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/741334#L515 for a test.

PR #1736, opened 2019-06-27
Inject GCP Maven mirror config

BanzaiMan:
To mitigate the effects of https://www.traviscistatus.com/incidents/mm020jn4nzcn.

### DIFF
--- a/lib/travis/build/appliances/maven_central_mirror.rb
+++ b/lib/travis/build/appliances/maven_central_mirror.rb
@@ -3,14 +3,14 @@ module Travis
   module Build
     module Appliances
       class MavenCentralMirror < Base
-        GOOGLE_MAVEN_MIRROR='https://maven-central.storage-download.googleapis.com/repos/central/data/'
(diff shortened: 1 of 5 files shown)


lib/travis/build/appliances/maven_central_mirror.rb
```diff
@@ -0,0 +1,18 @@
+require 'travis/build/appliances/base'
+module Travis
+  module Build
+    module Appliances
+      class MavenCentralMirror < Base
+        GOOGLE_MAVEN_MIRROR='https://maven-central.storage-download.googleapis.com/repos/central/data/'
```
joshk: can this please be configurable via config as we might not want to set a mirror by default

joshk: especially for on prem customers

BanzaiMan: sure. that's reasonable. we should first test this is a feasible solution to the problem.

BanzaiMan:
For the future reference: the information was culled from https://maven-central.storage-download.googleapis.com/index.html.

PR #1737, opened 2019-06-27
 npm caching by default

anarosas:
https://github.com/travis-ci/product/issues/206
updating docs: https://github.com/travis-ci/docs-travis-ci-com/pull/2412/files 

- [x]  tested on .com staging https://staging.travis-ci.com/travis-pro/test-project/builds/373702 
- [x] tested on .org staging https://staging.travis-ci.org/travis-repos/staging-test-repo


BanzaiMan:
We should add a couple of specs that checks that the generated Node.js script:
1. includes commands related to `casher` when the `cache` key is absent in config
1. excludes commands related to `casher` when the config explicitly disable caching with `cache: false` (or equivalent)

spec/build/script/node_js_spec.rb
```diff
@@ -32,6 +32,32 @@
...
+      end
+
+      context 'when cache key is not specified' do
+        it 'adds ${TRAVIS_HOME}/.cache/node_modules to directory cache' do
+          should include_sexp [:cmd, 'rvm $(travis_internal_ruby) --fuzzy do $CASHER_DIR/bin/casher add ${TRAVIS_HOME}/.cache/node_modules', timing: true]
+        end
+      end
+
+      context 'when cache is explicitly set' do
+        let(:config) { { cache: { npm: {} } } }
```
BanzaiMan: I don't think this is something we'd see in production (since it correspond to this YAML snippet:

```yaml
cache:
  npm: {}
```

Perhaps you meant:

```
let(:config) { { cache: "npm" } }
```

spec/build/script/node_js_spec.rb
```diff
@@ -32,6 +32,32 @@
         should include_sexp [:cmd, 'nvm install 0.9', echo: true, timing: true]
       end
 
+      context 'add cache by default' do
+        it 'adds ${TRAVIS_HOME}/.cache/node_modules to directory cache' do
+          should include_sexp [:cmd, 'rvm $(travis_internal_ruby) --fuzzy do $CASHER_DIR/bin/casher add ${TRAVIS_HOME}/.cache/node_modules', timing: true]
+        end
+      end
+
+      context 'when cache key is not specified' do
```
BanzaiMan: This looks like an identical case to the one above.

PR #1738, opened 2019-06-28
Skip setting mirror unless maven is configured

BanzaiMan:
Notably, our Mac images may not have it. https://travis-ci.community/t/previously-working-build-failing-with-services-are-not-supported-on-osx/3960


BanzaiMan:
I don't think we can create this file ahead of time and a later maven installation to preserve this mirror configuration.

PR #1739, opened 2019-07-07
Bump R release to 3.6.1

jeroen:
@jimhester

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -589,7 +589,7 @@ module Travis
 
         def normalized_r_version(v=Array(config[:r]).first.to_s)
           case v
-          when 'release' then '3.6.0'
+          when 'release' then '3.6.1'
           when 'oldrel' then '3.5.3'
           when '3.0' then '3.0.3'
           when '3.1' then '3.1.3'
@@ -597,7 +597,7 @@ module Travis
           when '3.3' then '3.3.3'
           when '3.4' then '3.4.4'
           when '3.5' then '3.5.3'
-          when '3.6' then '3.6.0'
+          when '3.6' then '3.6.1'
           when 'bioc-devel'
             config[:bioc_required] = true
             config[:bioc_use_devel] = true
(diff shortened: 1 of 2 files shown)


BanzaiMan:
Hello, there. Specs needs fixing. Instead of matching `3.6.0`, you can be a little more lax in matching version numbers (say, `\d(\.\d)*`).

jeroen:
I don't understand the suggestion. We don't match `3.6.0` anywhere; we just update the alias for `3.6` and `release` to mean `3.6.1`.

jimhester:
The specs do check that the current version matches what we expect it to, so they need to be updated. There is one additional spec you missed, the regex has an escape, so it is easy to miss, here is the diff to fix it. You can run `bundle exec rspec spec/build/script/r_spec.rb` to verify the tests are fixed locally.

```diff
diff --git a/spec/build/script/r_spec.rb b/spec/build/script/r_spec.rb
index f2ee4a23..14566b1d 100644
--- a/spec/build/script/r_spec.rb
+++ b/spec/build/script/r_spec.rb
@@ -54,7 +54,7 @@ describe Travis::Build::Script::R, :sexp do
   end

   it 'downloads and installs latest R' do
-    should include_sexp [:cmd, %r{^curl.*https://travis-ci\.rstudio\.org/R-3\.6\.0-\$\(lsb_release -cs\)\.xz},
+    should include_sexp [:cmd, %r{^curl.*https://travis-ci\.rstudio\.org/R-3\.6\.1-\$\(lsb_release -cs\)\.xz},
                          assert: true, echo: true, retry: true, timing: true]
   end

(END)
```

BanzaiMan:
```
should include_sexp [:cmd, %r{^curl.*https://travis-ci\.rstudio\.org/R-3\.6\.1-\$\(lsb_release -cs\)\.xz},
```
suffers from the same problem; this will break with the next update.

jeroen:
> suffers from the same problem; this will break with the next update.

So do you suggest we match relax the test to match any version? Doesn't that defeat the purpose of that we are trying to test here?

Either way I updated the PR to fix the versions everywhere.

jimhester:
Thanks for updating it, LGTM now.

PR #1745, opened 2019-07-16
Check presence of `*file` when using shortcuts

BanzaiMan:
For `cache: bundler` and `cache: cocoapods` (and equivalent),
we do not set up caching if the required file is not found.
This needs to be communicated to the users.


BanzaiMan:
https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/743249#L78-L79 shows an Objective-C build with both configured, but both are not properly configured on the repo.

PR #1746, opened 2019-07-17
Fail non-Linux Hack jobs

BanzaiMan:
Until we implement [Homebrew-based installation](https://docs.hhvm.com/hhvm/installation/mac), we should reject `language: hack` + `os: mac`.

### DIFF
--- a/lib/travis/build/script/hack.rb
+++ b/lib/travis/build/script/hack.rb
@@ -26,6 +26,11 @@ module Travis
         attr_accessor :hhvm_version, :hhvm_package_name, :lts_p
 
         def configure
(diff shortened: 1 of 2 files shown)


BanzaiMan:
We don't currently support PHP on Mac, either.

fredemmott:
Filed https://github.com/hhvm/user-documentation/issues/683 as mac has similar changes to the linux builds - `hhvm@3.30-lts`, `hhvm` (latest stable) and `hhvm-nightly` are still available, but all releases since 4.1 are available as `hhvm-4.x`  ,whether LTS or not

PR #1747, opened 2019-07-17
Update default Node.js version to 'lts/*'

BanzaiMan:


BanzaiMan:
Test: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/743257

BanzaiMan:
Coincidentally, default on windows doesn't seem to work currently. https://travis-ci.org/BanzaiMan/travis_production_test/jobs/560066644

BanzaiMan:
Testing on Windows: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/743259

BanzaiMan:
Caching doesn't work for `lts/*`, so we need to fix that before changing the default.

ljharb:
Why doesn’t caching work for that?

Separately, you may want to hold off - I’m planning to add `lts/-1` etc to nvm in the next couple weeks, so a better default might be `lts/-2` which would be the *oldest* LTS instead of the newest one.

BanzaiMan:
Because the `/` is the path separator and that cause the archive name to be invalid.

BanzaiMan:
…or so I thought, but it *_seems_* to be working? https://travis-ci.org/BanzaiMan/travis_production_test/jobs/566660374#L181

lib/travis/build/script/node_js.rb
```diff
@@ -4,7 +4,18 @@ module Travis
...
+        DEFAULTS = {
+          node_js: '0.10',
+        }
+
+        DEPRECATIONS = [
+          {
+            name: 'Node.js',
+            current_default: '0.10',
+            new_default: 'lts/*',
+            cutoff_date: '2019-09-01',
```
ljharb: It’d be great if you could delay this, pending `lts/-2` being available?

BanzaiMan: Sure. This date is completely arbitrary.

PR #1754, opened 2019-07-28
Enable apt addon for Bionic

native-api:
Fixes https://travis-ci.community/t/apt-addon-doesnt-run-in-bionic/4061.

I thought to make the hardcoded dist list unneeded by checking `$(lsb_release -is) == "Ubuntu"` but this can only be done if `travis_build` runs on the build machine. Is that true?

### DIFF
--- a/lib/travis/build/config.rb
+++ b/lib/travis/build/config.rb
@@ -44,12 +44,18 @@ module Travis
             'http://archive.ubuntu.com/ubuntu/'
           )
         },
-        apt_package_safelist: Hash.new { |hash, dist| hash[dist] = \
-            ENV.fetch("TRAVIS_BUILD_APT_PACKAGE_SAFELIST_#{dist.upcase}", '')
+        apt_package_safelist: {
+          precise: ENV.fetch('TRAVIS_BUILD_APT_PACKAGE_SAFELIST_PRECISE', ''),
+          trusty: ENV.fetch('TRAVIS_BUILD_APT_PACKAGE_SAFELIST_TRUSTY', ''),
+          xenial: ENV.fetch('TRAVIS_BUILD_APT_PACKAGE_SAFELIST_XENIAL', ''),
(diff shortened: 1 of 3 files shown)


stouset:
Hope we can get this merged soon!

stouset:
Any chance we can get this merged soon?

lib/travis/build/config.rb
```diff
@@ -44,16 +44,12 @@ def sc_data
             'http://archive.ubuntu.com/ubuntu/'
           )
         },
-        apt_package_safelist: {
-          precise: ENV.fetch('TRAVIS_BUILD_APT_PACKAGE_SAFELIST_PRECISE', ''),
-          trusty: ENV.fetch('TRAVIS_BUILD_APT_PACKAGE_SAFELIST_TRUSTY', ''),
-          xenial: ENV.fetch('TRAVIS_BUILD_APT_PACKAGE_SAFELIST_XENIAL', ''),
+        apt_package_safelist: Hash.new { |hash, dist| hash[dist] = \
```
BanzaiMan: Please expand this to add `bionic` explicitly.

BanzaiMan:
The build script is compiled on a separate machine, and then sent to the VM which runs the job. The job will have no knowledge of where to look for the safelisted sources and packages. This logic could potentially moved to the executioner, at the cost of increased reliance on network resources. (Such a change is _probably_ not friendly to Enterprise, either.)

native-api:
Didn't quite get the explanation -- even if the generated bash script is copied elsewhere afterwards, all the ruby logic seems to run on one machine, and the hash will have the necessary data for the specific job.
But without being able to see firsthand how stuff happens, I'll have to trust your judgement.

native-api:
Didn't quite get the explanation -- even if the generated bash script is copied elsewhere afterwards, all the ruby logic seems to run on one machine, and the hash will have the necessary data for the specific job. So the purpose of duplicating the hardcoded distro list in config is unclear.
But without being able to see firsthand how stuff happens, I'll have to trust your judgement.

PR #1756, opened 2019-08-01
remove redundant condition from r script

jeroen:
The sudo option is no longer needed: https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration.

@jimhester


jimhester:
That is true, LGTM

PR #1760, opened 2019-08-15
Replace myself as a Dart maintainer

nex3:
@athomas has volunteered to be an owner for the time being.


BanzaiMan:
Thanks for all your hitherto contributions!

PR #1761, opened 2019-08-15
use printf rather than echo -e to report command result

lightmare:
This PR makes the command string printed in "The command ${cmd} exited with ${result}" come out verbatim, without interpreting backslash escapes in the command string.

Example showing the issue, with this in .travis.yml:
```yaml
script:
  - printf >>config.py '%s = False\n' BENCHMARK DEMO
```

The command appears in build log twice, first verbatim, but then with a newline in place of `\n`:
```
$ printf >>config.py '%s = False\n' BENCHMARK DEMO
The command "printf >>config.py '%s = False
' BENCHMARK DEMO" exited with 0.
```


BanzaiMan:
Thank you!

PR #1764, opened 2019-08-19
R: fix R installation on macOS for versions < 3.2

nealrichardson:
When installing versions of R older than 3.2 on macOS, the installer fails on a certificate error:

```
$ curl -fLo /tmp/R.pkg http://cloud.r-project.org/bin/macosx/old/R-3.1.3.pkg
Installing OS X binary package for R
$ sudo installer -pkg "/tmp/R.pkg" -target /
installer: Package name is R 3.1.3 for Mac OS X 10.6 or higher (Snow Leopard build)
installer: Certificate used to sign package is not trusted. Use -allowUntrusted to override.
The command "sudo installer -pkg "/tmp/R.pkg" -target /" failed and exited with 1 during .
```

See https://travis-ci.org/ursa-labs/arrow-r-nightly/jobs/573914998 for example. 

This PR adds the -allowUntrusted argument for sufficiently old versions. I added a test assertion that the flag is added, but the test does not confirm that this executes successfully (just as the existing test passes even though builds fail).

cc @jeroen @jimhester

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -147,7 +147,13 @@ module Travis
                 sh.cmd "curl -fLo /tmp/R.pkg #{r_url}", retry: true
 
                 sh.echo 'Installing OS X binary package for R'
-                sh.cmd 'sudo installer -pkg "/tmp/R.pkg" -target /'
+                if r_version_less_than('3.2.0')
+                  # Old R installers would fail with:
+                  # "Certificate used to sign package is not trusted"
+                  sh.cmd 'sudo installer -pkg "/tmp/R.pkg" -target / -allowUntrusted'
+                else
+                  sh.cmd 'sudo installer -pkg "/tmp/R.pkg" -target /'
+                end
                 sh.rm '/tmp/R.pkg'
 
                 setup_fortran_osx if config[:fortran]
--- a/spec/build/script/r_spec.rb
+++ b/spec/build/script/r_spec.rb
@@ -66,11 +66,13 @@ describe Travis::Build::Script::R, :sexp do
     should include_sexp [:cmd, %r{^curl.*bin/macosx/old/R-3\.2\.4-revised\.pkg},
                          assert: true, echo: true, retry: true, timing: true]
   end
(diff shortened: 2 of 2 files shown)


jeroen:
I'm not sure if we need to support such old versions of R on MacOS at all. This version of R was still targeting MacOS mavericks?

Either way I'll let @jimhester weigh in.

nealrichardson:
Right, an alternative solution would be to bump the minimum R version required (or set one at all, I don't see where that is done currently) or at least revise the docs, which claim that versions >= 3.0 are supported. I don't personally need R that old; I was just testing as far back as I could.

jeroen:
Officially CRAN only supports the current and previous release (i.e. R 3.6 and 3.5.). I think RStudio officially tries to support back until R 3.2 where possible, but this is mostly for Linux server. Users on Mac / Windows should just upgrade.

jimhester:
I think I agree with @jeroen, I don't think it is important to support R versions before 3.2 now, particularly if you don't currently have a particular use case for this.

We could update the docs though, the relevant file is https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/languages/r.md

nealrichardson:
Ok, I'll revise the docs instead. The relevant sentence is: 

> Travis CI supports R versions `3.0.3` and above on Ubuntu Precise, Ubuntu Trusty and macOS. 

I can change that to say "Travis CI supports R versions `3.0.3` and above on Ubuntu and `3.2` and above on macOS" unless you're saying that we no longer support the older versions on Ubuntu either. 

Also, is there any reason to qualify which Ubuntu versions are supported? Xenial is the default version now (and I know you can use that with R builds) and Bionic is also available, but neither are named in the docs.

jimhester:
Because not all Ubuntu versions are supported, e.g. we are currently not building R binaries for Bionic. You can update the docs to say Xenial is supported however.

jimhester:
Well knitr and its dependencies no longer work for R < 3.2, so de-facto most packages no longer support < 3.2, I think it is fine to update the docs to say only 3.2+ is supported on all OSs.

PR #1765, opened 2019-08-20
Add freebsd support

DamianSzymanski:
This adds support for freebsd os


lib/travis/build/bash/travis_setup_env.bash
```diff
@@ -29,6 +29,9 @@ travis_setup_env() {
   msys*)
     export TRAVIS_OS_NAME=windows
     ;;
+  freebsd*)
+    export TRAVIS_OS_NAME=unix
```
ararslan: Would it not make more sense for this to be `freebsd`?

PR #1768, opened 2019-08-21
Append additional data to the timer end marker

BanzaiMan:
# Motivation

The additional data would be used to scan the logs and glean how long each step in a job would take.

# Notes
1. The new marker looks like this:
    `travis_time:end:083aa80c:start=1566335769883111186,finish=1566335769886321753,duration=3210567,event=check_unsupported`
1. `travis-web` is not affected by the additional data.
1. The overhead is in the few kilobytes range, based on small primitive tests.

# Specs
~Currently a significant number of specs are failing because this PR adds an additional key-value pair to `Travis::Build::Shell::Builder#cmd`. We need to tweak the matcher so that certain options are ignored.~


lib/travis/shell/generator/bash/cmd.rb
```diff
@@ -34,6 +34,7 @@ def opts(options)
               opts << "--display #{escape(options[:echo])}" if options[:echo].is_a?(String)
               opts << '--retry'  if options[:retry]
               opts << '--timing' if options[:timing]
+              opts << "--stage #{options[:stage].to_s}" if !options[:stage].to_s.empty?
```
BanzaiMan: I note that this _*must*_ precede handling of `secure`; `--secure` in `travis_cmd` adds ` 2>/dev/null`, so that this option must be the last one passed to `travis_cmd`.

BanzaiMan: Or, at least, that's what I observed while testing. (I'm now trying to reason about this with bash, but not exactly clear.)

lib/travis/build/appliances.rb
```diff
@@ -57,6 +57,11 @@ module Build
     module Appliances
       def apply(name)
         app = appliance(name)
+        if app.apply?
+          sh.raw "travis_time_start" if app.time?
+          app.apply
+          sh.raw "travis_time_finish #{name}" if app.time?
+        end
         app.apply if app.apply?
```
BanzaiMan: I think this will push the appliance's S expressions on to the stack twice.

lib/travis/build/bash/travis_cmd.bash
```diff
@@ -27,6 +27,10 @@ travis_cmd() {
       timing=true
       shift
       ;;
+    --stage)
```
BanzaiMan: The name `stage` is not great. Suggestions welcome.

lib/travis/build/appliances.rb
```diff
@@ -55,14 +55,23 @@
...
-        app.apply if app.apply?
+        @app = appliance(name)
+        with_timer(name) { app.apply } if app.apply?
       end
 
       def appliance(name)
         Appliances.const_get(name.to_s.camelize).new(self)
       end
+
+      def with_timer(name)
```
BanzaiMan: We probably want to move this to `Travis::Shell::Builder` if we allow timers to be stacked.

PR #1771, opened 2019-08-28
Don't cache packages from postgresql.org

cotsog:
To address https://travis-ci.community/t/sometimes-build-fails-when-apt-is-updating-postgresql-apt-repository/4872


sjaeckel:
It seems like this patch is incomplete as the issue still persists even though the patch is applied.

see e.g. https://travis-ci.org/libtom/libtomcrypt/jobs/596013723

PR #1789, opened 2019-09-30
Windows build installer changes for Julia 1.4

musm:
cc @ararslan @staticfloat 

If there is a better way to do this please let me know. 

We just need a version check that 1.4 and above (including nightlies) use the new flags.

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -70,7 +70,7 @@ module Travis
             when 'windows'
               sh.cmd %Q{curl -A "$CURL_USER_AGENT" -s -L --retry 7 -o julia-installer.exe '#{julia_url}'}
               sh.cmd 'chmod +x julia-installer.exe'
-              if (config[:julia].split('.').map{|s|s.to_i} <=> [1, 4]) >= 0 || config[:julia] == 'nightly'
+              if Gem::Version.new(config[:julia]) >= Gem::Version.new('1.4') || config[:julia] == 'nightly'
                 sh.cmd %Q{powershell -c 'Start-Process -FilePath julia-installer.exe -ArgumentList "/VERYSILENT /DIR=C:\\julia" -NoNewWindow -Wait'}
               else
                 sh.cmd %Q{powershell -c 'Start-Process -FilePath julia-installer.exe -ArgumentList "/S /D=C:\\julia" -NoNewWindow -Wait'}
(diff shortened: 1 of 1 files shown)


ararslan:
Do you know the precise Julia version where this changed? (You can find out based on the commit SHA using `./contrib/commit-name.sh <sha>` from the Julia source tree)

musm:
We could find out but why would it matter? When Julia 1.4 is released we just need the build script to call the correct branch and this should (hopefully) do that. You can't pass specific commits AFAIK to the julia version in the travis file.

We implemented a simple version check for Appveyor.jl build script as well, so I tried to copy that.

ararslan:
> You can't pass specific commits AFAIK to the julia version in the travis file.

You can pass more specific versions than x.y.

musm:
Ah ok. I personally don't think it's worth the effort to implement such complicated functionality at this point.  Feel free to update this PR if you would like to add this extension or if you have a better way of doing the version checks. For reference on Appveyor.jl we just simply do:  https://github.com/JuliaCI/Appveyor.jl/blob/version-1/bin/install.ps1#L48

musm:
@ararslan  @staticfloat  considerably simplified the verison check by using `Gem` . This is now ready on my end.

lib/travis/build/script/julia.rb
```diff
@@ -70,7 +70,11 @@ def setup
             when 'windows'
               sh.cmd %Q{curl -A "$CURL_USER_AGENT" -s -L --retry 7 -o julia-installer.exe '#{julia_url}'}
               sh.cmd 'chmod +x julia-installer.exe'
-              sh.cmd %Q{powershell -c 'Start-Process -FilePath julia-installer.exe -ArgumentList "/S /D=C:\\julia" -NoNewWindow -Wait'}
+              if Gem::Version.new(config[:julia]) >= Gem::Version.new('1.4') || config[:julia] == 'nightly'
```
ararslan: What happens if you pass `nightly` to `Gem::Version.new`?

musm: (errors)  malformed string

ararslan: Then won't this cause an error for Windows with Julia nightly?

musm: Yes I think you're right. I think I'll just have to switch around the short circuiting .

musm:
anything else to do here?

ararslan:
It would be good to be able to deploy this to staging to ensure it works. A passing test build with at least
```yaml
julia:
  - 1.3
  - nightly
```
would be sufficient I think.

musm:
Yes it would be good for someone to deploy test this before landing, but if it looks OK from someone with more experience I don't see it being a problem.

musm:
marge?

PR #1792, opened 2019-10-04
Show apt debug information if MariaDB package install fails

native-api:
The default apt-get error message is useless. This adds debug output from conflicts resolver that allows to see what exactly is wrong, See https://serverfault.com/questions/986463/get-more-information-from-apt-get-if-it-reports-an-impossible-situation


native-api:
Branch build: https://travis-ci.org/native-api/travis-build/builds/593353390
Failure is unrelated: it requires some Github OAuth token on branch builds.

PR #1794, opened 2019-10-06
travis_terminate: preserve previous return by default

Artoria2e5:
the usual `|| travis_terminate` idiom loses some information by not providing the old return value. Let's fix that.


BanzaiMan:
What is the issue, exactly? `travis_terminate` ends the job _immediately_, and there is no information to preserve when the execution gets to it.

PR #1795, opened 2019-10-09
Show Julia download errors

native-api:
https://travis-ci.community/t/is-jobs-always-needed-when-using-travis-ci-for-julia

Looks like the download fails but we cannot see why because of `curl -s`.

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -58,17 +58,17 @@ module Travis
                 end
               end
               sh.cmd 'mkdir -p ~/julia'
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/julia.rb
```diff
@@ -58,17 +58,17 @@ def setup
                 end
               end
               sh.cmd 'mkdir -p ~/julia'
-              sh.cmd %Q{curl -A "$CURL_USER_AGENT" -s -L --retry 7 '#{julia_url}' } \
+              sh.cmd %Q{curl -A "$CURL_USER_AGENT" -S --fail -L --retry 7 '#{julia_url}' } \
```
BanzaiMan: This is OK, but let us standardize on `-sSf`.

PR #1797, opened 2019-10-11
Add Hack support

BanzaiMan:
Another attempt at adding Hack support. For previous efforts, see #1318, #1672.

The default build logic is based on https://docs.hhvm.com/hack/getting-started/starting-a-real-project#configuring-travisci:

```sh
#!/bin/sh
set -ex
apt update -y
DEBIAN_FRONTEND=noninteractive apt install -y php-cli zip unzip
hhvm --version
php --version

(
  cd $(mktemp -d)
  curl https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
)
composer install

hh_client
vendor/bin/hacktest tests/
if !(hhvm --version | grep -q -- -dev); then
  vendor/bin/hhast-lint
fi
```

### DIFF
--- a/lib/travis/build/script.rb
+++ b/lib/travis/build/script.rb
@@ -35,6 +35,7 @@ require 'travis/build/script/objective_c'
 require 'travis/build/script/perl'
 require 'travis/build/script/perl6'
 require 'travis/build/script/php'
+require 'travis/build/script/hack'
 require 'travis/build/script/pure_java'
 require 'travis/build/script/python'
 require 'travis/build/script/r'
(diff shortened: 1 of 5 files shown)


BanzaiMan:
Test: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/750529

It seems to me that the current composer is not compatible with HHVM at this moment. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/750529#L258

BanzaiMan:
Hi, @fredemmott. 👋 

I've taken another shot at implementing Hack support. I've taken what's written in https://docs.hhvm.com/hack/getting-started/starting-a-real-project#configuring-travisci to decide on what `install` and `script` should be, while retaining HHVM logic in PHP (it may not be in heavy use, but removing that proved problematic in the past).

As you can see in the test job shown above, `composer` seems incompatible. Is that expected?

fredemmott:
Yes; HHVM is no longer a PHP implementation - while we’re still using composer for dependency management, it must be executed with php, not hhvm

fredemmott:
I think the difference with our instructions is that we don’t make “php” execute hhvm, or use phpenv etc. we now consider them entirely separate languages and runtimes

BanzaiMan:
@fredemmott I see. In that case, I suppose having a completely separate logic for Hack makes sense.

PR #1799, opened 2019-10-16
Use rustup 'minimal' distribution by default

woshilapin:
The last version of rustup is exposing different _distributions_ containing each more or less component (see [documentation](https://blog.rust-lang.org/2019/10/15/Rustup-1.20.0.html#profiles)). From a build perspective, the `minimal` profile got the compiler `rustc` and `cargo` which are pretty much all you need. All components like `rustfmt`, `clippy` can still be installed through a `before_script`.

I was not able to make the tests run on my machine to verify I was not breaking anything. But the change being relatively minimal, I'm hoping the CI can tell me it's OK. I might need help if this is not the case.

### DIFF
--- a/lib/travis/build/script/rust.rb
+++ b/lib/travis/build/script/rust.rb
@@ -3,7 +3,7 @@ module Travis
     class Script
       class Rust < Script
         RUST_RUSTUP = 'https://sh.rustup.rs'
-        RUSTUP_CMD = "curl -sSf https://sh.rustup.rs | sh -s -- --default-toolchain=$TRAVIS_RUST_VERSION -y"
+        RUSTUP_CMD = "curl -sSf https://sh.rustup.rs | sh -s -- --default-toolchain=$TRAVIS_RUST_VERSION --profile=minimal -y"
 
         DEFAULTS = {
           rust: 'stable',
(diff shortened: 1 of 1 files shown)


slonopotamus:
Please, yes. This would better be merged before people start massively depending on default profile installing `clippy` and `rustfmt` on Travis.

BanzaiMan:
Rust installation takes only a few seconds. I understand the motivation, but this is a breaking change (it might not be large in number, but those depending on these "non-minimal" tools will break), and I am not sure if it is worth the risk.

slonopotamus:
@BanzaiMan

> Rust installation takes only a few seconds.
This statement is false on Windows nodes. [Three minutes](https://travis-ci.org/slonopotamus/cargo-wix/jobs/599027366#L43).

BanzaiMan:
@slonopotamus Thanks for the extra input. That puts things in a better perspective.

I am now in favor of merging this. Could you add an extra statement to indicate this change? For example: "Rust is now installed with the minimal profile, which does not include tools X, Y, and Z. If you need them, please install them in `before_install`."

Installing Rust with the minimal profile on Windows takes [about 30 seconds](https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/750837#L56).

slonopotamus:
@woshilapin is PR author, not me.

BanzaiMan:
@slonopotamus Ha! Sorry about that. I meant to mention @woshilapin too but I forgot.

BanzaiMan:
There is no versioning with `travis-build`. There is one version used for all builds and jobs.

woshilapin:
Thanks @backspace for the guidance and I'm more than ok to put that into the documentation, you are right about that... although, I'm not sure where I should put that yet. Where exactly do you want me to add this extra statement you're mentioning? Is it in the commit?

As for the documentation, I figure this is going to be somewhere on [this file](https://github.com/travis-ci/docs-travis-ci-com/blob/master/user/languages/rust.md), right?

Thanks @slonopotamus for helping my cause 😄

slonopotamus:
@woshilapin: I believe that's the right file, it ends up on Travis website [here](https://docs.travis-ci.com/user/languages/rust/). Though, since it is in a separate repository, it will mean a separate PR.

woshilapin:
@slonopotamus Thanks, for the confirmation.

It seems @BanzaiMan wanted it also on some other place because he said this.

> We'll also need a similar change to the docs.

woshilapin:
I've just created https://github.com/travis-ci/docs-travis-ci-com/pull/2565 for the documentation part.

PR #1800, opened 2019-10-18
R: disable default -fopenmp flags on MacOS

jeroen:
A lot of people are running into the problem of being unable to install `data.table` dependency on MacOS because there are currently no CRAN binaries, and installing from source fails:

 - https://travis-ci.org/uptake/pkgnet/jobs/596707485
 - https://travis-ci.org/leeper/rio/jobs/593096935

The problem is that the R config defaults to `-fopenmp` in CFLAGS when compiling packages, but the xcode compilers that we use on travis do not support OpenMP. This results in compiler errors for packages like data.table.

@jimhester

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -32,6 +32,7 @@ module Travis
           bioc_check: false,
           bioc_use_devel: false,
           disable_homebrew: false,
+          disable_openmp: true,
           use_devtools: false,
           r: 'release'
         }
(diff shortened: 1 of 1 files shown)


jimhester:
As I said I don't think we should do this.

If we do decide to do it I think it would be slightly better to override the `SHLIB_OPENMP_*FLAGS` in `~/.Rprofile.site`, which we are already using to set the CRAN repository.

jeroen:
OK makes sense.

PR #1801, opened 2019-10-22
Inject job's arch value into cache name

BanzaiMan:
## Problem
Caches do not consider `arch` values when looking for and uploading the cache, so if multiple jobs share the same name but different `arch` values, the builds would fail if the cache is expected to contain binary files for certain CPU.

## A workaround

For this PR, we inject into the list of preferred cache URLs that contains the `arch` value in it. Admittedly this is not a clean way to achieve the result; it is far more logical to modify `Script#cache_slug` to include the `arch` value. Unfortunately, this is problematic because it breaks the `fetch_url` logic and accommodating such a change seems quite complicated. This is necessary because existing caches would not be found otherwise.

Over time, though, (about 45 days) after this code goes into production, all the caches in use would have arch values in them, and we can simply remove this ugly workaround.

## Tests:

1. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/751325#L128 fails to find a suitable cache (`amd64` one first on line 134, then one without arch on line 135), but uploads the cache with the arch value on line 148
1. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/751328#L132 subsequently fetches the cache with the correct name.

### DIFF
--- a/lib/travis/build/script/shared/directory_cache/base.rb
+++ b/lib/travis/build/script/shared/directory_cache/base.rb
@@ -133,18 +133,30 @@ module Travis
 
           def fetch_urls
             urls = [
+              fetch_url(uri_normalize_name(group), true, true),
+              fetch_url(uri_normalize_name(group), false, true),
+              fetch_url(normalize_name(group), true, true),
+              fetch_url(normalize_name(group), false, true),
               fetch_url(uri_normalize_name(group), true),
               fetch_url(uri_normalize_name(group)),
               fetch_url(normalize_name(group), true),
               fetch_url(normalize_name(group))
             ]
             if data.pull_request
+              urls << fetch_url(uri_normalize_name(data.branch), true, true)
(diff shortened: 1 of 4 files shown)


BanzaiMan:
## More tests on com-staging

1. https://staging.travis-ci.com/travis-pro/test-repo-staging1/jobs/490929#L136-L141 On ARM, it tries to find the arch-specific cache, falling back to a generic one (so the existing caches can be found)
    1. the same job creates and uploads the arch-specific cache: https://staging.travis-ci.com/travis-pro/test-repo-staging1/jobs/490929#L153
1. https://staging.travis-ci.com/travis-pro/test-repo-staging1/jobs/490932#L137-L138 shows that the subsequent job found the arch-specific cache correctly.

---

## Interaction with Workspaces

https://staging.travis-ci.com/travis-pro/test-repo-staging1/builds/490938 shows arch-specific caches and workspaces defined in the same build. They should work independently (as they do in this test build).

PR #1802, opened 2019-10-25
Install Cargo Audit in Rust Environment

naftulikay:
This will install Cargo Audit in the Rust build environment, allowing auditing of Cargo dependencies for security flaws.

### DIFF
--- a/lib/travis/build/script/rust.rb
+++ b/lib/travis/build/script/rust.rb
@@ -41,6 +41,12 @@ module Travis
               sh.cmd 'rustup update', echo: true
             end
           end
+
+          sh.fold('cargo-audit-install') do
+            sh.echo 'Installing Cargo Audit', ansi: :yellow
+            sh.cmd 'cargo install --force cargo-audit', echo: true
+            sh.cmd 'cargo generate-lockfile'
+          end
         end
 
         def announce
@@ -49,12 +55,14 @@ module Travis
           sh.cmd 'rustc --version',  assert: true
           sh.cmd 'rustup --version', assert: true
           sh.cmd 'cargo --version',  assert: true
+          sh.cmd 'cargo audit --version', assert: true
           sh.newline
         end
 
         def script
           sh.cmd 'cargo build --verbose'
           sh.cmd 'cargo test --verbose'
+          sh.cmd 'cargo audit'
         end
 
         def cache_slug


BanzaiMan:
Is this a standard tool? I don't know if it is CI's job to do this. If the user wants it, it is easy enough to add.

naftulikay:
So the main issue right now is the installation of `cargo-audit` is an extremely expensive operation (like 10 minute compilation).

@tarcieri and I were discussing this [on Twitter](https://twitter.com/bascule/status/1187818904598372352) and unfortunately right now there are no provided `cargo-audit` binaries available, so installation takes time. There's a way to reduce compilation time by caching it, thankfully.

I'm not sure what the standard practice is with security scanning of dependencies: do environments like NodeJS include the `npm audit` command? If these types of things are those that are always opt-in, then we can probably close this.

tarcieri:
The forthcoming stabilization of the `cargo install-update` command should simplify caching:

https://github.com/RustSec/cargo-audit/issues/143

BanzaiMan:
> unfortunately right now there are no provided cargo-audit binaries available, so installation takes time.

Am I correct to understand that this is going to add "time" to every build? That doesn't seem fair.

Also, the command

    cargo install --force cargo-audit

fails.

https://travis-ci.org/BanzaiMan/travis_production_test/builds/607886989#L206

tarcieri:
@BanzaiMan the minimum supported Rust version for `cargo-audit` is 1.36.0

Regarding adding time to each build, unless the result can be cached, I don't think it makes sense

BanzaiMan:
@tarcieri The command failed on 1.38.0. https://travis-ci.org/BanzaiMan/travis_production_test/builds/607886989#L169

In the end, I tend to think this is not something CI should do. If the user wants it, they can ensure that the version requirement is met, choose to cache binary files if necessary, and so on.

naftulikay:
Yeah I agree and don't see adding a lot of time to the build as being acceptable.

tarcieri:
@BanzaiMan looks like 1.30 to me:

> rustc 1.30.0-nightly (d5a448b3f 2018-08-13)

The error indicates a pre-2018 edition version of Rust. 1.31 was the first version to support the 2018 edition

BanzaiMan:
@tarcieri Hmmm. That's interesting (and strange). I'll look into it.

At any rate… Is there a reason to keep this open?

tarcieri:
If there's no way to cache the install, I'd say close it.

naftulikay:
Yeah, let's close. I was thinking that the Travis environment was similar to Circle CI where images are baked ahead of time but it seems that Travis just runs everything every time.

PR #1803, opened 2019-10-28
Fix missing /etc/nix directory

musteresel:
I noticed 

> tee: /etc/nix/nix.conf: No such file or directory

in my build logs.  tee can create files, but it does not create missing directories. `/etc/nix` does not seem to be existing.  Therefore the error message. Creating the /etc/nix directory before trying to create a file in there fixes this.

I did not test this change in any way.


BanzaiMan:
Could you point to the error you are seeing? Thanks.

musteresel:
Sure, here for example: https://travis-ci.com/musteresel/musteresel.github.io/builds/136784951#L148

PR #1805, opened 2019-10-31
[community: R] update bioc-devel r-version to R-devel

LiNk-NY:
With every Bioconductor release, the `lib/travis/build/script/r.rb` file has to
change so that the R versions are matching with those that are compatible
with Bioconductor release and devel versions. 

Here is the update to that file: 
Bioconductor version `3.11` uses R-devel (`4.0.0`). 

cc: @jimhester 

Reference: https://bioconductor.org/developers/how-to/useDevel/

Best,
Marcel

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -601,8 +601,8 @@ module Travis
           when 'bioc-devel'
             config[:bioc_required] = true
             config[:bioc_use_devel] = true
-            config[:r] = 'release'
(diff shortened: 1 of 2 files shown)


jimhester:
Thanks, you need to update the associated spec as well https://github.com/travis-ci/travis-build/blob/100d373caf54e701307d40e8c4d358394237e0f6/spec/build/script/r_spec.rb#L14

jimhester:
LGTM

PR #1814, opened 2019-11-21
Adapt artifacts to Windows

BanzaiMan:
Enable artifacts addon for Windows. See https://travis-ci.community/t/artifact-uploader-binary-not-found/4978

Before: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/753351#L95
After: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/753353#L91-L92 (The command `artifacts` runs, but fails because it's not configured correctly.)

### DIFF
--- a/lib/travis/build/bash/travis_artifacts_install.bash
+++ b/lib/travis/build/bash/travis_artifacts_install.bash
@@ -2,7 +2,7 @@ travis_artifacts_install() {
   local source="https://s3.amazonaws.com/travis-ci-gmbh/artifacts/stable/build/${TRAVIS_OS_NAME}/${TRAVIS_ARCH}/artifacts"
   local target="${TRAVIS_HOME}/bin/artifacts"
 
-  if [[ ${TRAVIS_OS_NAME} = "windows" ]]; then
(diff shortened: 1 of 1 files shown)


lib/travis/build/bash/travis_artifacts_install.bash
```diff
@@ -2,6 +2,10 @@ travis_artifacts_install() {
   local source="https://s3.amazonaws.com/travis-ci-gmbh/artifacts/stable/build/${TRAVIS_OS_NAME}/${TRAVIS_ARCH}/artifacts"
   local target="${TRAVIS_HOME}/bin/artifacts"
 
+  if [[ ${TRAVIS_OS_NAME} = "windows" ]]; then
```
native-api:
```suggestion
  if [[ ${TRAVIS_OS_NAME} == "windows" ]]; then
```
fixes `shfmt` violation which is what causes build failure. Technically, they are equivalent, but...

BanzaiMan: Thanks for catching that.

TheOpenDevProject:
This would be nice if it could be merged

PR #1816, opened 2019-11-29
allow `exit [n]` to terminate the build early properly

native-api:
https://travis-ci.community/t/exit-0-cannot-exit-successfully-on-arm/5731/7


lib/travis/build/bash/travis_terminate.bash
```diff
@@ -7,6 +7,17 @@ travis_terminate() {
...
+exit() {
+  local code="${1:-$?}"
+  # Should work as normal `exit` in subshells of the main shell.
+  # In unrelated shells, this code won't be imported.
+  if [ $$ -ne $BASHPID ]; then
+    _travis_terminate_exit "$code"
+  else
+    travis_terminate "$code"
+  fi
+}
```
fingolfin, line 19: Would it perhaps be possible (and simpler?) to use `trap travis_handle_exit EXIT` at the start of the main shell script, with `travis_handle_exit` a suitable script function that e.g. just invokes `travis_terminate` or so?

native-api: It looks so!

Tests show that

* The `EXIT` trap is not called upon exiting a subshell, saving us from having to handle that case.
* It's inherited into subshells in `bash` and not inherited in `zsh` (OSX), but since it's not called upon exiting subshells, that doesn't make a difference.

fingolfin:
This affects plenty other people, it seems, see https://travis-ci.community/t/7659/4 -- and as usual I am sure that for every user reporting the issue there's a dozen or more who don't and just give up on this.

PR #1823, opened 2019-12-06
Update tmate static url

DamianSzymanski:


lib/travis/build/appliances/debug_tools.rb
```diff
@@ -70,10 +70,11 @@ def install_dir
           # XXX the following does not apply to OSX
           def static_build_linux_url
             if config[:arch] == 'arm64'
-              "https://#{app_host}/files/tmate-static-linux-arm64v8.tar.xz"
+              arch = "#{config[:arch]}v8"
```
BanzaiMan: I'm thinking that we will be expanding the list of `arch` values in the future. For the sake of generality, then, we may consider this future.

```ruby
case config[:arch]
when 'arm64'
  arch = 'arm64v8'
else
  arch = 'amd64'
end
```
I'm not sure if we want to write `#{config[:arch]}v8` or similar here; it seems better to explicitly state the "arch" value in `tmate`'s file name, rather than somehow implying that it is related to our `arch` value.

PR #1827, opened 2019-12-11
Add support for GemStone on Ubuntu 16.04 and 18.04

fniephaus:
and drop check for POWER architecture (it's not supported anyway).

Closes https://github.com/hpi-swa/smalltalkCI/issues/445#issuecomment-564743137

### DIFF
--- a/lib/travis/build/script/smalltalk.rb
+++ b/lib/travis/build/script/smalltalk.rb
@@ -216,10 +216,14 @@ module Travis
               sh.cmd 'sudo ln -f -s /lib/i386-linux-gnu/libpam.so.0 /lib/libpam.so.0'
               sh.cmd 'sudo ln -f -s /usr/lib/i386-lin-gnu/libstdc++.so.6 /usr/lib/i386-linux-gnu/libstdc++.so'
             end
-            sh.if '$(uname -m) != ppc64le && $(lsb_release -cs) = trusty' do
+            sh.elif '$(lsb_release -cs) = trusty' do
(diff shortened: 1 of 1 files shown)


lib/travis/build/script/smalltalk.rb
```diff
@@ -216,11 +216,15 @@ def gemstone_prepare_linux_dependencies
               sh.cmd 'sudo ln -f -s /lib/i386-linux-gnu/libpam.so.0 /lib/libpam.so.0'
               sh.cmd 'sudo ln -f -s /usr/lib/i386-lin-gnu/libstdc++.so.6 /usr/lib/i386-linux-gnu/libstdc++.so'
             end
-            sh.if '$(uname -m) != ppc64le && $(lsb_release -cs) = trusty' do
+            sh.elif '$(lsb_release -cs) = trusty' do
               sh.cmd 'sudo dpkg --add-architecture i386'
               gemstone_install_linux_dependencies
               sh.cmd 'sudo ln -f -s /usr/lib/i386-lin-gnu/libstdc++.so.6 /usr/lib/i386-linux-gnu/libstdc++.so'
             end
+            sh.elif '$(lsb_release -cs) = xenial || $(lsb_release -cs) = bionic'
```
BanzaiMan: I think you need a `do` at the end of this line.

fniephaus:
To avoid issues like this in the future, I've added code that should let the build fail if the Linux dist isn't supported.

fniephaus:
Thanks for merging, @BanzaiMan!

PR #1833, opened 2020-01-04
nix: update to 2.3.6

asymmetric:
This gets nix from the `latest` [symlink](https://nixos.org/releases/nix/latest/) by default, so this file doesn't need to be kept in sync with nix releases.

### DIFF
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -49,13 +49,7 @@ module Travis
             sh.cmd "wget --retry-connrefused --waitretry=1 -O /tmp/nix-install https://nixos.org/releases/nix/#{version}/install"
             sh.cmd "yes | sh /tmp/nix-install"
 
-            if config[:os] == 'linux'
-              # single-user install (linux)
-              sh.cmd 'source ${TRAVIS_HOME}/.nix-profile/etc/profile.d/nix.sh'
-            else
-              # multi-user install (macos)
-              sh.cmd 'source /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
-            end
+            sh.cmd 'source /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
           end
         end
 
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -34,9 +34,7 @@ module Travis
             sh.cmd "sudo mount -o remount,exec /run/user"
             sh.cmd "sudo mkdir -p -m 0755 /nix/"
             sh.cmd "sudo chown $USER /nix/"
-            # Set nix config dir and make config Hydra compatible
-            sh.cmd "sudo mkdir -p -m 0755 /etc/nix"
-            sh.cmd "echo 'build-max-jobs = 4' | sudo tee /etc/nix/nix.conf > /dev/null"
+            sh.cmd "echo 'build-max-jobs = 4' | sudo tee /tmp/nix.conf > /dev/null"
           end
         end
 
@@ -45,7 +43,7 @@ module Travis
 
           sh.fold 'nix.install' do
             sh.cmd "wget --retry-connrefused --waitretry=1 -O /tmp/nix-install https://releases.nixos.org/nix/nix-#{config[:nix]}/install"
-            sh.cmd "yes | sh /tmp/nix-install --daemon"
+            sh.cmd "yes | sh /tmp/nix-install --daemon --nix-extra-conf-file /tmp/nix.conf"
 
             sh.cmd 'source /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
           end
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -32,8 +32,6 @@ module Travis
           if config[:os] == 'linux'
             sh.cmd "sudo mount -o remount,exec /run"
             sh.cmd "sudo mount -o remount,exec /run/user"
-            sh.cmd "sudo mkdir -p -m 0755 /nix/"
-            sh.cmd "sudo chown $USER /nix/"
             sh.cmd "echo 'build-max-jobs = 4' | sudo tee /tmp/nix.conf > /dev/null"
           end
         end
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -47,7 +47,7 @@ module Travis
 
           sh.fold 'nix.install' do
             sh.cmd "wget --retry-connrefused --waitretry=1 -O /tmp/nix-install https://nixos.org/releases/nix/#{version}/install"
-            sh.cmd "yes | sh /tmp/nix-install"
+            sh.cmd "yes | sh /tmp/nix-install --daemon"
 
             sh.cmd 'source /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
           end
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -9,7 +9,7 @@ module Travis
     class Script
       class Nix < Script
         DEFAULTS = {
-          nix: 'latest'
+          nix: '2.3.6'
         }
 
         def export
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -43,10 +43,8 @@ module Travis
         def setup
           super
 
-          version = config[:nix] == "latest" ? config[:nix] : "nix-#{config[:nix]}"
-
           sh.fold 'nix.install' do
-            sh.cmd "wget --retry-connrefused --waitretry=1 -O /tmp/nix-install https://nixos.org/releases/nix/#{version}/install"
+            sh.cmd "wget --retry-connrefused --waitretry=1 -O /tmp/nix-install https://releases.nixos.org/nix/nix-#{config[:nix]}/install"
             sh.cmd "yes | sh /tmp/nix-install --daemon"
 
             sh.cmd 'source /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
--- a/lib/travis/build/script/nix.rb
+++ b/lib/travis/build/script/nix.rb
@@ -9,7 +9,7 @@ module Travis
     class Script
       class Nix < Script
         DEFAULTS = {
-          nix: '2.0.4'
+          nix: 'latest'
         }
 
         def export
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Thanks! Please open a corresponding docs PR. :-D

asymmetric:
@matthewbauer this breaks on our [macOS builders](https://travis-ci.com/dapphub/dapptools/jobs/272430634#L120) - is it related to the [change in 2.1](https://nixos.org/nix/manual/#ssec-relnotes-2.1)? about multi-user installations?

matthewbauer:
> @matthewbauer this breaks our [macOS builders](https://travis-ci.com/dapphub/dapptools/jobs/272430634#L119) - is it related to the [change in 2.1](https://nixos.org/nix/manual/#ssec-relnotes-2.1) about multi-user installations?

That looks like it. Since the Travis builders have no trouble with --daemon, we should use it here. This would just require passing `--daemon` to /tmp/nix-install.

asymmetric:
@matthewbauer wouldn't there need to be a change [here](https://github.com/travis-ci/travis-build/blob/1aab2ac43b3a5645e67285494bd7823963ee2a8d/lib/travis/build/script/nix.rb#L52-L57) too?

Should it `source /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh` on both OSes, once `--daemon` is set?

matthewbauer:
> @matthewbauer wouldn't there need to be a change [here](https://github.com/travis-ci/travis-build/blob/1aab2ac43b3a5645e67285494bd7823963ee2a8d/lib/travis/build/script/nix.rb#L52-L57) too?
> 
> Should it `source /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh` on both OSes, once `--daemon` is set?

Yeah, I think that sourcing `nix-daemon.sh` on both would be necessary. Alternatively `daemon` could be an option that defaults to true.

lib/travis/build/script/nix.rb
```diff
@@ -43,19 +43,13 @@ def configure
...
             sh.cmd "yes | sh /tmp/nix-install"
 
-            if config[:os] == 'linux'
-              # single-user install (linux)
-              sh.cmd 'source ${TRAVIS_HOME}/.nix-profile/etc/profile.d/nix.sh'
-            else
-              # multi-user install (macos)
-              sh.cmd 'source /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
-            end
+            sh.cmd 'source /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh'
```
asymmetric: @matthewbauer this path does not exist on my NixOS 19.09 installation, do you know if the above will work?

matthewbauer: Yeah, it won't exist on NixOS because it has its own version of the nix-daemon.sh script. You can install it via `sudo nix-env -iA nixos.nix`.

But it should exist on anything that has run `install-multi-user.sh`.

BanzaiMan:
Is this PR ready for testing?

asymmetric:
@BanzaiMan Sure, if there's a way to test this before rolling it out that would be great!

BanzaiMan:
@asymmetric Set up a test repo on https://staging.travis-ci.org or https://staging.travis-ci.com, and let me know. Or, if you have a configuration you'd like to test, I can run a few as well.

asymmetric:
@BanzaiMan I've set up [this repo](https://staging.travis-ci.org/asymmetric/dapptools), but looking at [this job](https://staging.travis-ci.org/asymmetric/dapptools/jobs/756341)  I still see the old version of Nix being used. Is there some manual step you have to perform?

BanzaiMan:
@asymmetric It hasn't been deployed yet. Sorry. I'll do so in about 30 minutes.

BanzaiMan:
@asymmetric It's been deployed to org staging. Please restart your job.

BanzaiMan:
I tried the config https://staging.travis-ci.org/BanzaiMan/travis_staging_test/builds/756349, but [it appears](https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/756350#L165) that https://nixos.org/releases/nix/latest/install is not resulting in  `/nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh`. https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/756350#L195-L197

lib/travis/build/script/nix.rb
```diff
@@ -43,19 +43,13 @@ def configure
         def setup
           super
 
-          version = config[:nix]
+          version = config[:nix] == "latest" ? config[:nix] : "nix-#{config[:nix]}"
 
           sh.fold 'nix.install' do
-            sh.cmd "wget --retry-connrefused --waitretry=1 -O /tmp/nix-install https://nixos.org/releases/nix/nix-#{version}/install"
+            sh.cmd "wget --retry-connrefused --waitretry=1 -O /tmp/nix-install https://nixos.org/releases/nix/#{version}/install"
             sh.cmd "yes | sh /tmp/nix-install"
```
abathur, line 50:
```suggestion
            sh.cmd "yes | sh /tmp/nix-install --daemon"
```

I think this is the gap here.

abathur:
@asymmetric Not expecting a quick response since your status is on vacation, but I wanted to ping on this before I forget.

There's been a temporary disruption in the CI support recently as some SRI hashes had crept into nixpkgs over the last several months that are incompatible with Nix 2.0.4. These have been fixed and 2.0.4 should be working again, but it's also obvious that this is a temporary reprieve.

I suggested an edit that hopefully closes the gap.

lib/travis/build/script/nix.rb
matthewbauer: yes that should be sufficient

asymmetric:
@BanzaiMan is another deployment to staging necessary due to my commit above?

BanzaiMan:
The code still needs to be manually deployed. And I've done so just now. Please test it.

abathur:
@asymmetric @matthewbauer Erg. I gave this a try on my repo but it looks like there are still some issues :(

1. I guess the `/latest/` path has been removed since this started? My default build failed and I get a no such key error for https://nixos.org/releases/nix/latest/install.

2. I tried a second build specifying `nix: 2.3.6` and this also fails. It looks like it starts going bad around https://staging.travis-ci.org/github/abathur/resholved/jobs/760746#L232. I'm not sure what to make of these first two errors.

3. I guess the complaint about relics of an old install at https://staging.travis-ci.org/github/abathur/resholved/jobs/760746#L236 is probably caused by pre-writing /etc/nix/nix.conf at https://github.com/travis-ci/travis-build/blob/185e4da56dc49b6dea2786f6d810a453def0e6c9/lib/travis/build/script/nix.rb#L39. It looks like the [github-action script](https://github.com/cachix/install-nix-action/blob/master/lib/install-nix.sh) handles this by writing the config to /tmp/nix.conf and passing `--nix-extra-conf-file /tmp/nix.conf`.

@BanzaiMan I don't think it matters until we get the above issues ironed out, but I've noticed macOS builds never start on staging. Is it possible to use a mac builder on staging, or will we be unable to test that portion until the change is deployed to production?

asymmetric:
Changed the PR's title, since there's no more `latest` link (due to things being stored on S3, which doesn't have symlinks).

abathur:
@BanzaiMan I think maybe this is ready for another test :)

primeos:
Quick FYI: Since https://github.com/NixOS/nixpkgs/commit/8f9e4bef4c95f057b09cf5914933834c2277dc60 Nix >= 2.2 is required to evaluate Nixpkgs master (and of course the channels that are based on master). And since the default channel is `nixpkgs-unstable` all default builds should be currently broken (AFAIK, I'm a bit surprised that no one mentioned this - [example](https://travis-ci.org/github/primeos/nixpkgs-security-audit/builds/699760993)). This isn't a big problem since it's possible to override the Nix version, but I think it's definitely about time to update the default version :)

abathur:
@primeos I (think?) we're all aware of this minor dumpster fire :)

This PR is one of 3 attempts to update the default Nix version over the last ~8 months, but there's been a pattern of them breaking macOS builds (because the default macOS install mode changed) and getting gummed up around fixing the issue.

The change in Nixpkgs to require Nix >= 2.2 just formalized something that was becoming increasingly true as [SRI hashes incompatible with 2.0.4 have been turning up in Nixpkgs back in November](https://github.com/NixOS/nixpkgs/issues/89275) and breaking any CI builds that used a package with an SRI hash.

In any case, I think we're close to something that can be merged.

abathur:
@BanzaiMan Pinging on getting another deploy to staging to test this, and also my question about whether it's possible to test against a macOS builder on staging.

BanzaiMan:
it's been deployed to https://staging.travis-ci.org/.

abathur:
Progress! (but still one problem). https://staging.travis-ci.org/github/abathur/resholved/builds/760757

It ran clean on macOS, but the Linux build still trips up because /etc/nix exists (https://staging.travis-ci.org/github/abathur/resholved/jobs/760759#L235).

@asymmetric I've created a PR against your patch branch that hopefully fixes this: https://github.com/asymmetric/travis-build/pull/1 by cribbing from the github action installer.

@BanzaiMan Thanks for the deploy and for setting up a macOS builder.

asymmetric:
@abathur merged and pushed, thanks!

abathur:
@BanzaiMan ready for another test, I hope :)

BanzaiMan:
It's been deployed to https://staging.travis-ci.org

abathur:
@asymmetric @BanzaiMan I think this works, now. :)

I restarted all of my previous staging builds, and the first 3 to finish have run green on macOS and Linux:
- [Build 2](https://staging.travis-ci.org/github/abathur/resholved/builds/760741) uses default `language: nix`
- [Build 3](https://staging.travis-ci.org/github/abathur/resholved/builds/760744) specifies nix 2.3.6
- [build 4](https://staging.travis-ci.org/github/abathur/resholved/builds/760757) specifies nix 2.3.5

I 'm also re-running a [many-version build](https://staging.travis-ci.org/github/abathur/resholved/builds/760474) which hasn't finished yet (I'll update this paragraph when it does). This build *should not* be all green--since we started this process, the latest Nixpkgs now requires Nix >= 2.2, so the 2.0.x and 2.1.x builds should fail on each OS (but hopefully the installer completes on all of them).

asymmetric:
@abathur I wish I could give you write-access to this PR, you've been doing most of the work :)

Merged and pushed, fingers crossed!

abathur:
@BanzaiMan I think we're ready again. Thanks for your patience with this...

BanzaiMan:
Deployed again. I'm signing off for the day shortly, and I'll pick it up tomorrow if you need another deployment.

abathur:
@asymmetric @BanzaiMan Looks good!

- As expected, version-specific jobs in [build 1](https://staging.travis-ci.org/github/abathur/resholved/builds/760474) for both macOS and Linux succeed for Nix >=2.2.2. Earlier versions now install correctly, but fail since Nixpkgs no longer supports them.
- Jobs specifying no version in [build 2](https://staging.travis-ci.org/github/abathur/resholved/builds/760741) also succeed on both platforms.

abathur:
@BanzaiMan I went ahead and presumptively created a doc PR (https://github.com/travis-ci/docs-travis-ci-com/pull/2823) matching this change.

Mic92:
cc @domenkozar because this became multi-user, the ci user is no longer a private user, which might break users using custom binary caches i.e. cachix.

domenkozar:
This broke Cachix, travis should do `echo "trusted-users = $USER" | sudo tee -a /etc/nix/nix.conf` and restart the daemon at installation.

Mic92:
This is the fix we applied in our NUR template: https://github.com/nix-community/nur-packages-template/blob/a16929a6f9e2134b2488c82a4abaa4b5b37dba95/.travis.yml#L33

PR #1834, opened 2020-01-07
Use snap for Crystal latest & add cache support

bcardiff:
As a continuation of #1721, this PR make use of snap for latest Crystal release.

~~There was already a condition to avoid running snap in Trusty and Precise. (Although Trusty seems to be supported by snap https://ubuntu.com/blog/snaps-are-now-available-for-ubuntu-14-04-lts-desktop-and-server)~~

~~The condition is left so this is effectively dropping `language: crystal` support for trusty and precise.~~ 

[update: if `snap` is not available crystal latest is installed via apt]

Caching support was also added by:

```
cache: shards

# or 

cache:
  shards: true
```

I am not sure how to test this outside the codebase to double-check everything is in good shape.

I would've wanted to include the crystal/shards version in the cache_slug, and not only the configured version (latest/nightly). But I am not sure if it's possible to capture the output of a command and use that in the cache_slug.

### DIFF
--- a/lib/travis/build/script/crystal.rb
+++ b/lib/travis/build/script/crystal.rb
@@ -13,9 +13,9 @@ module Travis
             when 'linux'
               validate_version
               if config[:crystal] == 'nightly'
-                snap_install_crystal '--channel=latest/edge'
+                linux_nightly
               else
-                snap_install_crystal '--channel=latest/stable'
+                linux_latest
               end
             when 'osx'
               if config[:crystal] && config[:crystal] != "latest"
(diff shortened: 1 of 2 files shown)


bcardiff:
Hi @BanzaiMan I would like to check if there is anything I can do on my side regarding this PR.

BanzaiMan:
@bcardiff Hello! I believe we invited you to our slack so we can arrange testing. Have you seen the invitation?

bcardiff:
Now that you mentioned it, I find it and accepted it \o/. Thanks

lib/travis/build/script/crystal.rb
```diff
@@ -56,6 +58,18 @@ def script
...
+        def setup_cache
+          if data.cache?(:shards) && !cache_dirs.empty?
+            sh.fold 'cache.shards' do
+              directory_cache.add cache_dirs
+            end
+          end
+        end
+
+        def cache_slug
+          super << '-crystal'
```
BanzaiMan: You'd probably want to add a piece about the crystal version in `cache_slug`, in order to avoid name collision.

lib/travis/build/script/crystal.rb
```diff
@@ -65,6 +79,29 @@ def validate_version
...
+          sh.if "-n $(command -v snap)" do
+            snap_install_crystal '--channel=latest/edge'
+          end
+          sh.else do
+            sh.failure "Crystal nightlies will only be supported via snap. Use Xenial or later releases."
+          end
+        end
+
+        def snap_install_crystal(options)
+          sh.cmd %Q(sudo apt-get install -y gcc pkg-config git tzdata libpcre3-dev libevent-dev libyaml-dev libgmp-dev libssl-dev libxml2-dev), echo: true
```
BanzaiMan: Might want to consider sending this command's STDOUT and STDERR to `/dev/null` (or `File::NULL`), in order to avoid excess output. It's also OK if you choose to keep it.

lib/travis/build/script/crystal.rb
bcardiff: The crystal version, from the configuration, is nightly or latest. I could interpolate the configuration value in the slug. But if possible I would rather extract the value of `crystal --version` and use that in part of the slugh. Is there a way to do that? All other build scripts seems to emit the configuration value directly.

PR #1837, opened 2020-01-15
Handle `apt.sources` as a Hash

BanzaiMan:
This 

```yaml
apt:
  sources:
    sourcline: foobarbaz
```
is not documented structure, so it is not strictly required to support it.

Furthermore, `travis-yml` would normalize this incorrect form to something that `travis-build` can handle. So, in the not-so-distant future it would be unnecessary.

### DIFF
--- a/lib/travis/build/addons/apt.rb
+++ b/lib/travis/build/addons/apt.rb
@@ -226,7 +226,7 @@ module Travis
           end
 
           def config_sources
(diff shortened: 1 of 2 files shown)


BanzaiMan:
See https://travis-ci.community/t/odd-error-in-one-repo-works-in-another/6859 for a user report

PR #1842, opened 2020-01-20
Rewrite all maven repo URLs to use https://

BanzaiMan:


BanzaiMan:
See

1. https://travis-ci.community/t/maven-maven-central-repository-restricted-communication-via-https-only-which-breaks-the-stock-settings-xml/6900
1. https://blog.sonatype.com/central-repository-moving-to-https

PR #1843, opened 2020-01-23
Download sonarcloud CLI archive locally and serve it

BanzaiMan:
So that we are less susceptible to Maven Central's serving us the file we want.


BanzaiMan:
Tested: https://staging.travis-ci.org/BanzaiMan/travis_staging_test/jobs/755852#L474

PR #1849, opened 2020-02-03
Enable longpaths git option for Windows

BanzaiMan:
See https://stackoverflow.com/a/22575737/136333

https://travis-ci.community/t/windows-build-long-path-for-files/757

### DIFF
--- a/lib/travis/build/git.rb
+++ b/lib/travis/build/git.rb
@@ -27,6 +27,7 @@ module Travis
 
       def checkout
         disable_interactive_auth
+        enable_longpaths
         install_ssh_key if install_ssh_key?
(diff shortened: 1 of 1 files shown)


BanzaiMan:
Tested https://staging.travis-ci.org/github/BanzaiMan/travis_staging_test/builds/756282#L71

lib/travis/build/git.rb
```diff
@@ -27,6 +27,7 @@ def initialize(sh, data)
 
       def checkout
         disable_interactive_auth
+        enable_longpaths
```
pavel-d: I would move the condition here: `if config[:os] == 'windows'`, but it's just my personal preference

PR #1853, opened 2020-02-11
curl insecure if ssl verification disabled

GbArc:


lib/travis/build/script/shared/directory_cache/base.rb
```diff
@@ -301,7 +301,11 @@ def update_static_file(name, location, remote_location, assert = false)
               cmd_opts = {retry: true, assert: false, echo: 'Installing caching utilities'}
               if casher_branch == 'production'
                 static_file_location = "https://#{app_host}/files/#{name}".output_safe
-                sh.cmd curl_cmd(flags, location, static_file_location), cmd_opts
+                app_host_flags = flags
+                if Travis::Build.config&.ssl&.verify == false
```
BanzaiMan: Is this ever true? The config value actually gets set `false`?

GbArc: yes, for enterprise setup with self-signed certs

lib/travis/build/script/shared/directory_cache/base.rb
```diff
@@ -301,7 +301,11 @@ def update_static_file(name, location, remote_location, assert = false)
...
+                  app_host_flags += '-k'
```
BanzaiMan: I think you need a space before `-`.

lib/travis/build/script/shared/directory_cache/base.rb
BanzaiMan: And not the string `"false"`?

lib/travis/build/script/shared/directory_cache/base.rb
GbArc: indeed, works because flags contain already ' ' at the end, to be updated

lib/travis/build/script/shared/directory_cache/base.rb
GbArc: ssl.verify is FalseClass

GbArc:
needs also casher update with handling of self-signed certs to be fully functional

PR #1864, opened 2020-03-03
Fix for strange symbols in repository id in cache url

piotr-travisci:


BanzaiMan:
What strange symbols are you seeing? It would be useful to understand their nature, and add specs if they are legitimate.

piotr-travisci:
@BanzaiMan 
Repository id can be a number or (new feature) a string like {123-456}
The test is added.

PR #1869, opened 2020-03-11
Don't suppress Wget output

native-api:
https://travis-ci.community/t/the-command-wget-o-travis-home-erlang-tar-bz2-archive-url-failed-and-exited-with-8-during/7531


lib/travis/build/script/erlang.rb
```diff
@@ -74,12 +74,11 @@ def install_erlang(release)
             sh.echo "#{release} is not installed. Downloading and installing pre-build binary.", ansi: :yellow
 
             sh.echo "Downloading archive: ${archive_url}", ansi: :yellow
-            sh.cmd "wget -o ${TRAVIS_HOME}/erlang.tar.bz2 ${archive_url}"
```
BanzaiMan: I think we would want to keep `-O` instead.

I wonder how this worked before, though. 😮

native-api: It works by unpacking `#{archive_name(release)}` rather than "`erlang.tar.bz2`" further in the code. The bogus `erlang.tar.bz2` is ignored.
I don't have a preference here (thus decided to change as little as possible in the logic for the initial version), so using `-O` is fine by me.

PR #1876, opened 2020-03-14
Update elm integration for 2020

harrysarson:
Two main changes:

* A new patch release of elm has been released (elm `0.19.1`). This pull request uses it by default when `language: elm`.

* The way elm versions are tagged on npm has also changed (see [1](https://www.npmjs.com/package/elm?activeTab=versions), [2](https://www.npmjs.com/package/elm-test?activeTab=versions) and [3](https://www.npmjs.com/package/elm-format?activeTab=versions)) to use tags of the form `latest-0.19.1` instead of `elm0.19.1`. This tagging has been applied retrospectively to versions `0.18.0` onwards (all the versions that travis supports). Notably, the `0.19.1` release has *not* been tagged as `elm0.19.1` which means the [`.travis.yml` config](https://github.com/harrysarson/travis-ci/blob/f249a69f6c7c27f2f02e57033bb1372b5e8d88c8/.travis.yml#L1-L3):

  ```yml
  language: elm
  
  elm: elm0.19.1
  ```

  errors when [run in travis](https://travis-ci.com/github/harrysarson/travis-ci/builds/153320693). This PR updates travis-build to use tags of the form `latest-0.19.1` by default.

---

This PR builds #1718, #1613, #1605 and #934

cc: @avh4, @lukewestby, @stoeffel and @rtfeldman

### DIFF
--- a/lib/travis/build/script/elm.rb
+++ b/lib/travis/build/script/elm.rb
@@ -124,16 +124,12 @@ module Travis
             (config[:elm_format] || elm_version_tagged).to_s
           end
 
-          def version_prefix
-            if elm_version =~ /^elm/ || elm_version =~ /^latest-/
-              ""
-            else
-              "latest-"
-            end
-          end
-
           def elm_version_tagged
-            version_prefix + elm_version
+            if elm_version =~ /^(elm|latest-)/
+              elm_version
+            else
+              "latest-#{elm_version}"
+            end
           end
 
           def npm_install_global(package_name, package_version)
--- a/lib/travis/build/script/elm.rb
+++ b/lib/travis/build/script/elm.rb
@@ -3,7 +3,7 @@ module Travis
     class Script
       class Elm < NodeJs
         # Default NodeJS version to install
-        DEFAULT_NODE_VERSION = '13.11.0'
+        DEFAULT_NODE_VERSION = '12'
 
         DEFAULTS = {
           elm: 'latest-0.19.1',
--- a/lib/travis/build/script/elm.rb
+++ b/lib/travis/build/script/elm.rb
@@ -124,8 +124,16 @@ module Travis
             (config[:elm_format] || elm_version_tagged).to_s
           end
 
+          def version_prefix
+            if elm_version =~ /^elm/ || elm_version =~ /^latest-/
+              ""
+            else
+              "latest-"
+            end
+          end
+
           def elm_version_tagged
-            "elm" + elm_version.sub(/^elm/,"")
+            version_prefix + elm_version
           end
 
           def npm_install_global(package_name, package_version)
--- a/lib/travis/build/script/elm.rb
+++ b/lib/travis/build/script/elm.rb
@@ -3,13 +3,13 @@ module Travis
     class Script
       class Elm < NodeJs
         # Default NodeJS version to install
-        DEFAULT_NODE_VERSION = '10.13.0'
+        DEFAULT_NODE_VERSION = '13.11.0'
 
         DEFAULTS = {
           elm: 'elm0.19.0',
         }
 
-        ELM_TEST_REQUIRED_NODE_VERSION = '6.0.0'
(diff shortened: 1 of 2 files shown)


lib/travis/build/script/elm.rb
```diff
@@ -3,13 +3,13 @@ module Build
     class Script
       class Elm < NodeJs
         # Default NodeJS version to install
-        DEFAULT_NODE_VERSION = '10.13.0'
+        DEFAULT_NODE_VERSION = '13.11.0'
```
avh4: imo, we should stick to [node LTS (even-numbered) releases](https://nodejs.org/en/about/releases/), which would mean '12.16.1' for now

harrysarson:
```suggestion
        DEFAULT_NODE_VERSION = '12.16.1'
```

I agree, LTS is better here. Would specifying "12"?

BanzaiMan: This value should be passed to `nvm`, so if `nvm` understands it (I think it does), it's fine.

harrysarson:
```suggestion
        DEFAULT_NODE_VERSION = '12'
```

I agree, LTS is better here. Would specifying "12" work?

harrysarson:
Updated the pull request to use latest node version `12.x.x`.

harrysarson:
I think the CI failures are unrelated to this change as master has also gone red and the error log talks about files this PR does not touch.

lib/travis/build/script/elm.rb
```diff
@@ -124,8 +124,16 @@ def elm_format_version
             (config[:elm_format] || elm_version_tagged).to_s
           end
 
+          def version_prefix
```
BanzaiMan: IMO, this method somehow obscures what we want to achieve. Perhaps it's better to roll it in `#elm_version_tagged`:

```ruby
def elm_version_tagged
  if elm_version =~ /^(elm|latest-)/
    elm_version
  else
    "latest-#{elm_version}"
  end
end
```

lib/travis/build/script/elm.rb
```diff
@@ -3,13 +3,13 @@ module Build
...
-        DEFAULT_NODE_VERSION = '10.13.0'
+        DEFAULT_NODE_VERSION = '12'
 
         DEFAULTS = {
-          elm: 'elm0.19.0',
+          elm: 'latest-0.19.1',
         }
 
-        ELM_TEST_REQUIRED_NODE_VERSION = '6.0.0'
+        ELM_TEST_REQUIRED_NODE_VERSION = '8.0.0'
```
BanzaiMan: Did `elm-test` version change?

harrysarson: elm-test versions match elm language versions. (So elm-test 0.19.1 supports elm tests writing using elm 0.19.1 etc)

lib/travis/build/script/elm.rb
harrysarson: Good point, will change when I get to my computer next :+1:

lib/travis/build/script/elm.rb
BanzaiMan: Right. Unless elm 0.19.1 now requires Node.js 8, there is no reason to change this value.

harrysarson: elm-test 0.19.1 does now require node 8:

https://github.com/rtfeldman/node-test-runner/blob/master/package.json#L7

BanzaiMan: Thanks for checking!

lib/travis/build/script/elm.rb
harrysarson: Fixed!

harrysarson:
Have applied suggestions. Thanks @BanzaiMan

BanzaiMan:
Great! Planning a deployment early next week.

harrysarson:
Thanks @BanzaiMan!

PR #1878, opened 2020-03-30
Update 1 to point to 1.4

abelsiqueira:

PR #1886, opened 2020-04-04
Fix Julia "1" Windows installer options

kimikage:
Since the version "1" points to "1.4" now (cf. PR #1878) and "1" < "1.4", this changes the condition to match both "nightly" and "1".

### DIFF
--- a/lib/travis/build/script/julia.rb
+++ b/lib/travis/build/script/julia.rb
@@ -67,7 +67,7 @@ module Travis
             when 'windows'
               sh.cmd %Q{curl -A "$CURL_USER_AGENT" -sSf -L --retry 7 -o julia-installer.exe '#{julia_url}'}
               sh.cmd 'chmod +x julia-installer.exe'
-              if config[:julia] == 'nightly' || Gem::Version.new(config[:julia]) >= Gem::Version.new('1.4')
+              if uses_inno_setup?
                 sh.cmd %Q{powershell -c 'Start-Process -FilePath julia-installer.exe -ArgumentList "/VERYSILENT /DIR=C:\\julia" -NoNewWindow -Wait'}
               else
                 sh.cmd %Q{powershell -c 'Start-Process -FilePath julia-installer.exe -ArgumentList "/S /D=C:\\julia" -NoNewWindow -Wait'}
@@ -189,6 +189,12 @@ module Travis
               sh.failure "Unknown Julia version: #{julia_version}"
             end
             "https://#{url}"
+          end
+
+          def uses_inno_setup?
+            return false unless config[:os] == 'windows'
+            ver = Array(config[:julia]).first.to_s
+            ver == 'nightly' || ver == '1' || Gem::Version.new(ver) >= Gem::Version.new('1.4')
           end
       end
     end


kimikage:
@BanzaiMan, what should I do now? Should I ask a Julia expert for the review?

FYI, the branch build was passed: https://github.com/kimikage/travis-build/runs/561691017
However, I don't know the way to update the CI status of this PR.

lib/travis/build/script/julia.rb
```diff
@@ -67,7 +67,7 @@ def setup
             when 'windows'
               sh.cmd %Q{curl -A "$CURL_USER_AGENT" -sSf -L --retry 7 -o julia-installer.exe '#{julia_url}'}
               sh.cmd 'chmod +x julia-installer.exe'
-              if config[:julia] == 'nightly' || Gem::Version.new(config[:julia]) >= Gem::Version.new('1.4')
+              if config[:julia] =~ /^(?:nightly|1)$/ || Gem::Version.new(config[:julia]) >= Gem::Version.new('1.4')
```
ararslan, line 70:
```suggestion
              if config[:julia] =~ /^(?:nightly|\d\+)$/ || Gem::Version.new(config[:julia]) >= Gem::Version.new('1.4')
```
Future-proofing for versions after 1 :)

kimikage: Perhaps, when Julia 2.0 is released, more drastic fixes will be required. :smile:
I think it is readable to explicitly specify "1" since "2" is invalid for now.

ararslan: FWIW this is a serious suggestion; I'd rather not have something that we know will eventually be wrong, even if that eventuality is not soon, especially since the fix is just a few extra characters.

kimikage: :confused:
```ruby
irb(main):001:0> Gem::Version.new("2") >= Gem::Version.new("1.4")
=> true
```

BanzaiMan:
Isn't it more constructive to move the Windows (or any) default to `1.4`? Those who wishing to stay on the older versions would specify that, no?

(I see both viewpoints, to be sure.)

lib/travis/build/script/julia.rb
ararslan: Yeah, that's the behavior we want: we want to use the new installer setup for anything 1.4 or later.

ararslan:
> Isn't it more constructive to move the Windows (or any) default to 1.4?

Yep, we have `julia: 1` point to 1.4, but the problem is how the code is currently set up, since the Windows installer changed in 1.4 so we need to handle what options are passed to it accordingly.

BanzaiMan:
Right. So we want to change this line in addition?

https://github.com/travis-ci/travis-build/blob/2ce08b47f18e272e5d49e99277ba8fcaf285a06b/lib/travis/build/script/julia.rb#L14

lib/travis/build/script/julia.rb
BanzaiMan: Right, so numerals `2` and up will be taken care of by `Gem::Version`. A bare `1` is a little unsightly, but having it explicit seems OK to me.

ararslan:
Not sure what you mean, could you elaborate? The default is 1, which should use the latest 1.x release, which is currently 1.4. Do you mean we should set the default to 1.4 instead of 1? We'd still need to handle the user passing 1 when setting up the installer.

lib/travis/build/script/julia.rb
kimikage: Yes, I mean:
```ruby
irb(main):001:0> config = {:julia => '2'}
=> {:julia=>"2"}
irb(main):002:0> config[:julia] =~ /^(?:nightly|1)$/ || Gem::Version.new(config[:julia]) >= Gem::Version.new('1.4')
=> true
```

The "1" is a special case like "nightly", and other `/\d+/` are not.

BanzaiMan:
You're right. There's no need to change the default.

kimikage:
I think refactoring is a good thing. In fact, I don't really like the current script. :sweat_smile:
However, I doubt that *I* should do that *in this PR*. This PR is just a supplement to PR #1878.
Of course, I'm willing to make a separate method just for the one line.

BanzaiMan:
@kimikage #1878 is already merged, so you should think of this as a standalone PR.

kimikage:
Yes, this is definitely a standalone PR for the Travis build system. However, the two PRs are not separable for Julia developers. *Unfortunately*, the fix which this PR changes was left behind, though. :sob:

In any case, I will make the separate method.

kimikage:
I made the separate method (and ran `rebase master`), although I did not change "julia_spec.rb".
cc: @ararslan, @BanzaiMan

kimikage:
Thanks!

PR #1890, opened 2020-04-10
Add slug commit to docker build

DamianSzymanski:


Dockerfile
```diff
@@ -6,6 +6,8 @@ ARG GITHUB_OAUTH_TOKEN=notset
 COPY . .
 
 RUN git describe --always --dirty --tags | tee VERSION
+RUN git rev-parse --short HEAD | tee HEROKU_SLUG_COMMIT
```
BanzaiMan, line 9: I think it is a good idea to use a name that's not associated with Heroku at this time.

DamianSzymanski: @BanzaiMan I changed to HEROKU_SLUG_COMMIT

PR #1895, opened 2020-04-16
Julia: don't pollute package directory with 'juliamnt'

fingolfin:
This can screw up tests if software running in Travis scans its
directory recursively.

It would actually probably be better to also unmount the image; or in
fact, to use attach/detach instead of mount/unmount (see `r.rb`). But to
minimize risk of breakage, I took this conservative approach instead.

ping @ararslan


fingolfin:
I used /tmp because that is what is used for R in r.rb, which also uses hdiutil

BanzaiMan:
`/tmp` should be writable. On macOS, `$TMPDIR` might be safer (with greater isolation).

PR #1899, opened 2020-04-23
Apply branch-specific Settings env vars last

BanzaiMan:
When applying env vars defined in repository settings, be sure to apply
the branch-specific ones after the all-branch ones, so the intended
preferences are respected.

### DIFF
--- a/spec/build/env_spec.rb
+++ b/spec/build/env_spec.rb
@@ -7,7 +7,7 @@ describe Travis::Build::Env do
       pull_request: '100',
       config: { env: ['FOO=foo', 'SECURE BAR=bar'] },
       build: { id: '1', number: '1' },
-      job: { id: '1', number: '1.1', branch: 'foo-(dev)', commit: '03148a8', commit_range: '03148a8..f9da1fd', commit_message: 'the commit message', os: 'linux' },
+      job: { id: '1', number: '1.1', branch: branch, commit: '03148a8', commit_range: '03148a8..f9da1fd', commit_message: 'the commit message', os: 'linux' },
       repository: { slug: 'travis-ci/travis-ci' },
       env_vars: [
         { name: 'BAM', value: 'bam', public: true },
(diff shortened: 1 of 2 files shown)


BanzaiMan:
See https://travis-ci.community/t/precedence-of-environment-variables-available-to-all-branches-vs-to-specific-branch/8310 for motivation.

spec/build/env_spec.rb
```diff
@@ -77,11 +79,11 @@
       end
 
       it 'marks secure vars as secure' do
-        expect(vars.last).to be_secure
+        expect(vars.select {|v| v.secure?}).to_not be_empty
```
BanzaiMan, line 82: The order of `vars` is not guaranteed, so we should not test `.last`.

spec/build/env_spec.rb
```diff
@@ -77,11 +79,11 @@
...
       end
 
       it 'taints secure var values' do
-        expect(vars.last.value).to be_tainted
+        expect(vars.select {|v| v.secure?}.map {|x| x.value.tainted?}.uniq).to eq [true]
```
BanzaiMan, line 86: Ditto.

spec/build/env_spec.rb
```diff
@@ -101,13 +103,31 @@
         expect(keys).to_not include('FOOMASTER')
       end
       it 'includes values restricted to foo-(dev) branch' do
-        expect(vars.find {|var| var.key == 'MULTIBRANCHVARIABLE'}.value).to eq('foodevvalue')
+        expect(vars.select {|var| var.key == 'MULTIBRANCHVARIABLE'}.last.value).to eq('foodevvalue')
```
BanzaiMan, line 106: Ditto.

BanzaiMan, line 106: We want to test that the _last_ one that sets `MULTIBRANCHVARIABLE` has the right value, not the _first_ one returned by `#find`.

BanzaiMan:
Tested https://staging.travis-ci.org/github/BanzaiMan/travis_staging_test/builds/759847#L188

Settings look like:

Screen Shot 2020-04-24 at 14 52 02

PR #1906, opened 2020-05-08
Add `git.symlink` option

BanzaiMan:


spec/build/git/clone_spec.rb
```diff
@@ -175,4 +175,20 @@
...
+    end
+
+    context 'when it is set to true' do
+      before { payload[:config][:git]['symlink'] = true }
+      it { should include_sexp [:cmd, /^git config --global core\.symlinks true/]}
+    end
+
+    context 'when it is set to false' do
+      before { payload[:config][:git]['symlink'] = false }
+      it { should include_sexp [:cmd, /^git config --global core\.symlinks false/]}
```
native-api, line 191: The Git option is [`core.symlinks`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coresymlinks).
The config option is also called "`symlinks`"  is other PRs linked from https://travis-ci.community/t/git-symlinks-support/274/3.

(The 1st test is actuall funny in that it's always passes. It's a known fact in science that it's impossible to prove a negative :-) )

native-api: (same goes for the other changed file)

PR #1913, opened 2020-06-02
Warn use of Chrome on Trusty

BanzaiMan:


BanzaiMan:
Tested here: https://staging.travis-ci.org/github/BanzaiMan/travis_staging_test/builds/760642#L419

BanzaiMan:
It is in a fold and it is hard to see in the beginning. Also, it does not fail the job, so users may have a false impression that it worked. We can discuss whether or not this is the right behavior.

PR #1915, opened 2020-06-03
Set C/C++ defaults for FreeBSD

BanzaiMan:
`cc` for C, `c++` for C++ by default. `gcc` and `g++` are also available on the image.

### DIFF
--- a/lib/travis/build/script.rb
+++ b/lib/travis/build/script.rb
@@ -4,6 +4,7 @@ require 'core_ext/object/false'
 require 'erb'
 require 'rbconfig'
 require 'date'
+require 'byebug'
 
 require 'travis/build/addons'
 require 'travis/build/appliances'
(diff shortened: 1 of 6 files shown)


BanzaiMan:
Since FreeBSD on staging envs are a little broken at the moment, we cannot completely test this PR there. However, we can test the Linux jobs (https://staging.travis-ci.org/github/BanzaiMan/travis_staging_test/builds/760515 https://staging.travis-ci.org/github/BanzaiMan/travis_staging_test/builds/760524).

Since their behavior is preserved, we will deploy this to production for an extra round of testing (with FreeBSD) before merging.

BanzaiMan:
Stealthily deployed to production, and here's the result of a default C++ setup:

https://travis-ci.org/github/BanzaiMan/travis_production_test/builds/695057012#L32-L41

```sh-session
$ export TRAVIS_COMPILER=c++
$ export CXX=${CXX:-c++}
$ export CXX_FOR_BUILD=${CXX_FOR_BUILD:-c++}
$ export CC=${CC:-cc}
$ export CC_FOR_BUILD=${CC_FOR_BUILD:-cc}
$ c++ --version
FreeBSD clang version 8.0.1 (tags/RELEASE_801/final 366581) (based on LLVM 8.0.1)
Target: x86_64-unknown-freebsd12.1
Thread model: posix
InstalledDir: /usr/bin
```

PR #1921, opened 2020-06-27
r: enable c2d4u ppa for R 4.0

jeroen:
See: https://launchpad.net/~c2d4u.team/+archive/ubuntu/c2d4u4.0+


wmay:
I don't want to get on anyone's case, but it seems like this pull request is taking an excessively long time, and this is an important feature of Travis CI for R. Is there something preventing this from being merged?

PR #1922, opened 2020-06-28
Fixes for debug mode on FreeBSD

gezalore:
This contains some fixes that I think should make debug mode somewhat useable on FreeBSD.

What I actually had to do to gain SSH access into a FreeBSD instance is this:

Hook into early bit of build.sh:
https://github.com/verilator/verilator/blob/dc92f34ca48a216d3dd441f23ffe2808cb04dc1a/.travis.yml#L17-L21

Which runs this:
https://github.com/verilator/verilator/blob/dc92f34ca48a216d3dd441f23ffe2808cb04dc1a/ci/travis-env-hack.bash#L11-L37

Hopefully this PR reproduces that though I don't know how I could thest the pathc in the PR itself. The only outstanding isse (if any) should be related to baked in /bin/bash and /usr/bin/tmate which don't exist on FreeBSD.

With the above I can ssh into the instance and debug my build works, though I coulnd't find the travis_run_* functions so some issues might remain.


lib/travis/build/appliances/debug_tools.rb
```diff
@@ -26,6 +26,17 @@ def apply_enabled
                 sh.cmd "wget -q -O tmate.tar.xz #{static_build_linux_url}", echo: false, retry: true
                 sh.cmd "tar --strip-components=1 -xf tmate.tar.xz", echo: false
               end
+              sh.elif "$(uname) = 'FreeBSD'" do
+                sh.cmd 'sudo pkg install -y libevent msgpack libssh', echo: true
```
ararslan, line 30: Note that `sudo` is not available by default on BSDs; there are a few places in the codebase here where that's explicitly worked around by using `su`.

BanzaiMan:
First, thanks for the PR.

While I appreciate the effort (and the difficulty with which `travis-build` PRs need to be tested), I am skeptical that compiling `tmate` from source is the right approach.

I do have a FreeBSD debug session work in https://github.com/travis-ci/travis-build/compare/ha-freebsd; I've tested it a few times, but I have thus far been unable to start a `tmate` session.

gezalore:
No worries, I appreciate the issue of building from source. This PR is more for your information on what you might need to do to get it working. I tried what you have done there, and as I mentioned, the version of tmate in the pkg repository (2.2.3 I think) does not work for me (it reports "[lost server]" on launch) Googling the error message suggest upgrading works, and indeed using 2.4.0 as built from source did the trick. You will also need to fix the baked in file system paths but other than that I could ssh in with my hacks so personally I am set for what I need. I am happy to leave it to you on how best to utilize this PR, including abandon after fixing debug mode properly.

PR #1923, opened 2020-06-29
Support restoring workspace archives from another OS

native-api:
by creating any necessary directory hierarchy on the fly

https://travis-ci.community/t/introducing-workspaces/4249/9

This allows to restore absolute paths regardless of the OS.

Tested the command line on a Travis Python archive in Bionic, OSX Catalina and Windows Git Bash.

---

The stumbling block is currently that it's impossible to extract stuff to `/home` in OSX since it's an automounter mount point.

That could be fixed by altering OSX images:

* remove/comment the `/home` line from `/etc/auto_master` in OSX images; this must be done in the image, or run `sudo launchctl kill HUP system/com.apple.autofsd` after changing it.

`/home` automount point in OSX is only used for mounting a network home dir from a LDAP service, so we don't need it anyway.

---

This may still fail in some unusual setups -- e.g. if root is a readonly filesystem (is this the case with Milti-arch VMs?).

### DIFF
--- a/lib/travis/build/script/shared/workspace.rb
+++ b/lib/travis/build/script/shared/workspace.rb
@@ -40,7 +40,12 @@ module Travis
         end
 
         def expand
-          sh.cmd "tar -xPzf ${CASHER_DIR}/#{name}-fetch.tgz"
+          archive = "${CASHER_DIR}/#{name}-fetch.tgz"
+          # For any directories where `tar` would directly place files,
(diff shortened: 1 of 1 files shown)


vladimiry:
Any news? Looking forward to https://travis-ci.community/t/allow-to-restore-a-workspace-in-another-os/9165 and https://travis-ci.community/t/workspaces-do-not-work-nicely-with-cross-platform-builds/4461 to be fixed.

PR #1926, opened 2020-07-09
Add new name for Perl6: Raku

rbt:
This is not quite backward compatible with perl6 due to renaming environment variables, binaries, etc. The old perl6 support should remain in place for some period (1 year?) for those testing against older versions (before Jan 2020) but I think Perl6 should be replaced in the documentation with Raku. I will provide a doc patch if this way forward makes sense.

I've also used rakubrew to replace rakudobrew. It's setup to use downloaded binaries when available rather than building them; this may save up to 9 minutes when it can be done.

@tony-o, @ugexe, @hoelzro, @paultcochrane: reviews appreciated.


JJ:
Ping? Is anyone paying attention to this?

PR #1935, opened 2020-08-11
Halt Node.js job if `nvm install` fails

BanzaiMan:


BanzaiMan:
Tested: https://staging.travis-ci.org/github/BanzaiMan/travis_staging_test/builds/762847#L426

Probably a good idea to include a reason why the build terminated.

BanzaiMan:
Tested: https://staging.travis-ci.org/github/BanzaiMan/travis_staging_test/builds/762855#L422

PR #1941, opened 2020-08-17
Download R builds from https://github.com/rstudio/r-builds

jimhester:
This changes the installation script to use the R builds available at
https://github.com/rstudio/r-builds rather than the custom travis
specific builds we used previously.

This will reduce the maintenance needed specifically for travis.

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -119,7 +119,7 @@ module Travis
                 os_version = "$(lsb_release -rs | tr -d '.')"
                 r_url = "https://cdn.rstudio.com/r/ubuntu-#{os_version}/pkgs/#{r_filename}"
                 sh.cmd "curl -fLo /tmp/#{r_filename} #{r_url}", retry: true
-                sh.cmd "sudo apt-get install gdebi-core"
+                sh.cmd "sudo apt-get install -y gdebi-core"
                 sh.cmd "sudo gdebi r-#{r_version}_1_amd64.deb"
                 sh.export 'PATH', "/opt/R/#{r_version}/bin:$PATH", echo: false
                 sh.rm "/tmp/#{r_filename}"
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -113,7 +113,7 @@ module Travis
                   "#{optional_apt_pkgs}", retry: true
 
                 r_filename = "r-#{r_version}_1_amd64.deb"
-                os_version = "$(lsb_release -ds | perl -a -e '$F[1] =~ tr/[.]//d; print $F[1]')"
+                os_version = "$(lsb_release -rs | tr -d '.')"
                 r_url = "https://cdn.rstudio.com/r/ubuntu-#{os_version}/pkgs/#{r_filename}"
                 sh.cmd "curl -fLo /tmp/#{r_filename} #{r_url}", retry: true
                 sh.cmd "sudo apt-get install gdebi-core"
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -112,12 +112,13 @@ module Travis
                   'cdbs qpdf texinfo libssh2-1-dev devscripts '\
                   "#{optional_apt_pkgs}", retry: true
 
-                r_filename = "R-#{r_version}-$(lsb_release -cs).xz"
-                r_url = "https://travis-ci.rstudio.org/#{r_filename}"
+                r_filename = "r-#{r_version}_1_amd64.deb"
+                os_version = "$(lsb_release -ds | perl -a -e '$F[1] =~ tr/[.]//d; print $F[1]')"
+                r_url = "https://cdn.rstudio.com/r/ubuntu-#{os_version}/pkgs/#{r_filename}"
                 sh.cmd "curl -fLo /tmp/#{r_filename} #{r_url}", retry: true
-                sh.cmd "tar xJf /tmp/#{r_filename} -C ~"
-                sh.export 'PATH', "${TRAVIS_HOME}/R-bin/bin:$PATH", echo: false
-                sh.export 'LD_LIBRARY_PATH', "${TRAVIS_HOME}/R-bin/lib:$LD_LIBRARY_PATH", echo: false
+                sh.cmd "sudo apt-get install gdebi-core"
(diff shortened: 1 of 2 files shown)


jeroen:
lgtm. @BanzaiMan this should fix the failures with `language: r` on Ubuntu Bionic and up

lib/travis/build/script/r.rb
```diff
@@ -112,12 +112,13 @@ def configure
                   'cdbs qpdf texinfo libssh2-1-dev devscripts '\
                   "#{optional_apt_pkgs}", retry: true
 
-                r_filename = "R-#{r_version}-$(lsb_release -cs).xz"
-                r_url = "https://travis-ci.rstudio.org/#{r_filename}"
+                r_filename = "r-#{r_version}_1_amd64.deb"
+                os_version = "$(lsb_release -ds | perl -a -e '$F[1] =~ tr/[.]//d; print $F[1]')"
```
jeroen, line 116: This looks like bash... does this work in ruby?

jimhester:
If we could get a deploy of this to the staging environment first that would be useful to verify it is correct.

lib/travis/build/script/r.rb
jimhester: Yeah the entire thing gets inserted into the shell commands, so the `$()` does get interpreted by bash, not ruby.

jeroen:
Testing on my ubuntu 20.04 server:

```sh
lsb_release -sc
# focal

lsb_release -ds
# Ubuntu 20.04.1 LTS

lsb_release -ds | perl -a -e '$F[1] =~ tr/[.]//d; print $F[1]'
# 20041
```

lib/travis/build/script/r.rb
```diff
@@ -112,12 +112,13 @@ def configure
                   'cdbs qpdf texinfo libssh2-1-dev devscripts '\
                   "#{optional_apt_pkgs}", retry: true
 
-                r_filename = "R-#{r_version}-$(lsb_release -cs).xz"
-                r_url = "https://travis-ci.rstudio.org/#{r_filename}"
+                r_filename = "r-#{r_version}_1_amd64.deb"
```
BanzaiMan, line 115: Is R meant to work on ARM, or just x86_64? If the former, the file needs to be appropriately specified, and if the latter, the build just halt.

jimhester:
@jeroen how about `lsb_release -rs | tr -d '.'` ? does that give you `2004` on your server? The docker container I am using does not have the point release I guess.

lib/travis/build/script/r.rb
jimhester: Just x86_64, what is the best way to condition on ARM?

jeroen:
@jimhester yes that works:

```sh
lsb_release -rs | tr -d '.'
# 2004
```

lib/travis/build/script/r.rb
```diff
@@ -112,12 +112,13 @@ def configure
...
+                os_version = "$(lsb_release -rs | tr -d '.')"
+                r_url = "https://cdn.rstudio.com/r/ubuntu-#{os_version}/pkgs/#{r_filename}"
                 sh.cmd "curl -fLo /tmp/#{r_filename} #{r_url}", retry: true
-                sh.cmd "tar xJf /tmp/#{r_filename} -C ~"
-                sh.export 'PATH', "${TRAVIS_HOME}/R-bin/bin:$PATH", echo: false
-                sh.export 'LD_LIBRARY_PATH', "${TRAVIS_HOME}/R-bin/lib:$LD_LIBRARY_PATH", echo: false
+                sh.cmd "sudo apt-get install gdebi-core"
```
jeroen, line 119: I think you need `-y` here to prevent a user prompt

jimhester: yeah 👍

jeroen:
I think gdebi also asks for user input. If I try it locally:

```
gdebi -q r-4.0.2_1_amd64.deb
Requires the installation of the following packages: bzip2-doc gfortran gfortran-9 icu-devtools libblas-dev libblas3 libbz2-dev libgfortran-9-dev libgfortran5 libice6 libicu-dev liblapack-dev liblapack3 liblzma-dev libpcre2-16-0 libpcre2-32-0 libpcre2-dev libpcre2-posix2 libsm6 libtcl8.6 libtk8.6 libxss1 libxt6 unzip zip zlib1g-dev

GNU R statistical computation and graphics system
Do you want to install the software package? [y/N]
```

jeroen:
@jimhester you need to add `--non-interactive` to `gdebi` and then it works. See the [manual page](http://manpages.ubuntu.com/manpages/bionic/man1/gdebi.1.html)

jeroen:
@BanzaiMan can we give this a go, so that we can start using `language: r` on Bionic / Focal?

BanzaiMan:
This has been deployed to .org staging. Bionic image works https://staging.travis-ci.org/github/BanzaiMan/travis_staging_test/builds/763066#L21 but Focal still needs some backend work.

jeroen:
Hmm based on the link you send it fails on bionic with `gdebi error, file not found: r-4.0.0_1_amd64.deb`?

jeroen:
Hmm based on the link you send it fails on bionic with `gdebi error, file not found: r-4.0.0_1_amd64.deb`...

@jimhester there is a bug here, you are saving the file in `/tmp/r-4.0.0_1_amd64.deb` in the curl command, however, you are ommitting the `/tmp/` prefix in the gdebi command.

jimhester:
@jeroen Thanks, should be fixed after 08755

jeroen:
Thanks! @BanzaiMan can we try this again? 😄

BanzaiMan:
Sure thing! https://staging.travis-ci.org/github/BanzaiMan/travis_staging_test/builds/763068

BanzaiMan:
The current build failure is due to GHC version update, and it's resolved by https://github.com/travis-ci/travis-build/pull/1948.

jeroen:
OK. So other than that it looks good?

lib/travis/build/script/r.rb
BanzaiMan: I think this is address in the `if` above.

BanzaiMan:
It looks good, as far as I am concerned.

jeroen:
OK so let's deploy?

PR #1942, opened 2020-08-17
Since environment variable values cannot be used as workspace names, do not expand them in the messages either.

adelton:


adelton:
Of course, another possibility would be to get the code and https://docs.travis-ci.com/user/using-workspaces/#workspace-names-must-be-pre-determined changed to actually expand environment variables in workspaces names.

PR #1954, opened 2020-08-31
tweaks for language-r

jeroen:
As a consequence of https://github.com/travis-ci/travis-build/pull/1941, we have dropped support for trusty (the new cdn does not provide R binaries for trusty). This error makes that more obvious for people that are still using `dist: trusty`.


jimhester:
LGTM

jeroen:
Thanks! 🎉

PR #1955, opened 2020-09-04
Show envvar changes made by R installation logic

native-api:
Not seeing them makes diagnosing issues like https://travis-ci.community/t/in-r-version-4-0-0-library-path-not-writable/9744 and https://travis-ci.community/t/r-support-failed-on-xenial/9822 impossible


jeroen:
This can be closed.

native-api:
@jeroen You mean, this information won't be needed in the future? From the build output, it's completely unclear when the `PATH` alteration happens relative to other events. I've no idea how [you figured out](https://github.com/travis-ci/travis-build/pull/1957) that Bioconductor installation runs too early.

native-api:
I also thought to expose the values of envvars in the `export` step. They are arguably even more important than the current installation output: they can be used in build logic so it would be helpful to confirm their existence and exact values in case something that relies on them goes wrong.

jeroen:
I don't think showing all the things on the path will be very helpful to debug these things...

PR #1956, opened 2020-09-06
Install R packages to per-user location

native-api:
r-4.0.0 deb sets the default library location under /opt which is not writable by the user

Should resolve https://travis-ci.community/t/in-r-version-4-0-0-library-path-not-writable/9744

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -327,7 +327,7 @@ module Travis
           sh.echo "Installing R packages: #{packages.join(', ')}"
           pkg_arg = packages_as_arg(packages)
           install_script =
-            "install.packages(#{pkg_arg});"\
+            "install.packages(#{pkg_arg}, lib = Sys.getenv(\"R_LIBS_USER\"));"\
             "if (!all(#{pkg_arg} %in% installed.packages())) {"\
             ' q(status = 1, save = "no")'\
             '}'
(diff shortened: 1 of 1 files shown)


jeroen:
I think it's easier to make the default directory writable.

native-api:
> I don't understand why I am not seeing this in mybuilds. Does this only happen for BioConductor?
> 

I _think_ the problem is with:

https://www.rdocumentation.org/packages/BiocManager/versions/1.30.10/topics/install:

> When installing CRAN or Bioconductor packages, typical arguments include: lib.loc, passed to old.packages() and used to determine the library location of installed packages to be updated

There's lots of preinstalled stuff in `r-4.0.0_1_amd64.deb` in `/opt/R/4.0.0/lib/R/library`. So if some of the packages to be installed already exist there, BiocManager will try to update them.

So I guess we have to _both_ install into `R_USER_LIB` **and** make `/opt` writable.

> I think it's easier to make the default directory writable.

Then caching is going to be a problem. Which of the `.linbPaths` directories to cache? We will also cache the preinstalled `.deb` stuff if we're going to cache `/opt/<...>`.

native-api:
Need to test running equivalent commands manually with `--verbose` to check this :\ Currently, it's just a guess.
Can't do this right now.

jeroen:
Well the current solution is too complex. 

I don't use bioconductor myself, but it sounds like the problem is not specific to travis, and can be simulated by installing R in a docker image and then running BiocInstaller?

native-api:
> Well the current pr is too complex and I don't think it solves the problem.

I agree.

> I don't use bioconductor myself, but it sounds like the problem is not specific to travis, and can be simulated by installing R in a docker image and then running BiocInstaller?

Who knows. The current logic doesn't give any diagnostic output to understand what's happening.

My best idea is to clone OP's repo, `travis compile` a script for it locally, then override `install:` in its `.travis.yml` to run the equivalent commands with `--verbose`. Hopefully, we can get some diagnostic info from R this way to understand what exactly is wrong.

native-api:
Closing this for now as it's unclear without testing manually whether this solution would help

PR #1958, opened 2020-09-12
Install bioc and requested packages at setup stage

native-api:
This allows the user to override install: without losing this critical part of R build logic.

See https://travis-ci.community/t/in-r-version-4-0-0-library-path-not-writable/9744/22?u=native-api for the user problem that this fixes. I also have trouble checking things out since I cannot get a large part of the logic without a dummy R package or override `install:` which is extremely inconvenient.

### DIFF
--- a/lib/travis/build/script/r.rb
+++ b/lib/travis/build/script/r.rb
@@ -189,6 +189,24 @@ module Travis
           end
         end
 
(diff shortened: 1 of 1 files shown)


native-api:
ping @jeroen @jimhester

PR #1959, opened 2020-09-14
Allow R packages to install scripts

native-api:
Implements option 1 from https://travis-ci.community/t/in-r-version-4-0-0-library-path-not-writable/9744/27?u=native-api.

R has no per-user script directory to complement R_LIBS_USER.
So packages have no option but to install scripts to $R_HOME/bin.
With the `r` Apt package since 4.0, regular users don't have permission
to write there.

The commit allows the current user to create new executables there while
disallowing messing with existing ones.


sfirke:
Looking forward to this.  In case it helps: I have an [R package Travis build](https://travis-ci.com/github/sfirke/janitor/builds/184403579) that is failing for R release but passing on R devel.  On R release, 4.0.2, it says:
> Warning in install.packages(update[instlib == l, "Package"], l, contriburl = contriburl,  :
  'lib = "/opt/R/4.0.2/lib/R/library"' is not writable

PR #1964, opened 2020-10-15
Create ~/.ssh if necessary

native-api:
https://travis-ci.community/t/pushing-to-github-from-windows-host-key-verification-failed/9495

### DIFF
--- a/lib/travis/build/addons/ssh_known_hosts.rb
+++ b/lib/travis/build/addons/ssh_known_hosts.rb
@@ -25,6 +25,7 @@ module Travis
           def add_ssh_known_hosts
             sh.fold 'ssh_known_hosts.0' do
               sh.echo "Adding ssh known hosts", ansi: :yellow
+              sh.mkdir "${TRAVIS_HOME}/.ssh", recursive: true
               config.each do |host|
                 begin
                   host_uri = URI("ssh://#{host}")


lib/travis/build/addons/ssh_known_hosts.rb
```diff
@@ -25,6 +25,7 @@ def config
           def add_ssh_known_hosts
             sh.fold 'ssh_known_hosts.0' do
               sh.echo "Adding ssh known hosts", ansi: :yellow
+              sh.cmd 'mkdir -p "${TRAVIS_HOME}/.ssh"'
```
BanzaiMan, line 28: We should be more uniform in use of certain methods, but here, we can use:

```ruby
sh.mkdir "${TRAVIS_HOME}/.ssh", recursive: true
```

native-api: Sure. Didn't think of it since some code below used a raw command for mkdir.

BanzaiMan: Right. We should do better with respect to the uniformity throughout the code base.

annamajorova0278:
@ #1964

annamajorova0278:
296b15

annamajorova0278:
### 1964

PR #1966, opened 2020-10-30
use r devel for bioc-devel

LiNk-NY:
This updates Travis to use the appropriate version of R with the Bioconductor release.

Also, R release should be `4.0.3`. I didn't change it but it should be changed. 
Thanks!
@BanzaiMan @jimhester @jeroen


LiNk-NY:
Checks are no passing on Bioconductor. This is something that needs to be changed every Bioconductor release. 
Is there any way to get this merged in a timely manner? Otherwise, TravisCI can't be used by the Bioconductor
community. 
@BanzaiMan @jimhester @jeroen
Thanks, 
Marcel

PR #1972, opened 2020-11-16
Install prerelease versions of `dpl`

native-api:
https://travis-ci.community/t/cant-deploy-to-pypi-anymore-pkg-resources-contextualversionconflict-importlib-metadata-0-18/10494/23

**WARNING!!! This will make _every_ commit to `dpl/v1` go into production (if the build is successful)!!!** 

But **without this, you need to bump `DPL::VERSION` and make a tag by hand every time when `dpl/v1` should go into production.**

So think which one you want!

[ ] **Supplemental PR:** https://github.com/travis-ci/dpl/pull/1224

### DIFF
--- a/lib/travis/build/addons/deploy/script.rb
+++ b/lib/travis/build/addons/deploy/script.rb
@@ -240,14 +240,14 @@ module Travis
                 build_gem_locally_from(src, branch)
               end
 
-              command = "gem install"
+              command = "gem install --pre"
               if install_local?(edge)
                 command << " $TRAVIS_BUILD_DIR/dpl-*.gem"
               else
                 command << " dpl"
               end
-              command << " -v '< 1.9' " if want_pre_19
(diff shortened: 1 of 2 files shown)


BanzaiMan:
I think it is best to keep releasing the v1.x gems.

native-api:
Glad that this is over! 🎉

This entire premise started when you didn't after https://github.com/travis-ci/dpl/commit/a6787407acbb20cbef8ded0302166fd5a1857932 and ppl complained about not seeing the fix -- so I started questioning what the intended release strategy is...

native-api:
I think I'll leave the branch put for now in case you change your mind.

AFAICS, `dpl v1` is only getting bugfixes now -- which should be released ASAP. So making a release by hand after each one looks like a lot of extra work -- especially now when you are apparently all overworked with the new integrations and stuff.

PR #1976, opened 2020-12-27
Allow newlines and other shell metacharacters in deploy:script: and o…

native-api:
…ther DPL arguments

https://travis-ci.community/t/deploy-scripts-not-supporting-multiline-strings/8297

Considered alternatives:
* `eval` the values encoded with %p in DPL -- unsafe
* use JSON encoding -- unnecessary and still need to escape shell metacharacters in the result


BanzaiMan:
Thank you!

BanzaiMan:
Looks like this is causing https://travis-ci.community/t/s3-aws-deploy-access-key-id/10967.

I'll revert it.

PR #1980, opened 2021-03-11
Updates sbt-extras version to current master

froth:
The old version is broken for sbt versions starting from 1.4.9

See https://travis-ci.community/t/scala-build-broken-for-sbt-1-4-9/11284

Also: https://github.com/scala/scala-dev/issues/765


froth:
I am pretty unsure whether the failing build output has something to do with my change. It does not seem so but if it does: if someone explains it toe I am willing to fix it

PR #1984, opened 2021-04-13
Unsilence JDK install commands 

native-api:
There's currently an error installing JDK 8 in ARM64 and due to silencing, it's impossible to see what's wrong.

https://travis-ci.community/t/maven-deploy-multiarch/11450/7

[Later in the build, the same commands work okay](https://travis-ci.com/github/native-api/test_travis/builds/222977541) -- so I can't disgnose the problem otherwise.


native-api:
Alternatively, I know there's a staging environment. If it has Arm64 machines (ideally, from the same pool as production ones), I could tests there if I'm given access.

native-api:
Failure unrelated, it's the Docker pull limits.

PR #2039, opened 2022-08-05
Trusty JDK workaround

smokodon:
A workaround for Trusty-AMD issues with new JDK provider.


smokodon:
Need to get some information about Trusty maintenance before merging.

r-arek (approved):

PR #2079, opened 2023-03-28
BSFY-232: Dev fetchdepth

everything14:


lib/travis/vcs/git/clone.rb
```diff
@@ -145,7 +145,8 @@ def pull_args
           end
 
           def fetch_args
-            args = ""
+            args = " "
```
murtaza-swati, line 148: is this expected change or is this accidental?

murtaza-swati (changes requested):

lib/travis/vcs/git/clone.rb
GbArc: yes, by purpose, in next previously it was " --quiet", now the space is needed somewhere before depth_flag, so either depth_flag needs to be changed, or space put here or in next line

murtaza-swati: Thanks @GbArc for explaining, I was just making sure.

PR #2121, opened 2024-05-25
TBT-17

piotrmilcarz:


murtaza-swati (approved):

murtaza-swati (changes requested):
@piotrmilcarz the build is failing probably due to ghc, can you please fix it?
try `bundle exec rake ghc_json`

PR #2164, opened 2025-07-18
TBT-396 Skip deploy when dpl is v2

tonoyansergey-devt:
Change: build.sh does not run DPLv2 instructions under old Windows and prints message in job log instead


murtaza-swati (approved):
LGTM

